PR #70, opened 2018-02-06
Fetch and WS clients init (OGUI-124)

vladimirkosmala:


Frontend/js/src/WebSocketClient.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // handling authentification success
+    if (parsed.command == 'authed') {
+      this.authed = true;
+      this.dispatchEvent(new Event('authed', parsed));
+      return;
+    }
+
+    // handling token refresh error
```
awegrzyn: After our discussion I actually disabled token refresh but the code can stay in case

Frontend/js/src/sessionService.js
```diff
@@ -0,0 +1,61 @@
+/*global: window*/
```
awegrzyn: I'd say it's more like key-value (or parameter) service. There are no sessions, what do you think?

PR #75, opened 2018-02-08
Add skeleton project to docs

vladimirkosmala:


Skeleton/index.js
```diff
@@ -0,0 +1,35 @@
+// Import the backend framework
+const {HttpServer, Log, WebSocket, WebSocketMessage} = require('Gui');
+
+// Activate debug logs
+Log.configure({winston: {}});
```
awegrzyn: I'm not sure this line is required

Skeleton/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
awegrzyn: Having another package.json within the project won't break anything?

Skeleton/index.js
vladimirkosmala: Indeed, a default is created on the fly

Skeleton/package.json
vladimirkosmala: it's ok to have a package.json on each folder you want to have special dependencies.
the resolver only search for node_modules by going up in the file tree

vladimirkosmala:
@awegrzyn backend without OAuth should send token and userid like usual for the front to work well. Let's say a random token and "Anonymous" userid. Then the front will work again in this mode.
The front now throw when this occurs.

awegrzyn:
@vladimirkosmala Indeed very good remark, I'll implement changes as you described

awegrzyn:
Okay, it's been addressed in https://github.com/AliceO2Group/Gui/pull/78

PR #96, opened 2018-03-13
Fixed issue with JSON Web token module not being able to load

freekdotcom:
When trying to complete the tutorial for the time server, I had to create a JSON Web token. When trying to run the server, I got the error: "JwToken is undefined" at the moment when the example server is executed. I fixed this with adding the  modules to the index.js.


awegrzyn:
Replacing by https://github.com/AliceO2Group/WebUi/pull/97

PR #106, opened 2018-04-06
Added mysql guide

vladimirkosmala:


awegrzyn:
@vladimirkosmala I've changed a bit the structure and fixed the code example as it did not work

PR #127, opened 2018-04-24
[WIP] Added QCG frontend tests

vladimirkosmala:


vladimirkosmala:
Here are some basic tests using chrome headless.
Run `npm start test/test-config.js`
Run `npm run mocha`

awegrzyn:
It looks really nice, I assume I need to start chrome manually?
I will try to add it to travis.

awegrzyn:
Ok so on my machine 3 are passing and 4 failing

vladimirkosmala:
Could you post the output?
I changed command to start server in addition to tests.

`npm run mocha`

awegrzyn:
Now the problem is that mocha tests start before the server is present therefore they fail due to connection refused

awegrzyn:
Okay after b9ecd61 everything works!

awegrzyn:
Ok it works on travis. Do you think we can use `config-default.js` instead of `test/test-config.js`?
Hostname and port in `test/mocha-index.js` can be parametrised with `${config.http.hostname}` and `${config.http.port}`.

PR #131, opened 2018-04-25
Added compatibility checker to WebUi and QCG

vladimirkosmala:
This will need WebUi to be published then QCG upgraded to the new version.

### DIFF
--- a/Framework/Backend/http/server.js
+++ b/Framework/Backend/http/server.js
@@ -73,7 +73,7 @@ class HttpServer {
       directives: {
         /* eslint-disable */
         defaultSrc: ["'self'", "data:"],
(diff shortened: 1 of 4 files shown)


vladimirkosmala:
https://alice.its.cern.ch/jira/browse/OGUI-157

awegrzyn:
Are you sure `alert` is a good way to communicate with the user? There are many browsers blocking alters by default.

awegrzyn:
In addition users with older browsers will get multiple alerts.

vladimirkosmala:
=> changing to static webpage to show incompatibility

awegrzyn:
I'd also display generic error message, and display which test failed in the console only

PR #133, opened 2018-04-30
Added online mode to QCG

vladimirkosmala:


awegrzyn:
Two small improvements:
- Either `config-default` informationService section should be uncommented or default values hardcoded
- I think `zeromq` dependency is not necessary

vladimirkosmala:
Yes definitely

PR #161, opened 2018-07-09
Fix router overflow by too much events

vladimirkosmala:


awegrzyn:
Do I understand well that `callThroughput` limits method execution `fn` to once per `time` window specified as second parameter?

vladimirkosmala:
I agree with the `callRateLimiter` proposal, it's a better name

PR #176, opened 2018-07-25
Added time serie based charting to framework

vladimirkosmala:


vladimirkosmala:
It serves more than chart, it's a tool to serve all statics files of the framework to work on it (like CSS).

Will add into devel documentation.

PR #182, opened 2018-07-31
Added message on connection lost for ILG

vladimirkosmala:


awegrzyn:
I've noticed that in same places we use arrow functions and in others `bind(this)`. Which syntax would you recommend?

awegrzyn:
Ok thanks for the explanation

PR #189, opened 2018-08-02
Allow OAuth in Control and InfoLogger

awegrzyn:


awegrzyn:
In general OAuth is optional and it will be used only in production, so I wouldn't include it in default config. If the values are missing there the OAuth configuration will be skipped.

PR #192, opened 2018-08-02
Store users configuration data in local JSON file instead of MySQL

vladimirkosmala:


awegrzyn:
Looks good, I will just add travis test and we can merge.

QualityControl/test/mocha-config.js
```diff
@@ -0,0 +1,45 @@
...
+      assert.strictEqual(TEST_LAYOUT.owner_id, layout.owner_id);
+      done();
+    });
+  });
+
+  it('Create layout with the same name', (done) => {
+    let layout = TEST_LAYOUT;
+    layout.id = 321;
+    jsonConfig.createLayout(layout).then(() => {
+      done(); // THIS SHOULD NOT BE POSSIBLE !!!
```
awegrzyn: This should not pass

awegrzyn:
I'd also rename `writeLayout` to `updateLayout`

QualityControl/test/mocha-config.js
vladimirkosmala: dd

awegrzyn:
Okay, but I think now it's a bit too strict, all users share the same layout name space. Two different users should be allowed setting the same name.

vladimirkosmala:
@awegrzyn ready

vladimirkosmala:
EXDEV error come from the fact that rename() should be atomic but it's impossible if both locations are in two separate disk.

I will place the temp file on same directory.

PR #204, opened 2018-08-08
Renamed properties of time-series

vladimirkosmala:


vladimirkosmala:
@awegrzyn you might want to update jsfiddle after publishing patch of WebUi

awegrzyn:
Will do

PR #205, opened 2018-08-08
Removed timezone choice

vladimirkosmala:


vladimirkosmala:
Still needed to use CEST timezone only

PR #211, opened 2018-08-09
QCG public API

vladimirkosmala:


vladimirkosmala:
see https://github.com/AliceO2Group/WebUi/pull/213

PR #217, opened 2018-08-14
Label the logs

awegrzyn:


Framework/Backend/log/InfoLoggerReceiver.js
```diff
@@ -17,7 +17,7 @@ module.exports = class InfoLoggerReceiver extends EventEmitter {
    */
   constructor() {
     super();
-
+    this.log = new Log('InfoLoggerReceiver');
```
vladimirkosmala: I would have put Log instance otherwise like:
`const log = new require('./Log.js')('InfoLoggerReceiver');`

To avoid pollution of namespace `this.log =` and so the use of Log is also the same for non-class files like index.js
it is also compatible with the current code, less modifications and better readability (imo).
What do you think?

vladimirkosmala: Here is an example from expressjs:

https://github.com/expressjs/express/blob/master/lib/router/route.js#L16

awegrzyn: I think it's a good idea. I'll make changes to the PR.

vladimirkosmala:
The changes to log.js (Log.js now) are breaking changes, bump version should be 2.0.0

it also can be made compatible (thus, keeping 1.6.0) by adding the previous static methods to the new Log class.

awegrzyn:
I'll adapt the code in other projects

PR #221, opened 2018-08-15
Added default JWT config

vladimirkosmala:


awegrzyn:
I'm not sure this is what we want to do. The idea was to provide minimal config, and JWT is optional. I'd be more in favour of documenting it in the JWT guide.

vladimirkosmala:
Indeed, but we also don't want user to have JwtToken error if not configured correctly.

So, an other solution is to remove Math.random() inside Framework for token generation.

vladimirkosmala:
Moreover, even if config-default.js is minimal, it is nice to have examples (even commented) to quickly configure without spending too much time in documentation. Feedback from Barth.

awegrzyn:
We can't remove `Math.random()` from the secret generator as all application would have the same secret.
I think having commented part of default config is good tradeoff. But in such case we should add also examples of other possible configs.

vladimirkosmala:
I agree with commented parts.

So still, user might see JwtToken if server is restart but client not reloaded, do we assume that? Piotr experimented the error. Token could be refreshed on error, but this implies implementation.

awegrzyn:
We could just redirect to the main page when the token error occurs. This should regenerated the token.

vladimirkosmala:
@awegrzyn another review could be done

awegrzyn:
Okay, so eventually just commented JWT config?

vladimirkosmala:
- Commented jwt default config
- added WS to QCG with 'close' event to ask page reload, which renew token

PR #271, opened 2019-04-10
Workaround for seg. fault when asynchronously converting objects afte…

awegrzyn:
…r server launch


awegrzyn:
@GeorgeRaduta Could you take a look whether it solves the issue on your laptop as well?

graduta:
> @GeorgeRaduta Could you take a look whether it solves the issue on your laptop as well?

Seems to work fine. Will keep an eye on it

PR #275, opened 2019-04-26
OGUI-257 scroll to last message

graduta:


awegrzyn:
Hi,
So ALT-arrow works well, but when I hit "Query" button it does not scroll to the last message, maybe I'm missing something.

graduta:
> Hi,
> So ALT-arrow works well, but when I hit "Query" button it does not scroll to the last message, maybe I'm missing something.

Action to go to last item in the list was not linked to the Query action. Pushed change now

PR #280, opened 2019-05-09
OGUI-260 Disable filter boxes when mode=live-running

graduta:
Changes in this PR:
* Input Filters were moved under the "User button bar" so that they do not affect the position of the bar when hiding filters in live mode
* Input Filters will be shown in Query mode and hidden in Live mode

_Query Mode:_
Screenshot 2019-05-09 at 13 08 53
_Live Mode Running:_
Screenshot 2019-05-09 at 13 09 10
_Live Mode Paused:_
Screenshot 2019-05-09 at 13 59 53

### DIFF
--- a/Control/protobuf/o2control.proto
+++ b/Control/protobuf/o2control.proto
@@ -83,12 +83,22 @@ message StatusUpdate {
 // Framework
 ////////////////////////////////////////
 message GetFrameworkInfoRequest {}
+message Version {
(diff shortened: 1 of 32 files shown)


awegrzyn:
Hmm, I think the filter should be enabled when leaving "Live" mode.
Currently if I want to filter something I need to first hit "Query" button which ends up querying all the logs (this may take quite some time).

PR #294, opened 2019-05-23
[OGUI-314] Bump node-addon-api as it contains fixes for Napi::AsyncWorker

awegrzyn:


awegrzyn:
This works well with `aliBuild` but it fails on `zmq.node` compilation. Let's wait for QC online mode which totally drops` zeromq`.

PR #296, opened 2019-05-28
Fix log.trace call

awegrzyn:


graduta:
Could you please add the same changes for this as well? https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/http/server.js#L147

PR #335, opened 2019-06-20
[OGUI-352] Move shutdown option to all envs page

graduta:
Changes in PR:
* `Delete` button from environment page was removed
* A `Shut Down` button was added in the environments page table
* User will be asked to confirm shutting down the environment
* In case state is not `locked` a danger notification(5s) will be displayed  when user tries to shut down the environment

Screenshot 2019-06-21 at 13 16 13
Screenshot 2019-06-21 at 13 16 46

### DIFF
--- a/Control/public/common/showTableList.js
+++ b/Control/public/common/showTableList.js
@@ -5,15 +5,17 @@ import parseObject from './utils.js';
  * Generic table to show list of objects
  * This can be forked to show more specific data (format date, colors, more buttons...)
  * @param {Array.<Object.<string, Any>>} list - things to be shown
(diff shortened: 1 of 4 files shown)


vascobarroso:
Looks good! 
I'd suggest 2 small changes: 

- Shut Down => Shutdown (single word)
- should be disabled if env is RUNNING

graduta:
> Looks good!
> I'd suggest 2 small changes:
> 
>     * Shut Down => Shutdown (single word)
> 
>     * should be disabled if env is RUNNING

Comments addressed. Just in case the user would try to shutdown a running environment they would not be able to do that an a red notification would be prompted

PR #349, opened 2019-07-08
[OMON-214] Update Grafana panel URLs

awegrzyn:


Control/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@aliceo2/control",
   "description": "ALICE O2 Control GUI",
-  "version": "1.1.2",
+  "version": "1.1.3",
```
graduta: Kafka browser notifications are included right now in dev so they will go out with the next release.
We might want to release with 1.2.0

awegrzyn: Ok, could you proceed with that?

graduta: Sure

PR #354, opened 2019-07-19
[OGUI-371] Disable input fields rather than hiding them

graduta:
Screenshot 2019-07-19 at 13 51 28

To be debated during meeting.


awegrzyn:
Let's wait with this until tomorrow.

awegrzyn:
So after discussions users prefer to change filters without going out of "Live" mode.
This means updated filter function should be passed to server on `onfocusout` event.

PR #355, opened 2019-07-22
[OGUI-373] Add testing to SQLDataSource & Fix bug of rows returned

graduta:
PR containing: 
* Bug fix in which rows from db were returned from top rather than bottom
* Testing and refactoring for the SQLDataSource


InfoLogger/test/lib/mocha-sqldatasource.js
```diff
@@ -0,0 +1,221 @@
...
+const SQLDataSource = require('../../lib/SQLDataSource.js');
+const {MySQL} = require('@aliceo2/web-ui');
+
+
+describe('SQLDataSource', () => {
+  const config = {
+    mysql: {
+      host: 'localhost',
+      user: 'root',
+      password: 'admintest',
```
awegrzyn: We can't have any password in the code, even dummy ones.

InfoLogger/test/lib/mocha-sqldatasource.js
```diff
@@ -0,0 +1,221 @@
...
+      database: 'INFOLOGGER',
+      port: 8909
```
awegrzyn: I have DB running at different port, maybe let's just have MySQL checks only at CI? This is already the case for https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/test/mocha-db.js

graduta: These are just unit tests and will not make use of any mysql instances. The configuration is only used to mock creating an instance of mysql driver

InfoLogger/test/lib/mocha-sqldatasource.js
graduta: them

InfoLogger/test/lib/mocha-sqldatasource.js
awegrzyn: Ah okay, maybe just use values from `config-default.js`?

graduta: Updated to use `config-default.js`

PR #357, opened 2019-07-23
Fix array assert in ILG test

awegrzyn:


InfoLogger/test/mocha-index.js
```diff
@@ -141,9 +141,8 @@ describe('InfoLogger', function() {
         window.model.log.filter.setCriteria('pid', 'in', '123 456');
         return window.model.log.filter.criterias.pid.$in;
       });
-
       assert.strictEqual($in.length, 2);
-      assert.strictEqual($in ['123', '456']);
+      assert.deepEqual($in, ['123', '456']);
```
graduta: deepEqual is deprecated so we should move away from it. 
I updated this test file with all assertions to use `deepStrictEqual` in https://github.com/AliceO2Group/WebUi/pull/356

graduta: #356 was now merged so we can close this PR

awegrzyn:
Are you sure `assert.deepStrictEqual($in['123', '456']);` is correct?

PR #374, opened 2019-09-16
Migrate Travis jobs to GitHub Actions

awegrzyn:


.github/workflows/qc.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '10.x'
+      - run: (cd QualityControl/lib; sed -i -e "s/require('..\/build\/Release\/tobject2json.node')/null/g" TObject2JsonClient.js)
+      - run: (cd QualityControl; rm -f binding.gyp; sed -i.bak -e '7,8d' package.json)
+      - run: (cd QualityControl; npm install )
+      - run: (cd QualityControl; npm run coverage )
+      - run: (cd QualityControl; ./node_modules/.bin/nyc report --reporter=text-lcov > coverage.lcov)
```
graduta: I do not think you need to specify the whole path for nyc. 
In package.json we have the `npm run coverage` command which makes use of simply `nyc`.
Applicable in all 3 projects

awegrzyn: Full path is required, otherwise it fails.

PR #376, opened 2019-09-18
[OGUI-397] Add alp as default option

graduta:


awegrzyn:
Ok, so this test is failing also in Travis

PR #401, opened 2019-10-28
[OGUI-414] Bump dependencies all

graduta:


awegrzyn:
That's weird that after bump timeout had to be increased from 5s to 20s.

graduta:
> That's weird that after bump timeout had to be increased from 5s to 20s.

I was not able to replicate the timeout issue anymore so I reverted to 5s

PR #402, opened 2019-11-05
[OGUI-420] Improve task table

graduta:
Changes:
* All Tasks are now requested as soon as the environment has been loaded. 
* Removed some columns from task table and updated others
* Each task request corresponds to a different RemoteData to allow the user to know if one of them fails.

### DIFF
--- a/Control/package-lock.json
+++ b/Control/package-lock.json
@@ -248,9 +248,9 @@
       "dev": true
     },
     "agent-base": {
-      "version": "4.2.1",
-      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.1.tgz",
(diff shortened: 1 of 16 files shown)


graduta:
> Actually, we should not do any magic parsing environment details, because when it changes it will lead to errors, what about asking Teo to provide additional values?

Which values are you referring to? Are you referring to the task name?

vascobarroso:
I agree that for the task short name, it would make sense to have it provided by the core (it already exists in the `yaml` definition). But I would prefer to have this available now and then remove the parsing once core provides it.

vascobarroso:
For this to make it into FLP Suite v0.3.0, we need a new release by next Wednesday (13/Nov). As mentioned before, I'd prefer to have the magic parsing now to unblock this and later remove it once the core provides the shortened task name.

PR #403, opened 2019-11-06
[OGUI-424] Allow copying selected row

graduta:


awegrzyn:
Could you make sure that it works on FF @ CC7?

PR #445, opened 2020-02-11
[OGUI-485] Inform user on ILG Server connectivity isues

graduta:
Notify the user with a notification if the InfoLogger Server is not responding. A new attempt is made after 5 seconds.


awegrzyn:
This is good, but when I start server (`npm start`) I still get exception from `Backend/db/mysql.js:41`.

graduta:
> This is good, but when I start server (`npm start`) I still get exception from `Backend/db/mysql.js:41`.

This PR concerns only the InfoLogger Server connectivity issues which is being used for Live mode. 

The one regarding MySQL is OGUI-484 and will need changes in both InfoLogger and Framework mysql contructor.

awegrzyn:
ok sorry my bad

PR #451, opened 2020-02-18
Migrate OAuth to OpenID Connect

awegrzyn:
TODO:
- [x] state
- [x] egroups


Framework/Backend/http/openid.js
```diff
@@ -0,0 +1,84 @@
...
+  * @param {object} config - id, secret, welll_known, redirect_uri
+  */
+  constructor(config) {
+    assert(config.id, 'Missing config value: OpenID.id');
+    assert(config.secret, 'Missing config value: OpenID.secret');
+    assert(config.well_known, 'Missing config value: OpenId.well_known');
+    assert(config.redirect_uri, 'Missing config value: OpenId.redirect_uri');
+    this.config = config;
+    this.code_verifier = generators.codeVerifier();
+    this.createIssuer();
```
graduta: A constructor should only initialize the module and not try to create the issuer. 

This was the same issue we had with the constructor crashing the server.

Framework/Backend/http/openid.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  /**
+   * Create OpenID client based on well-known endpoint
+   */
+  createIssuer() {
+    Issuer.discover(this.config.well_known)
+      .then((issuer) => {
```
graduta: There is no `catch` here, what happens if fails?

Framework/Backend/http/openid.js
```diff
@@ -0,0 +1,84 @@
...
+        this.client = new issuer.Client({
+          client_id: this.config.id,
+          client_secret: this.config.secret,
+          redirect_uris: [this.config.redirect_uri],
+          response_types: ['code'],
+          id_token_signed_response_alg: 'RS256',
+          token_endpoint_auth_method: 'client_secret_basic'
+        });
+        log.info('OpenID client initalized');
```
graduta: `initialized`

Framework/Backend/http/openid.js
```diff
@@ -0,0 +1,84 @@
...
+      scope: 'openid',
+      codeChallenge,
+      code_challenge_method: 'S256',
+      state: state
+    });
+  }
+
+  /**
+   * Handles callback
+   * @param {object} req - callback reqeest object
```
graduta: `request`

Framework/Backend/http/server.js
```diff
@@ -34,15 +34,15 @@ class HttpServer {
...
    */
-  constructor(httpConfig, jwtConfig, oAuthConfig = null) {
+  constructor(httpConfig, jwtConfig, connectIdConfig = null) {
     this.app = express();
     this.configureHelmet(httpConfig.hostname);
 
     this.jwt = new JwtToken(jwtConfig);
-    if (oAuthConfig != null) {
-      this.oauth = new OAuth(oAuthConfig);
+    if (connectIdConfig != null) {
```
graduta: `if (connectIdConfig)` should be enough

Framework/Backend/http/server.js
```diff
@@ -242,65 +242,73 @@ class HttpServer {
...
-   * - Prints out an error when code is not valid
-   * The query arguments are serialized and kept in the 'state' parameter through OAuth process
+   * Handles authentication flow (default path of the app: '/')
+   * - If JWT query.token is valid it grants the access to the application
+   * - Redirects to the OpenID flow otherwise, and sets the current state of URL
    * @param {object} req - HTTP request
    * @param {object} res - HTTP response
-   * @param {object} next - serves static paths when OAuth suceeds
-   * @return {object} redirects to OAuth flow or displays the page if JWT token is valid
+   * @param {object} next - serves static paths when OpenId suceeds
```
graduta: `succeeds`

Framework/Backend/http/server.js
```diff
@@ -242,65 +242,73 @@ class HttpServer {
...
+        name: details.name,
+        token: this.jwt.generateToken(details.cern_person_id, details.cern_upn, this.authorise(details)),
+      };
 
-    this.oauth.createTokenAndProvideDetails(code)
-      .then((details) => {
-        query.personid = details.user.personid;
-        query.name = details.user.name;
+      // Read back user params from state
+      const userQuery = JSON.parse(new Buffer(req.query.state, 'base64').toString('ascii'));
```
graduta: `new Buffer` is deprecated and is recommended to use `Buffer.from` or other equivalents.

Framework/Backend/http/server.js
```diff
@@ -242,65 +242,73 @@ class HttpServer {
...
 
-        // Generates JWT token and adds it to the details object
-        query.token = this.jwt.generateToken(details.user.personid, details.user.username, 1);
+      // Concatanates with predefined files and user query
```
graduta: `Concatenates`

Framework/Backend/http/server.js
```diff
@@ -242,65 +242,73 @@ class HttpServer {
...
+      Object.assign(query, this.templateData);
+      Object.assign(query, userQuery);
 
-        // Concatanates details from oAuth flow with data directly passed by user
-        Object.assign(query, this.templateData);
+      res.redirect(url.format({pathname: '/', query: query}));
+    }).catch((reason) => {
+      log.debug('OpenId failed: ' + reason);
```
graduta: If this is a failure shouldn't we log it as an error?

Framework/README.md
```diff
@@ -34,8 +34,8 @@ The goal of this UI framework is to identify library and framework sets, provide
 - Asynchronous data fetching (Ajax) and bi-directional socket (WebSockets)
 - MVC engine with a "diffing" algorithm
 
-### System requirements
-* `nodejs` >= 8.9.4
+### Backend requirements
+* `nodejs` >= v10.13.0
```
graduta: This will be true only for the newer versions of the framework. Should we specify that we still support `nodejs >= 8.9.4` for the framework up to version 1.9.4?

Framework/Backend/http/openid.js
awegrzyn: In this case if we can't create issuer we should not start server.

Framework/README.md
awegrzyn: I think we want users to use the latest version

PR #475, opened 2020-03-12
[OGUI-519] Add Windows build to Framework workflow

awegrzyn:


.github/workflows/framework.yml
```diff
@@ -37,4 +37,14 @@ jobs:
...
       - run: (cd Framework; npm run coverage )
       - run: (cd Framework; ./node_modules/.bin/nyc report --reporter=text-lcov > coverage.lcov)
-      - run: (cd Framework; bash <(curl -s https://codecov.io/bash) -t ${{ secrets.CODECOV_TOKEN }} -F framework)
\ No newline at end of file
+      - run: (cd Framework; bash <(curl -s https://codecov.io/bash) -t ${{ secrets.CODECOV_TOKEN }} -F framework)
+  windows:
+    name: Test on windows-latest
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
```
graduta: Any particular differences between `actions/checkout@v1` (used for the rest of the workflows and `v2`?

awegrzyn: For our use case it does not matter, both work just fine

PR #483, opened 2020-03-17
[OGUI-503] Add tooltip texts to Configuration panel

graduta:


awegrzyn:
Some of the labels contain list of options which are already available in drop-down select box, in such cases I'd either change it to provide actual explanation or just remove it

graduta:
> Some of the labels contain list of options which are already available in drop-down select box, in such cases I'd either change it to provide actual explanation or just remove it

Updated so that tooltips are not displayed if there is no useful information

PR #491, opened 2020-04-02
[OGUI-367] Add checkers view

graduta:
ObjectView Page:
image

LayoutShow Page:
image

LayoutShow Page - small scrollable and wrappable panel
image

Edit Layout Sidebar Tree:
image

Object Tree Page:
image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -12,25 +12,25 @@ coverage:
         flags:
           - framework 
         target: 50%
-        threshold: null
+        threshold: 0
         if_not_found: success
       QualityControl:
         flags:
           - qualitycontrol
         target: 60%
-        threshold: null
+        threshold: 0
         if_not_found: success
       InfoLogger:
         flags:
           - infologger
         target: 70%
-        threshold: null
+        threshold: 0
         if_not_found: success
       Control:
         flags:
           - control
         target: 69%
-        threshold: null
+        threshold: 0
         if_not_found: success
     patch: false
     changes: false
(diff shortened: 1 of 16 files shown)


graduta:
Coverage is still using old targets rather than the current one and when opening the report from codecov it will say coverage dropped

This is due to the fact that it compares it to the percent on `dev` which tests all projects.

Barthelemy:
Very nice !
2 small comments on the lists (inputs and metadata) : 
- Could you remove the [ ] ? 
- How does it look like if there are many elements ? 

Thank you for this nice development.

graduta:
> Very nice !
> 2 small comments on the lists (inputs and metadata) :
> 
>     * Could you remove the [ ] ?
> 
>     * How does it look like if there are many elements ?
> 
> 
> Thank you for this nice development.
Hi, @Barthelemy ,
* Yes, the square brackets can be removed. I did not know exactly what type of content can be expected so I added parsers for Arrays and JSONs
* What do you mean but "many elements"? Many checkers, many elements in the checkers or many elements in the array of inputs of the checker?

Barthelemy:
* all of them
* many inputs or many user metadata

graduta:
>     * all of them
> 
>     * many inputs or many user metadata

In both cases ,many inputs or metadara, they will grow as a list, one element per line. If they are longer than the curent height of the element the panel will transform in a scrollable panel as per:
* LayoutShow Page - small scrollable and wrappable panel
image

Barthelemy:
ok, thank you. 
So let's remove the brackets for the time being. Piotr and myself were wondering if we could dynamically decrease the font if there are too many things in there but it would not scale. So scrolling is good.

graduta:
Removed the square brackets and attached example of objectTree page with a large list of inputs:
image

graduta:
> ok, thank you.
> So let's remove the brackets for the time being. Piotr and myself were wondering if we could dynamically decrease the font if there are too many things in there but it would not scale. So scrolling is good.

Padding between the elements is adjusted based on the location(size of panel differs). Font can be adjusted in the same way based on location rather than size of list of inputs

Barthelemy:
It is good like that, thanks a lot !

PR #502, opened 2020-04-04
Added listen, close and address method

mvegter:
This change allows for more freedom over starting/stopping of the server. It is however breaking for any applications currently using WebUI!

Reason for this change is that it allows us to configure the server before starting it. This also allows us to wait for the database to be up and running before starting the web server. Simplifies graceful shutdown as we do not need to access the underlying server layer. It also simplifies the testing with supertest.

### DIFF
--- a/Framework/Backend/test/mocha-http.js
+++ b/Framework/Backend/test/mocha-http.js
@@ -244,8 +244,8 @@ describe('REST API', () => {
 });
 
 describe('HTTP server', () => {
-  after(() => {
-    httpServer.getServer.close();
+  after(async () => {
+    await httpServer.close();
   });
 
   it('Add and verify custom static path', (done) => {
--- a/Framework/Backend/test/mocha-http.js
+++ b/Framework/Backend/test/mocha-http.js
@@ -13,9 +13,8 @@
 
 process.env.NODE_ENV = 'test';
 
-const assert = require('assert');
+const request = require('supertest');
 const path = require('path');
-const http = require('http');
 const url = require('url');
 const config = require('./../config-default.json');
 const JwtToken = require('./../jwt/token.js');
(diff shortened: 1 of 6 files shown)


awegrzyn:
Could you elaborate what's the advantage of using `supertest`? We're trying to keep the dependency list (even dev) as small as possible.

mvegter:
@awegrzyn I have always used supertest in combination with Express for testing. It hides/simplifies a lot of logic as can be seen by the before/after of the test commit. It is not a necessary / related change more of an improvement.

mvegter:
@awegrzyn any feedback regarding this pull request? Should I drop the `supertest` commit?

awegrzyn:
Hi,
Sorry for a late reply, we had discussion about it and I think we'll not be able to accept this PR in this form:
- address() -> OK
- close () -> OK
- listen() - at first it breaks all 3 currents UIs therefore it'd need to land in new major version (3.x), in addition our current policy about external services (like databases) it that they should not interfere with UI. If DB is not connected there should be appropriate message in UI instead of having UI down. In addition if a service goes down the UI should stay up, and reconnect to the services as soon as it's back. This comes from our experience from Run 2. If you really need this we'd merge it to 3.x release branch (all breaking changes would land in 3.x)
-  `supertest` - I see advantages of having simplified test syntax, and I guess you'll have a lot of such test so after all this could be useful. @graduta are you OK with this?

mvegter:
We could introduce a feature/config flag which overrides the default behaviour of auto listening, this way it should be compatible for versions 2.X and can later be changed to default in 3.X

graduta:
Hi, both. 
Thank you, @mvegter for your contributions. 

I agree with Adam regarding keeping the current behavior due to the fact that we already have 3 projects built on `web-ui`.

Moreover, I like the idea of using a flag to allow the developer to select the mode for the server.
I had a quick look on your changes and I can see you opted as a default to listen. I would like to request as default to keep the current behavior of not waiting on any external connection.

Framework/Backend/test/mocha-http.js
```diff
@@ -40,222 +39,129 @@ describe('REST API', () => {
...
+      .expect(200)
+      .expect({ok: 1}, done);
   });
 });
 
 describe('HTTP server', () => {
-  after(() => {
-    httpServer.getServer.close();
+  after(async () => {
+    await httpServer.close();
```
graduta, line 157: As we are going with the flag approach, there is no need to for changes in the `after` section

Framework/Backend/test/mocha-ws.js
```diff
@@ -163,8 +163,8 @@ describe('websocket', () => {
     });
   });
 
-  after(() => {
+  after(async () => {
     ws.shutdown();
-    http.getServer.close();
+    await http.close();
```
graduta, line 168: Same here as above

Framework/docs/guide/http-server.md
```diff
@@ -19,12 +19,22 @@ Where:
      * [`HTTPS_PORT`] - HTTPS port number, TLS must be enabled
      * [`TLS_KEY`] - private key filepath, TLS must be enabled
      * [`TLS_CERT`] - certificate filepath, TLS must be enabled
+     * [`AUTO_LISTEN`] - flag that enables/disables automatic listening (default: `true`)
```
graduta, line 22: Default for `AUTO_LISTEN` should be false

mvegter:
@graduta was the approval including the last commit or without, as it seems that the review comments you left disappeared.

graduta:
> @graduta was the approval including the last commit or without, as it seems that the review comments you left disappeared.

Approval is for both cases now. I was confused by the name `autolisten` which made me think of a opposite behavior initially, apologies for that. 

The only bit that was an optional change was the log message which you removed in the last commit. 

Looks good to me

mvegter:
@graduta triggering GitHub actions to do all jobs again "solved" the failing jobs, apparently some test instability.

PR #503, opened 2020-04-04
404 on the api router should be a JSON respone

mvegter:


awegrzyn:
I'm just wondering whether it does not make more sense to add this handler to `this.router` and `this.routerPublic` instead of `this.app.use('/api', ...` but I guess final effect will be the same.

mvegter:
@awegrzyn this implementations keeps the behaviour the same (right or wrong 😉), calling a route of `/api` without a token results in a 403 response if it exists or not. Calling an unknown route with token results in a 404, but in HTML.

Using `this.router.use(...)` instead of `this.app.use('/api', ...)` results in 404 on all api routes, I'm not sure why.

PR #510, opened 2020-04-15
[OGUI-530] Add flp selector panel

graduta:


graduta:
> Looks good, the only change I'd propose is to move it above "Environment variables" as I think Vasco mentioned that env vars should be allowed to overwrite FLP selection (please verify).
> ...and this would "impose" the hierarchy

I will move it on top, but right now I am not allowing one selection to override the other. If the user happens to both "Select" an FLP and also "Add" a key for hosts than an error message will be shown that it cannot do both

PR #511, opened 2020-04-15
[OGUI-534] Support multiple express middlewares per route

mvegter:
Currently the framework does not support middleware, this is **one** implementation to support it.

### DIFF
--- a/Framework/Backend/http/server.js
+++ b/Framework/Backend/http/server.js
@@ -255,12 +255,7 @@ class HttpServer {
    * @param {boolean} [options.public] - true to remove token verification
    */
   get(path, callback, options = {}) {
-    if (options.public) {
-      this.routerPublic.get(path, callback);
-      return;
-    }
-
-    this.router.get(path, callback);
+    this._all('get', path, callback, options);
   }
 
   /**
@@ -275,12 +270,7 @@ class HttpServer {
    * @param {boolean} [options.public] - true to remove token verification
    */
   post(path, callback, options = {}) {
-    if (options.public) {
-      this.routerPublic.post(path, callback);
-      return;
-    }
-
-    this.router.post(path, callback);
+    this._all('post', path, callback, options);
   }
 
   /**
@@ -295,12 +285,7 @@ class HttpServer {
    * @param {boolean} [options.public] - true to remove token verification
    */
   put(path, callback, options = {}) {
-    if (options.public) {
-      this.routerPublic.put(path, callback);
-      return;
-    }
-
-    this.router.put(path, callback);
+    this._all('put', path, callback, options);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


awegrzyn:
Hi,
Could you just elaborate which problem you're trying to solve in Jiskefet backend?

mvegter:
For example a middleware for validating request bodies and transforming them to their respective DTO.

awegrzyn:
Sure, but the only use-case of having more than 1 middleware on specific route I can imagine is when you want to use third party middleware, is that the case?

mvegter:
Being able to use more than one callback (or middleware), regardless of first- or third-party, allows for smaller units of logic. Which increases the modularity and testability of the application.

awegrzyn:
Sure I agree, but what's stops you from calling multiple methods from single middleware? Using multiple middleware seems like overkill, but correct me if I'm wrong.
I'm not against, I'm just trying to understand.

mvegter:
Using multiple middleware like this is the convention. Nothing prevents us from calling methods within a single middleware, but that is not the convention nor does it help with isolated units of logic.

awegrzyn:
Okay, in such case we'll add this, could you just provide reference to "the convention"?

mvegter:
https://expressjs.com/en/guide/using-middleware.html

One of the Application-level middleware:
```js
function logOriginalUrl (req, res, next) {
  console.log('Request URL:', req.originalUrl)
  next()
}

function logMethod (req, res, next) {
  console.log('Request Type:', req.method)
  next()
}

var logStuff = [logOriginalUrl, logMethod]
app.get('/user/:id', logStuff, function (req, res, next) {
  res.send('User Info')
})
```

awegrzyn:
I understand that such feature exists but it definitely does not mean that this is convention.
In addition there's another syntax that allows doing the same without adding any changes to the framework:
```js
http.get('/test', (request, response, next) => {
  console.log('middleware 1');
  next();
}, { public: true }); 

http.get('/test', (request, response) => {
  console.log('middleware 2');
  response.send('ok');
}, { public: true }); 
```

mvegter:
Well yes, you can define the route multiple times for each desired callback. And you can define these callbacks as stand alone function and pass the object. Still, this will clutter the source file and as shown in your example you will have to define the exact same options else you will get some nasty side effects. Your provided example is in my opinion workable but not maintainable.

awegrzyn:
Okay, I'm just wondering how many routes you foresee to have more than 2 middlewares? I can imagine that you may need to set additional middleware per entire router, but multiple different middlewares per each route does not seem reasonable.
That's why my first questions was what is your requirement.

mvegter:
Not sure what the relevance is of the number of middleware. Being able to support one or multiple can be done with the same implementation. At this moment no middleware is supported. With this implementation you can have zero or more middleware. So I'm not sure if I understand your question, but yes realistically most of the times you will use one middleware (and a route handler) and on occasion two/three.

PR #524, opened 2020-05-09
[OGUI-555] Update release documentation

graduta:


awegrzyn:
Is it necessary to remove `.github/workflows/framework.yml` from path? I think we want the job to run when workflow changes.

PR #535, opened 2020-05-29
[OGUI-572] Add prefix to consul query

graduta:


awegrzyn:
Could you add changes to README as well?

PR #537, opened 2020-05-30
[OGUI-575] Include a message why the JWT failed

mvegter:
The current JWT error response does not explain to the end user why their request has failed. The proposed implementation uses the error message of the JWT package to atleast give a hint.

#### Before
```json
{
  "message": "JsonWebTokenError"
}
```

#### After
```json
{
  "error": "403 - Json Web Token Error",
  "message": "You must provide a JWT token"
}
```

#### Before
```json
{
  "message": "JsonWebTokenError"
}
```

#### After
```json
{
  "error": "403 - Json Web Token Error",
  "message": "Invalid JWT token provided"
}
```

#### I DON'T have JIRA ticket
- [X] test are added

- [ ] branch and/or PR name(s) includes JIRA ID
- [ ] PR labels are selected

### DIFF
--- a/Framework/Backend/http/server.js
+++ b/Framework/Backend/http/server.js
@@ -470,9 +470,22 @@ class HttpServer {
           name: data.username
         };
         next();
-      }, (error) => {
-        log.warn(`${error.name} : ${error.message}`);
-        res.status(403).json({message: error.name});
+      }, ({name, message}) => {
+        log.warn(`${name} : ${message}`);
+
+        const response = {error: '403 - Json Web Token Error'};
+
+        // Allow for a custom message for known error messages
+        switch (message) {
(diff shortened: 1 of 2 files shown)


mvegter:
@awegrzyn I agree with you on the fact that this is a too simple solution. I dont think hardcoding one message is sufficient as an invalid token or missing token are general cases which should be covered. I did find more info on the errors from the documentation so we could hard-code custom messages for the general errors and those not covered return a generic failed response. I will have a look at improving it later on.

Reference: https://www.npmjs.com/package/jsonwebtoken#errors--codes

awegrzyn:
Okay thanks, I understand, keep in mind we keep full error messages on the server side as logs.

PR #544, opened 2020-06-15
[OGUI-577] Use set for Consul Keys

graduta:
In order to prevent potential issues when Ansible creates keys in Consul KV Store with empty values, we should use a set to remove duplicate keys


awegrzyn:
This has not already been fixed by Kostas in Ansible role?

graduta:
> This has not already been fixed by Kostas in Ansible role?

I fixed it in Ansible by ensuring that a duplicated key is not created for that task. 

In the future that prefix of the key will be heavily changed with multiple new additions and there is no mechanism too prevent adding keys with no values. 

I figured is better to add a check on the UI as well

PR #550, opened 2020-06-23
[OGUI-587] tackle lgtm warnings

MichaelaBrezinova:
- [x] PR labels are selected

### DIFF
--- a/InfoLogger/lib/api.js
+++ b/InfoLogger/lib/api.js
@@ -164,9 +164,13 @@ module.exports.attachTo = (http, ws) => {
    * @param {Error} error
    * @param {number} status
    */
-  function handleError(res, error) {
+  function handleError(res, error, status) {
     log.trace(error);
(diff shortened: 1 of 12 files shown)


graduta:
It seems that your changes from the other PR #549 are added as new in this PR. 
Did you create the branch from an updated dev one?

InfoLogger/lib/api.js
```diff
@@ -164,9 +164,13 @@ module.exports.attachTo = (http, ws) => {
    * @param {Error} error
    * @param {number} status
    */
-  function handleError(res, error) {
+  function handleError(res, error, status) {
```
graduta, line 167: Instead of using the if condition, you can mention a default value for status in the function header and it will automatically assign the value if the function call is not passing a parameter
```
  function handleError(res, error, status = 500) {
```

MichaelaBrezinova: Fixed it.

PR #555, opened 2020-06-29
Free console resources even if the render failed

mvegter:
Free console resources even if the render throws an error. Currently, the timer will keep running on the background, calling `console.time('render');` won't restart the timer but display a warning (and thus keeps the existing time). This fix will always stop and clear the timer.

#### I DON'T have JIRA ticket
- [ ] test are added


awegrzyn:
@graduta please merge if this is fine with you

PR #562, opened 2020-07-14
Use new, simpler, API (QC-368)

Barthelemy:
#### I DON'T have JIRA ticket
- [ ] explain what this PR does
- [ ] test are added

- [ ] PR labels are selected

### DIFF
--- a/QualityControl/tobject2json.cc
+++ b/QualityControl/tobject2json.cc
@@ -19,8 +19,7 @@ public:
 protected:
   void Execute() override
   {
-    const auto slashIndex = path.find_first_of('/');
-    output = BackendInstance->retrieveMOJson(path.substr(0, slashIndex), path.substr(slashIndex + 1));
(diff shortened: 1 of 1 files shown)


Barthelemy:
The former interface is still available in case of problems.

graduta:
@Barthelemy Could you please link in the description to the PR on the QualityControl side containing related changes and also to alidist PR with the version bump

Barthelemy:
The change in QC has happened in April and thus the versions of QC and QCG do not need to be synchronized. In other words, QC v1.0 already contains the necessary changes. 

FYI it was part of a massive overhaul of the CCDB API and storage format (https://alice.its.cern.ch/jira/browse/QC-239).

graduta:
Will keep this PR open for a bit more. This is due to the fact that we are currently investigating the possibility of requesting objects data and files directly from CCDB

PR #563, opened 2020-07-15
[OGUI-421] Add JIRA service to framework

awegrzyn:
- [x] PR labels are selected

### DIFF
--- a/Framework/Backend/services/jira.js
+++ b/Framework/Backend/services/jira.js
@@ -1,0 +1,87 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
+ * distributed under the terms of the GNU General Public License v3 (GPL
+ * Version 3), copied verbatim in the file "COPYING".
+ *
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+const http = require('http');
+const querystring = require('querystring');
+
+/**
+ * Handles creating JIRA issues
+ */
+class Jira {
+
+  constructor(config) {
+    this.hostname = 'alice.its.cern.ch'
+    this.path = '/jira/rest/api/2/issue'
+    this.auth = 'user:pass'
+  }
+
+  async createIssue() {
+    const postData = querystring.stringify({
+    "fields": {
+      "issuetype": {
+        "id": "1"
+      },
+      "project": {
+        "id": "11400"
+      },
+      "description": "Order entry fails when selecting supplier.",
+        "reporter": {
+        "name": "awegrzyn"
+      },
+      "assignee": {
+        "name": "awegrzyn"
+      }
+    }
+    });
+
+    const requestOptions = {
+      hostname: this.hostname,
+      port: 80,
+      path: this.path,
+      method: 'POST',
+      auth: this.auth,
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+        'Content-Length': Buffer.byteLength(postData)
+      }
+    };
+    return new Promise((resolve, reject) => {
+      const requestHandler = (response) => {
+        if (response.statusCode < 200 || response.statusCode > 299) {
+          reject(new Error('Non-2xx status code: ' + response.statusCode));
+          return;
+        }
+        const bodyChunks = [];
+        response.on('data', (chunk) => bodyChunks.push(chunk));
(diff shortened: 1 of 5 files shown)


Framework/Backend/services/jira.js
```diff
@@ -0,0 +1,103 @@
...
+    const requestOptions = {
+      method: 'POST',
+      auth: this.accountUser + ':' + this.accountPass,
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+        'Content-Length': Buffer.byteLength(postData)
+      }
+    };
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
awegrzyn: Not sure if I understood reason behind it, could you explain?

Framework/Backend/services/jira.js
```diff
@@ -0,0 +1,121 @@
...
+    if (!config) {
+      throw new Error('Configuration object cannot be empty');
+    }
+    if (!config.url) {
+      throw new Error('JIRA URL must be defined');
+    }
+    if (!config.serviceAccount || !config.serviceAccount.user || !config.serviceAccount.pass) {
+      throw new Error('Service account for JIRA must be defined');
+    }
+    if (config.projectId) {
```
graduta, line 34: I believe here the condition should be negated. To prevent this in the future, it would be useful to have a test covering this case

Framework/Backend/services/jira.js
```diff
@@ -0,0 +1,121 @@
...
+   * @param {string} reporter - reporter of issue as NICE login
+   * @param {string} assignee - asignee of issue as NICE login
+   * @param {string} summary - title of issue
+   * @param {string} description - issue description (optional)
+   * @return {Promise} - resolve object contains issue ID (id) , key (key) and URL (self)
+   */
+  async createBugIssue(reporter, assignee, summary, description = '') {
+    if (!reporter || !assignee || !summary) {
+      log.warn('Creating bug issue failed: undefined arguments');
+      return Promise.reject(new Error('Creating JIRA issue failed'));
```
graduta, line 97: As we know why the creation of the ticket failed, it would be helpful for the user to receive a more helpful message such as "Invalid parameters". Otherwise we cannot tell if the problem is on our side or maybe the jira instance is not up and running

Framework/Backend/services/jira.js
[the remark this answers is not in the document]
graduta: I was suggesting that the functionality related to the HTTP request can be extracted in a separate function so that in the future it can be reused by a potential `getTickets` method or something similar.

Framework/Backend/test/mocha-jira.js
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  describe('Check creating bug issue', function() {
+    const jira = new Jira({url: 'http://localhost/jira/rest/api/2/issue', serviceAccount: {user: 'test', pass: 'test'}, projectId: 0});
+
+    it('should successfully create a ticket', () => {
+      nock('https://localhost')
+        .post('/jira/rest/api/2/issue')
+        .reply(200, '{"key":"OPRO-123", "self":"https://localhost/jira/OPRO-123", "id":1234}');
+      return jira.createBugIssue('alice', 'bob', 'Run fails').then((res) => {assert.deepStrictEqual(res.key, 'OPRO-123')});
```
graduta, line 61: For string comparisons it is recommended to use "strictEqual" rather than "deepStrictEqual"

Framework/Backend/test/mocha-jira.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    it('should reject with error if is unable to parse response', async () => {
+      nock('https://localhost')
+        .post('/jira/rest/api/2/issue')
+        .reply(200, 'Not a JSON resposne');
```
graduta, line 67: Small typo in `response`

Framework/Backend/test/mocha-jira.js
awegrzyn: Ok, if so it should be changed in many other places

graduta: Yes, I have been updating them along the way every time I was encountering them

Framework/Backend/test/mocha-jira.js
```diff
@@ -0,0 +1,95 @@
...
+ * or submit itself to any jurisdiction.
+*/
+
+const Jira = require('./../services/jira.js');
+const assert = require('assert');
+const nock = require('nock');
+
+describe('JIRA service test suite', function() {
+  before(nock.activate);
+  describe('Check Initialization of ConsulService', function() {
```
graduta, line 21: JiraService instead of ConsulService

PR #591, opened 2020-08-25
[OGUI-330] navigate object timestamp history

graduta:
- [x] PR labels are selected

**Acceptance Criteria:**
History of a timestamp feature is present on pages `objectTree` and `objectView`.  
For both pages:
* Select an object to display (e.g  `qc/TST/External-1/hello_0`)
* Click `i` (info) button to see last modified timestamp
* Ensure last modified timestamp from info panel is the same as the timestamp from the dropdown
* Select a different timestamp (preferably a few hours older so that plot difference it's easily seen) from the dropdown (bottom in the `objectTree` page and top for `objectView`)
* Ensure plot changed
* Ensure  info panel still displays the correct last modified timestamp. It  should not change
* Select a different object (this will request latest information of the object)
* Ensure last modified timestamp from info panel is the same as the timestamp from the dropdown

### DIFF
--- a/QualityControl/lib/CCDBConnector.js
+++ b/QualityControl/lib/CCDBConnector.js
@@ -1,5 +1,5 @@
 const http = require('http');
-const log = new (require('@aliceo2/web-ui').Log)('QCG-CCDBConnector');
+const log = new (require('@aliceo2/web-ui').Log)('CCDBConnector');
 
 /**
  * Gateway for all CCDB calls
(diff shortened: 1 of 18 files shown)


awegrzyn:
How can I test it using `ccdb-test` ?

graduta:
> How can I test it using `ccdb-test` ?

I updated the description with the correct information. Sorry for the confusion

awegrzyn (approved):
Tested, works very well.
The only minor think I'd change is date format, currently it is US style "mm/dd/yyyy" I'd rather use "dd/mm/yyyy". Also using 24-hours instead of AM/PM switch. @Barthelemy what do you think?

graduta:
> Tested, works very well.
> The only minor think I'd change is date format, currently it is US style "mm/dd/yyyy" I'd rather use "dd/mm/yyyy". Also using 24-hours instead of AM/PM switch. @Barthelemy what do you think?

The date format is using `locale` which is taking the format from the user's browser as I was not sure about the users regions

Barthelemy:
I am in favour of using a reasonible date format such as European or Internet one rather than US style. However it seems that it is currently decided on the user's side which is even better !

graduta:
> I am in favour of using a reasonible date format such as European or Internet one rather than US style. However it seems that it is currently decided on the user's side which is even better !

The problem with basing ourselves on the browser of the user is that by default, browsers are set in US Style.

Barthelemy:
I see, I did not realize that. I would then use the European format ("dd/mm/yyyy", 24h time format).

graduta:
> I see, I did not realize that. I would then use the European format ("dd/mm/yyyy", 24h time format).

Sure thing, I will add that

PR #597, opened 2020-09-07
[OGUI-659] Add Trigger radio buttons for env KV panel

graduta:
By default no option will be set for the trigger option: 
default-view

If the user selects "OFF" than the variable `roc_ctp_emulator_enabled` will be added as a key with value `"false"`.
If the user selects "EMU" than the variable `roc_ctp_emulator_enabled` will be added as a key with value `"true"`.

This key will not be displayed to the user and will only be sent to core if the user does not add the key as well in the Advanced Configuration panel. 

If the user both selects an option and add the key than an error will be shown when trying to create the environment. 

Screenshot 2020-09-07 at 16 19 52

### DIFF
--- a/Control/public/workflow/Workflow.js
+++ b/Control/public/workflow/Workflow.js
@@ -41,6 +41,7 @@ export default class Workflow extends Observable {
       revision: 'master',
       template: '',
       variables: {},
+      basicVariables: {},
       hosts: []
     };
 
@@ -185,8 +186,13 @@ export default class Workflow extends Observable {
     const templates = this.templatesMap.payload;
     const repository = this.form.repository;
     const variables = JSON.parse(JSON.stringify(this.form.variables));
-    // Check FLP Selection is not duplicated in vars host
-    if (this.form.variables.hosts && this.form.variables.hosts.length > 0 && this.form.hosts.length > 0) {
+    const sameKeys = Object.keys(this.form.basicVariables).filter((key) => this.form.variables[key]);
+    // Check the user did not introduce items with the same key in Basic Configuration and Advanced Configuration
+    if (sameKeys.length !== 0) {
+      this.model.environment.itemNew =
+        RemoteData.failure(`Due to Basic Configuration selection, you cannot use the following keys: ${sameKeys}`);
+    } else if (this.form.variables.hosts && this.form.variables.hosts.length > 0 && this.form.hosts.length > 0) {
+      // Check FLP Selection is not duplicated in vars host
       this.model.environment.itemNew =
         RemoteData.failure('Selecting FLPs and adding an environment variable with key `hosts` is not possible');
     } else {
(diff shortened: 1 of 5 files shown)


vascobarroso (changes requested):
Looks good! 
The only thing I'd change at this time is the titles of the sections "Configuration of environment" and "Environment variables". Both are used to configured the environment, the only difference is that one is easier to use than the other. Maybe something like: 

"Configuration of environment" => "Basic Configuration"
"Environment variables" => "Advanced configuration"

awegrzyn:
Do you think it's needed to display `roc_ctp_emulator_enabled` variable under `Environment variables` section?
When selecting FLP hostnames the variable `hostnames` does not appear in the table.

graduta:
> Do you think it's needed to display `roc_ctp_emulator_enabled` variable under `Environment variables` section?
> When selecting FLP hostnames the variable `hostnames` does not appear in the table.

I added it because compared to `hosts`, `roc_ctp_emulator_enabled` is not that clear what it refers to. Especially as option "EMU" means "ON" and in the future we will have other options for the trigger which when enabled will modify other variables. (Might be wrong here, @vascobarroso knows more here)

vascobarroso:
In the future we will indeed need to change the variable mapping for this specific case. What I expect is something like this (names might vary slightly): 

- `OFF`: `roc_ctp_emulator_enabled` = `false`, `ltu_enabled` = `false`
- `EMU`: `roc_ctp_emulator_enabled` = `true`, `ltu_enabled` = `false`
- `LTU`: `roc_ctp_emulator_enabled` = `false`, `ltu_enabled` = `true`

Or we could simply introduce a trigger variable: `trigger_mode` = `off` | `emu` | `ltu`.   
Either way, as soon as we introduce dedicated panels there's no really need for the end users to still use variables. 

The only use case where we might add a variable for an existing panel is to test new features in a workflow. But that's mainly for us, and we can easily leave the panel empty and insert the variable manually.

So I'm more inclined into **not** displaying the variable as suggested by Adam.

graduta:
> In the future we will indeed need to change the variable mapping for this specific case. What I expect is something like this (names might vary slightly):
> 
>     * `OFF`: `roc_ctp_emulator_enabled` = `false`, `ltu_enabled` = `false`
> 
>     * `EMU`: `roc_ctp_emulator_enabled` = `true`, `ltu_enabled` = `false`
> 
>     * `LTU`: `roc_ctp_emulator_enabled` = `false`, `ltu_enabled` = `true`
> 
> 
> Or we could simply introduce a trigger variable: `trigger_mode` = `off` | `emu` | `ltu`.
> Either way, as soon as we introduce dedicated panels there's no really need for the end users to still use variables.
> 
> The only use case where we might add a variable for an existing panel is to test new features in a workflow. But that's mainly for us, and we can easily leave the panel empty and insert the variable manually.
> 
> So I'm more inclined into **not** displaying the variable as suggested by Adam.

Ok, I updated the description with the new requirements and made the changes

vascobarroso (approved):

PR #603, opened 2020-09-09
[OGUI-661] Add Readout URI input field

graduta:
* Adds a text input field in the Basic Configuration panel so that the user can easily add the path to the readout configuration file. This field will send to core the value under the key `readout_cfg_uri`
* Changes name of "Templates" to "Workflow" Panel
* Bumps patch version

### DIFF
--- a/Control/package-lock.json
+++ b/Control/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@aliceo2/control",
-  "version": "1.6.8",
+  "version": "1.6.9",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/Control/package.json
+++ b/Control/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aliceo2/control",
-  "version": "1.6.8",
+  "version": "1.6.9",
   "description": "ALICE O2 Control GUI",
   "author": "Adam Wegrzynek",
   "contributors": [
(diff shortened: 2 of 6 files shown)


awegrzyn (commented):
URI syntax requires to have scheme specified (eg.: `file://`) ? If scheme does not need to be specified I'd call it `Readout (file) path`.

graduta:
> URI syntax requires to have scheme specified (eg.: `file://`) ? If scheme does not need to be specified I'd call it `Readout (file) path`.

I believe it has to be in that format but I cannot tell only from the documentation: https://github.com/AliceO2Group/ControlWorkflows 

@vascobarroso would you happen to know more about this?

awegrzyn:
Ok, in such case maybe you could hardcode `file://` before input box? It would be more straightforward.

vascobarroso:
It doesn't need to be a path to a file, it can be a Consul endpoint (`consul://path/to/config`). Therefore, I think `Readout URI` or maybe `Readout Config URI` are good labels.

vascobarroso:
> Ok, in such case maybe you could hardcode `file://` before input box? It would be more straightforward.

That's probably not a bad idea, but then it should be a dropdown to select between `file://` and `consul://`.

awegrzyn:
All right, so in such case we either leave it as it is or add a drop down menu with `file://` and `consul://`

graduta:
> All right, so in such case we either leave it as it is or add a drop down menu with `file://` and `consul://`
Added a dropdown to allow the user to specify the type of path: `consul:\\` or `file:\\`
If no path will be provided in the input text field than no variable will be sent to AliECS
image

awegrzyn (approved):

PR #680, opened 2021-02-01
[OGUI-665] Add button to each task to redirect to Mesos

graduta:
- [x] PR labels are selected

PR which:
* Environment Details page:
  * brings `Locked` column in line with `Tasks` page where icons are used to show the state;
  * Groups the tasks by FLP and adds log button
* Tasks page
  * adds log button

﻿
Screenshot 2021-02-04 at 09 13 33

### DIFF
--- a/Control/public/app.css
+++ b/Control/public/app.css
@@ -22,7 +22,7 @@
 
 .grafana-font { font-family: Roboto, Helvetica Neue, Arial, sans-serif; font-size: 12px; line-height: 1.5; color: #52545c; font-weight: 515; text-align: center; }
 
-tr:hover td { background-color: var(--color-gray-light) }
+/* tr:hover td { background-color: var(--color-gray-light) } */
 
 .actionable-row:hover { cursor: pointer; }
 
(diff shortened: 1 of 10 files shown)


awegrzyn (changes requested):
If I'm not mistaken all URLs use TaskID of first task

graduta:
> If I'm not mistaken all URLs use TaskID of first task

URL is used directly from `core` without any change. I checked now and indeed all 3 of them have the same mesos URL so if they should be different, we will need to look at `core`

Moreover, the 3 tasks have the following IDs:
* `2P2WT341zwE`
* `2P2WT341KyN`
* `2P2WT341fTJ`

The mesos built URL contains no such IDs. I checked the log file and contains logs for all 3 tasks. Thus, if we want to proceed with this URL than we will need to move the button to the environment level rather than the task level

EDIT: Moving the PR to `draft` until misunderstanding is fixed

graduta:
Following the update of the requirements, PR has been updated accordingly

awegrzyn (approved):

PR #693, opened 2021-02-25
Add DPL Minimal workflow panel

awegrzyn:


awegrzyn:
I guess this may also break integration tests? Do we use div selectors there as well?

graduta:
> I guess this may also break integration tests? Do we use div selectors there as well?

We do but not for the basic panel. During the FLP Integration Tests we use the advanced configuration panel which can be configured through the `test-config.js` file.

graduta (approved):

PR #707, opened 2021-03-12
[OGUI-803] Update facility of the logs

graduta:
- [x] PR labels are selected

### DIFF
--- a/.github/workflows/framework.yml
+++ b/.github/workflows/framework.yml
@@ -82,8 +82,6 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: '12.x'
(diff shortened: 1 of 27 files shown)


awegrzyn (changes requested):
Do we really want to prepend each log message with `[xxx]` ? I'd be more in favour of passing it as an argument when creating `log` instance. In addition `QualityControl` is static text it could be set somewhere when initialising the app.
So my proposal would be:
1. Set `QualityControl` prefix in index file
2. Create logger instance `const log = new (require('@aliceo2/web-ui').Log)('CCDBConnector')`

From above I'd expect log with `system = GUI, and facaility = QualityControl/CCDBConnector` and log message without any prefix.

awegrzyn (approved):

PR #719, opened 2021-04-07
[OGUI-823] Add panel in environment page with selected user vars

graduta:
- [x] PR labels are selected

On the environment details page, mimic the basic panel from new environment with radio buttons to be disabled.
image

### DIFF
--- a/Control/public/environment/Environment.js
+++ b/Control/public/environment/Environment.js
@@ -80,7 +80,6 @@ export default class Environment extends Observable {
    * Check if the passed variable is known to belong to radio button group
    */
   isKVPairInConsulUriGroup(key, value) {
-    console.log(['qc_config_uri', 'readout_cfg_uri'].includes(key) && value.includes('consul'))
     return ['qc_config_uri', 'readout_cfg_uri'].includes(key) && value.includes('consul');
   }
 
--- a/Control/public/environment/Environment.js
+++ b/Control/public/environment/Environment.js
@@ -70,6 +70,21 @@ export default class Environment extends Observable {
   }
 
   /**
+   * Check if the passed variable is known to belong to radio button group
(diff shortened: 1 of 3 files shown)


awegrzyn:
Hi,
1. Is there any reason to pollute UI with "disabled" buttons, maybe simple value `ON`, `OFF`, etc would be enough? @vascobarroso ?
2. I think host list could be formatted as well? Eg. just comma separated instead of JSON array dump.

graduta:
> Hi,
> 
>     1. Is there any reason to pollute UI with "disabled" buttons, maybe simple value `ON`, `OFF`, etc would be enough? @vascobarroso ?

The initial proposal was the use of values but after evaluation it was requested to use radio buttons as it is providing a more intuitive and easy to read form;

> 
>     2. I think host list could be formatted as well? Eg. just comma separated instead of JSON array dump.

It is comma separated and parsed compared to the old version of it. Screenshot below
image
I left it on purpose between square brackets so that it can be easily understood core expects a list and copy-paste it if needed. 
I would not split the names on a new line or anything similar as that list can contain potentially 200 items and it will take a lot of space. But I could add a space or two between the items to improve readability

awegrzyn:
> it was requested to use radio buttons as it is providing a more intuitive and easy to read form

I'd agree if we did allow to modify these values at some stage but we do not

> It is comma separated and parsed compared to the old version of it.

Ok, indeed, thanks !

vascobarroso:
Concerning the radio buttons vs value display, what I wanted to avoid was the discrepancy between the labels and values used in the env creation page (e.g `QC: ON`) and in the env details page (e.g. `QC: false`). If we use the same thing in both pages, then indeed the radio buttons might be redundant and can be removed.

PR #727, opened 2021-04-15
[OGUI-829] Pass run variable to Grafana when loading environment monitoring panels

awegrzyn:
- [x] PR labels are selected

### DIFF
--- a/Control/package-lock.json
+++ b/Control/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@aliceo2/control",
-  "version": "1.17.1",
+  "version": "1.17.2",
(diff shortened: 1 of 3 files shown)


graduta (changes requested):

graduta (approved):

graduta:
Hi @awegrzyn ,
As this is a change that needs to be in line with a change in monitoring, thus nothing being broken as of now. 
Should this be a minor version bump rather than a patch?

awegrzyn:
It does not need to be in line, it does no break anything, it's backward and forward compatible.

PR #731, opened 2021-04-19
[O2-2245] Bump NodeJS version

graduta:
- [x] PR labels are selected

### DIFF
--- a/.github/workflows/control.yml
+++ b/.github/workflows/control.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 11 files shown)


awegrzyn:
I'm wonder whether we shall create NodeJS version matrix for Framework? 12 and 14? What do you think?

graduta:
We could for sure add `12.x` but it will only cover the framework unit tests. I updated to `14.x` as it will be supported for another 2 years compared to `12.x` 1 year. 

For the compatibility tests of `QCG`, `COG` and `ILG` we will not be able to use `12.x` as they will use the `rmSync` method from node `14.x`

PR #740, opened 2021-04-30
[OGUI-840] Status of integrated services

graduta:
- [x] PR labels are selected

Adds a new request to Core (`GetIntegratedServices`) to query for the info & status of those. 
* If the call is successful than:
   * each service will have a row with details about its connection.
   * if the value is empty than the service will be displayed as "Service is disabled" with future work to add a button to redirect to consul for enabeling the service.
* If the call is unsuccessful (e.g. Core times out or is down) than GUI does not know which services are part of the integration. Thus, a single row with the error message from core will be displayed.

Screenshot of call successful to CORE with various states of services:
image

Screenshot of call failure to Core; A single table will be added as the GUI is not aware of what services are there:
image


awegrzyn:
Looks very nice, just two small things:
1. I've noticed on Firefox misalignment of boxes, see screenshot below
2. I'd not mark not configured services as "error", stating "This service was not configured" is good enough
Screenshot 2021-05-05 at 09 58 20

graduta:
> Looks very nice, just two small things:
> 
>     1. I've noticed on Firefox misalignment of boxes, see screenshot below
> 
>     2. I'd not mark not configured services as "error", stating "This service was not configured" is good enough
>        Screenshot 2021-05-05 at 09 58 20

1. Indeed there is some extra padding there for AliECS tables. I will removeit
2. I will make the updates

PR #742, opened 2021-05-03
[OGUI-850] Remove end slash if exists

graduta:
- [x] PR labels are selected
- [] FLP integration tests were ran successful

Following `AliECS` changes, worfklow may or may not contain a trailing slash.
Ensure that the URL is properly built in both cases

### DIFF
--- a/Control/public/workflow/Workflow.js
+++ b/Control/public/workflow/Workflow.js
@@ -227,9 +227,9 @@ export default class Workflow extends Observable {
           if (template !== '') {
             let path = '';
             if (revision === '(no-revision-by-default)') {
-              path = repository + 'workflows/' + template;
+              path = this.parseRepository(repository) + `/workflows/${template}`;
             } else {
-              path = repository + 'workflows/' + template + '@' + revision;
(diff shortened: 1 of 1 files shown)


awegrzyn (approved):
Ok for me but IMO this should be handled on the core side.

graduta:
> Ok for me but IMO this should be handled on the core side.

I think it should still be on our side. I might have explained it a bit wrong. Please find below a more detailed explanation. 

This is not a change to remove the `/` at the end of the URL we send to core. 
We build the `workflow-URL` on our side from: `repository`, `worfklow` and `revision`.

Before we were getting from `Core`: `<name_of_repository>/` but now the slash was removed so instead of building a proper URL `repository/revision@workflow` we were sending to core `repositoryrevision@workflow`.

awegrzyn:
Ok thx for explanation, maybe `path.join` or similar could be used?

graduta:
> Ok thx for explanation, maybe `path.join` or similar could be used?

I think `path.join()` is a method from the module in NodeJS `path`. 
The URL selection and editing happens on the client side

PR #768, opened 2021-06-03
[OGUI-872] Add full name to JWT token

awegrzyn:


Framework/Backend/http/server.js
```diff
@@ -417,7 +418,7 @@ class HttpServer {
         personid: details.cern_person_id,
         name: details.name,
         access: this.authorise(details),
-        token: this.jwt.generateToken(details.cern_person_id, details.cern_upn, this.authorise(details)),
+        token: this.jwt.generateToken(details.cern_person_id, details.cern_upn, details.name, this.authorise(details)),
```
graduta, line 421: For compatiblity reasons, wouldn't it be better to add `name` as parameter at the end of the list of existing parameters so that any forgotten uses are covered? Preferably with a default value?

For example we use the `generateToken` method in `server.js` as well which I do not see being updated here

graduta (changes requested):

Framework/Backend/http/server.js
awegrzyn: I see your point but there's already one optional argument in that method, in addition it's difficult to have default value for name.

Regarding `generateToken` method in `server.js`, could you clarify? It looks good to me.

graduta: I do not think there is anything preventing us from setting default values to parameters in method declaration. 

1. While for the `server.js` I am referring to this particular case: https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/http/server.js#L420 

In here, now the method expects the 3rd parameter to be the name but the server will send the authorise code.

2. While in here: https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/http/server.js#L240
We pass a name for the username and in the future I guess we will want to display the name of the user, so I think this line should be updated as well in this PR

graduta (approved):

awegrzyn:
Okay, I've added some docs, could you merge, set version, release before Friday's meeting?

graduta:
> Okay, I've added some docs, could you merge, set version, release before Friday's meeting?

Sure thing

awegrzyn:
So compatibility test fail as I modified username from `Anonymous` to `anonymous`. Shall I revert it? If not, I can't find a place where this needs to be changed.

graduta:
> So compatibility test fail as I modified username from `Anonymous` to `anonymous`. Shall I revert it? If not, I can't find a place where this needs to be changed.

Compatibility tests are ran on the components themselves.
The new `webui` version is installed and than tests are ran. So if you want to find the place to update the tests, it would be in `Control` and `QualityControl`

I think we should proceed as such. It only fails on checking the value is the same `Anonymous` but I think a better test (on components not webui) would be that a string is received in both cases `username` and `name` as defaults can change  in time

graduta:
I will merge, bump and than update the other projects as well

PR #804, opened 2021-07-27
[OGUI-566] Custom configuration panel built via varSpecMap

graduta:
- [x] PR labels are selected

If the workflow template provided by `AliECS Core` contains a field `varSpecMap` than the GUI will attempt to build multiple custom panels based on that information

It can contain 6 different widgets, each with its own properties such as label, description, defaultValue, index, etc. Definition here https://github.com/AliceO2Group/WebUi/blob/dev/Control/protobuf/o2control.proto#L380

The variables will be grouped by the panel they belong to and sorted by the index and if not present alphabetically by their label.


awegrzyn:
Waiting for AliECS release to test it

awegrzyn:
@teo Could you comment in here once core is released?

vascobarroso:
> @teo Could you comment in here once core is released?

@awegrzyn I think @teo just tagged AliECS v0.25.80: https://github.com/alisw/alidist/pull/3198

awegrzyn:
Hi, for the moment just sharing few ideas increasing usability and are not difficult to implement:
- I'd leave initial layout, that users are used to, as it was, meaning on 2nd level I'd keep Basic and Advanced configuration (and probably rename Advanced to Key-Value), below I'd add "Advanced configuration" title and display additional panels
-  Not to overwhelm I'd display all new panels collapsed (only title with grey background is displayed initially), then user can click to display it; optionally, a panel which is visible could have a small "OK"/"Done" button, that would collapse it and changed title background to green, meaning that all options in that panel are set...
- All true/false selection should be presented as toggle button
- When hovering on option name, cursor changes to question mark and additional explanation is given for the option; Instead I'd display question mark icon just after option name, and make it "hoverable" instead of entire name; Moreover, instead of pop-up, there could be dedicated, fixed footer that would display all hints, but also errors coming from the core (this can be useful as now going down requires scrolling)

awegrzyn:
@vascobarroso I think some default parameters from workflows are not up-to-date, and we don't want them to overwrite what's in Consul, eg. see Monitoring, with default params monitoring is `off`. I think they should be reviewed.

vascobarroso:
@awegrzyn I've asked that exact same question to @graduta and @teo this morning :)   
The difficulty is that we cannot put the defaults in the workflow because they are specific to each deployment. 
We need to find a way to add the new functionality without disrupting P2 activities.

PR #806, opened 2021-07-30
Upload release TARballs to S3

awegrzyn:


awegrzyn:
@graduta This should upload all TARballs to our S3 bucket.

PR #870, opened 2021-09-24
Fix winston logging from InfoLoggerSender

awegrzyn:
This unifies `winston` vs `winston.instance` and avoids node crashing when misused.


graduta:
I am not sure I see how this helps. 
Rather than using `winston.instance` we use `winston` but the constructor now receives `winston.instance` so we basically use the same instance

graduta (approved):

PR #872, opened 2021-09-27
[OMON-487] Add hidden option to modify frame-src CSP

awegrzyn:
This is to needed for embedding Grafana plots from different origin.
Documentation has intentionally not been added.


awegrzyn:
I'd need a bump as well, can I release `1.20.1` ?

graduta (approved):

PR #873, opened 2021-09-27
[OMON-487] Embed Grafana plots from different origin

awegrzyn:
This enables embedding Grafana plots from different origin, in configuration `grafana.url` is required instead of `grafana.port`

### DIFF
--- a/Control/index.js
+++ b/Control/index.js
@@ -23,6 +23,7 @@ const api = require('./lib/api.js');
 
 buildPublicConfig(config);
 
+config.http.iframeCsp = (config?.grafana?.url) ? [ config.grafana.url ] : [];
 const http = new HttpServer(config.http, config.jwt, config.openId);
 const ws = new WebSocket(http);
 http.addStaticPath(path.join(__dirname, 'public'));
--- a/Control/lib/config/configProvider.js
+++ b/Control/lib/config/configProvider.js
@@ -42,4 +42,8 @@ const config = require(configFile);
 Log.configure(config);
 log.info(`Read config file "${configFile}"`);
 
+// Add Grafana URL to frame-src CSP
+config.http.iframeCsp = (config?.grafana?.url) ? [ config.grafana.url ] : [];
+log.debug("Setting frame-src CSP ", config.http.iframeCsp);
+
 module.exports = config;
--- a/Control/lib/config/configProvider.js
+++ b/Control/lib/config/configProvider.js
@@ -42,8 +42,4 @@ const config = require(configFile);
 Log.configure(config);
 log.info(`Read config file "${configFile}"`);
 
-// Add Grafana URL to frame-src CSP
-config.http.iframeCsp = (config?.grafana?.url) ? [ config.grafana.url ] : [];
-log.debug("Setting frame-src CSP ", config.http.iframeCsp);
-
 module.exports = config;
--- a/Control/lib/config/publicConfigProvider.js
+++ b/Control/lib/config/publicConfigProvider.js
@@ -68,8 +68,8 @@ function getConsulConfig(config) {
  * @return {JSON}
  */
 function _getGrafanaConfig(config) {
-  if (config?.grafana && config?.http?.hostname && config?.grafana?.port) {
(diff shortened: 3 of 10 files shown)


Control/index.js
```diff
@@ -23,6 +23,7 @@ const api = require('./lib/api.js');
 
 buildPublicConfig(config);
 
+config.http.iframeCsp = (config?.grafana?.url) ? [ config.grafana.url ] : [];
```
graduta, line 26: You can rewrite this by using the nullish coalescing:

`config.http.iframeCsp = config.grafana.url ?? [];`

 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

Control/lib/config/configProvider.js
```diff
@@ -42,4 +42,8 @@ const config = require(configFile);
 Log.configure(config);
 log.info(`Read config file "${configFile}"`);
 
+// Add Grafana URL to frame-src CSP
```
graduta: Why is there a need to add twice the `iframeCsp`? Here and index?

`index.js` is using the configuration provided by this module

graduta (changes requested):

Control/lib/config/configProvider.js
awegrzyn: good catch, I forgot to remove it

Control/index.js
awegrzyn: actually, it might not work as I need array with this element inside

graduta: Indeed

Control/lib/config/publicConfigProvider.js
```diff
@@ -68,8 +68,8 @@ function getConsulConfig(config) {
  * @return {JSON}
  */
 function _getGrafanaConfig(config) {
-  if (config?.grafana && config?.http?.hostname && config?.grafana?.port) {
-    const hostPort = `http://${config.http.hostname}:${config.grafana.port}`;
+  if (config?.grafana && config?.grafana?.url) {
```
graduta: Here there is no need for the first check anymore. There was no need in my previous check either 😄 
`if (config?.grafana?.url)` will do the trick

Control/test/lib/mocha-status-service.js
```diff
@@ -115,34 +115,34 @@ describe('StatusService test suite', () => {
...
       const status = new StatusService(config, {}, {});
       const grafanaStatus = await status.getGrafanaStatus();
-      nock(`http://${config.http.hostname}:${config.grafana.port}`)
+      nock(config.grafana.url)
         .get('/api/health')
         .reply(200, {});
-      expectedInfo.status = {ok: true, configured: true};
-      assert.deepStrictEqual(grafanaStatus, expectedInfo);
+      assert.deepStrictEqual(grafanaStatus.status, {ok: true, configured: true});
+      assert.equal(grafanaStatus.protocol, expectedInfo.protocol);
```
graduta: `equal` is deprecated.
It is better to use `deepStrictEqual` for objects and `strictEqual` for primitive types

Control/test/lib/mocha-status-service.js
```diff
@@ -115,34 +115,34 @@ describe('StatusService test suite', () => {
...
-      assert.deepStrictEqual(grafanaStatus, expectedInfo);
+      assert.deepStrictEqual(grafanaStatus.status, {ok: false, configured: true, message: 'Error: Unable to connect'});
     });
 
     it('should successfully return that grafana was not configured if configuration is not provided', async () => {
       const status = new StatusService({}, {}, {});
       const grafanaStatus = await status.getGrafanaStatus();
-      const expected = {status: {ok: false, configured: false, message: 'This service was not configured'}};
-      assert.deepStrictEqual(grafanaStatus, expected);
+      assert.deepEqual(grafanaStatus.status, {ok: false, configured: false, message: 'This service was not configured'});
```
graduta: `deepEqual` is deprecated.
It is better to use `deepStrictEqual` for objects and `strictEqual` for primitive types

graduta (changes requested):

graduta (approved):

PR #877, opened 2021-09-29
[OGUI-980] Move Kafka connector to framework and use KafkaJS

awegrzyn:
- [x] PR labels are selected

### DIFF
--- a/Framework/Backend/services/kafka.js
+++ b/Framework/Backend/services/kafka.js
@@ -12,7 +12,7 @@
  * or submit itself to any jurisdiction.
 */
 
-const { Kafka } = require('kafkajs')
+const { Kafka, logLevel } = require('kafkajs')
 const WebSocketMessage = require('../websocket/message.js');
 const log = new (require('./../log/Log.js'))(`${process.env.npm_config_log_label ?? 'web-ui'}/kafka`);
 
@@ -25,20 +25,35 @@ class KafkaConnector {
    * @param {object} config Config with list of Kafka brokers
    */
   constructor(config) {
-    if (config?.brokers?.length < 1) {
+    if (!config?.brokers || config.brokers.length < 1) {
       throw new Error(`Kafka broker list was not provided`);
     }
     this.kafka = new Kafka({
       clientId: 'webui',
-      brokers: config.brokers
+      brokers: config.brokers,
+      retry: {retries: 3},
+      logLevel: logLevel.NOTHING
     });
+
+    this.admin = this.kafka.admin();
     this.consumer = null;
     this.webSocket = null;
     log.info('Kafka connector configured');
   }
 
   /**
+   * Provides healthstatus of Kafka cluster
+   */
+  health() {
+    return Promise.resolve()
+      .then(() => this.admin.connect())
+      .then(() => this.admin.disconnect());
+  }
+
+  /**
    * Sends a message to selected topic
+   * @param {string} topic Kafka topic
+   * @param {string} message message to be sent
    */
   _send(topic, message) {
     const producer = this.kafka.producer();
@@ -74,13 +89,13 @@ class KafkaConnector {
   proxyWebNotificationToWs(webSocket) {
     this.webSocket = webSocket;
     this.consumer = this.kafka.consumer({groupId: 'webnotification-group'});
-
+    log.info('Listening for notifications');
     return Promise.resolve()
       .then(() => this.consumer.connect())
       .then(() => this.consumer.subscribe({topic: 'webnotification', fromBeginning: false}))
       .then(() => {
         return this.consumer.run({eachMessage: async ({topic, partition, message}) => {
-          log.debug(`Received message on ${topic} topic from ${partition} partition`)
+          log.debug(`Received message on ${topic} topic from ${partition} partition`);
           this.webSocket.broadcast(
             new WebSocketMessage().setCommand('notification').setPayload(message.value.toString())
           );
--- a/Framework/Backend/services/kafka.js
+++ b/Framework/Backend/services/kafka.js
@@ -82,7 +82,10 @@ class KafkaConnector {
    * @param {url}    url   URL referencing notification
    * @returns {Promise}
    */
-  sendToMattermost(title, body, url = '') {
+  sendToMattermost(title, body = '', url = '') {
+    if (!title || title.length < 3) {
+      throw new Error('Notification title needs to be at least 3 characters long');
+    }
     return this._send('mattermost', JSON.stringify({description: `**${title}**`, client_url: url, details: body}))
   }
 
@@ -92,6 +95,9 @@ class KafkaConnector {
    * @returns {Promise}
    */
   triggerWebNotification(message) {
+    if (!message || message.length < 3) {
+      throw new Error('Notification message needs to be at least 3 characters long');
+    }
     return this._send('webnotification', message);
   }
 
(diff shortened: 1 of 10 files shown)


Framework/Backend/services/kafka.js
```diff
@@ -0,0 +1,127 @@
...
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+const { Kafka, logLevel } = require('kafkajs')
+const WebSocketMessage = require('../websocket/message.js');
+const log = new (require('./../log/Log.js'))(`${process.env.npm_config_log_label ?? 'web-ui'}/kafka`);
```
graduta: I think it would be good to stick to one naming convention for the label of the framework
Here I can see `web-ui` is used while in the rest of the services we have `framework`

Framework/Backend/services/kafka.js
```diff
@@ -0,0 +1,127 @@
...
+   */
+  isConfigured() {
+    return (this.kafka !== undefined && this.admin !== undefined);
+  }
+
+  /**
+   * Provides healthstatus of Kafka cluster
+   */
+  health() {
+    return Promise.resolve()
```
graduta: Could this and rest of the methods be written using `async/await` format?
Example from the `kafkajs` module: https://www.npmjs.com/package/kafkajs#usage

Framework/Backend/services/kafka.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * Sends notification to mattermost channel
+   * @param {string} title Title of notification
+   * @param {string} body  Body of notification
+   * @param {url}    url   URL referencing notification
+   */
+  sendToMattermost(title, body, url = '') {
+    return this._send('mattermost', JSON.stringify({description: `**${title}**`, client_url: url, details: body}))
```
graduta: I think `title` and `body` should have default values as well otherwise the call to `JSON.stringify` will omit the field if it is undefined 
e.g
```
t = undefined

const a = {aa: t}

JSON.stringify(a)
```
This will give `{}`

Framework/Backend/services/kafka.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * Sends message in order to be display in WebUI-based GUI
+   * @param {string} message Notification message
+   */
```
graduta, line 126: Can we add for all methods the return type promise as documented field?

Framework/Backend/test/mocha-consul.service.js
```diff
@@ -14,7 +14,7 @@
 
 /* eslint-disable max-len */
 const ConsulService = require('./../services/consul.service.js');
-const config = require('./test-config.js');
+const config = require('./../config-default.json');
```
graduta, line 17: I do not think this change is done on purpose

Framework/Backend/test/mocha-kafka.js
```diff
@@ -0,0 +1,85 @@
...
+
+    it('should throw error due to missing all mandatory fields in config', () => {
+      assert.throws(() => {
+        new KafkaConnector({});
+      }, new Error('Kafka broker list was not provided'));
+    });
+
+    it('should successfully create a kafka connector', () => {
+      const kafka = new KafkaConnector(config.kafka);
+      assert(kafka.kafka !== undefined);
```
graduta: For such conditions we should use `assert.ok()` instead of just `assert`

Framework/Backend/test/mocha-kafka.js
```diff
@@ -0,0 +1,85 @@
...
+      assert.deepStrictEqual(kafka.isConfigured(), true);
```
graduta: Here it is better if se use ` assert.ok(kafka.isConfigured());`

Framework/Backend/test/mocha-kafka.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+    it('should return false for kafka not being properly configured', () => {
+      const kafka = new KafkaConnector();
+      assert.deepStrictEqual(kafka.isConfigured(), false);
```
graduta: No need for `deepStrictEqual` here. The `ok` or `strictEqual` is enough

Framework/Backend/test/mocha-kafka.js
```diff
@@ -0,0 +1,85 @@
...
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+    const WebSocket = require('./../websocket/server');
+    const HttpServer = require('./../http/server');
+    const JwtToken = require('./../jwt/token.js');
+    const wsClient = require('ws');
+    let wsServer, http, kafka;
+
+    it('should report health status', async () => {
+      const kafka = new KafkaConnector(config.kafka);
+      await kafka.health();
```
graduta: I am not sure this test is validating the result as it is missing the assert statement?

Framework/Backend/test/mocha-kafka.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+    it('should return false for kafka not being properly configured', () => {
+      const kafka = new KafkaConnector();
+      assert.deepStrictEqual(kafka.isConfigured(), false);
+    });
+  });
+
+  /// Remove .skip to actually run tests
+  describe.skip('Check integration with Kafka', () => {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
graduta: Can we put these statements in a `before` block?

graduta (changes requested):

Framework/Backend/test/mocha-kafka.js
awegrzyn: You mean all `require` statements as well? I'd need to keep the list of vars there anyway.

Framework/Backend/services/kafka.js
awegrzyn: Indeed, but in general without `title` and `body` this should fail, shall I return `Promise.reject` in such case?

Framework/Backend/services/kafka.js
awegrzyn: yep definitelly

Framework/Backend/services/kafka.js
graduta: If we want it to fail, than yes, throwing an error is better
And if the `async/await` will be used, the error thrown will be returned automatically as a rejected promise

graduta (approved):

awegrzyn:
I'll make changes to AliECS GUI in separate PR

Framework/Backend/services/kafka.js
```diff
@@ -0,0 +1,175 @@
...
+ * Gateway for all Kafka notification service
+ */
+class KafkaConnector {
+  /**
+   * Prepares connector
+   * @param {object} config Config with list of Kafka brokers
+   */
+  constructor(config) {
+    if (!config) {
+      log.warn('Missing configuration');
```
graduta, line 30: Should we throw an error here as well to be in line with line 33 where if there are missing brokers we throw error?

awegrzyn: In this case we might not want to throw as it represents case when configuration is missing, therefore Kafka should be disabled.

PR #933, opened 2021-11-08
[OGUI-1030] Query logs with LIKE only when wildcard is present in filter

awegrzyn:


graduta (approved):
Following tests for LIKE

PR #941, opened 2021-11-10
[OGUI-1028] Explicitly disable severity scroll on Safari

awegrzyn:
This fixes my issue with scrollbar

### DIFF
--- a/InfoLogger/public/Model.js
+++ b/InfoLogger/public/Model.js
@@ -77,6 +77,10 @@ export default class Model extends Observable {
     // Model can change very often we protect router with callRateLimiter
     // Router limit: 100 calls per 30 seconds max = 30ms, 2 FPS is enough (500ms)
     this.observe(callRateLimiter(this.updateRouteOnModelChange.bind(this), 500));
+
+    this.safariNav = navigator.vendor && navigator.vendor.indexOf('Apple') > -1 &&
+      navigator.userAgent && navigator.userAgent.indexOf('CriOS') == -1 &&
+      navigator.userAgent.indexOf('FxiOS') == -1;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


InfoLogger/public/view.js
```diff
@@ -63,7 +63,7 @@ const logsTable = (model) =>
     // table scrollable content
     h('.flex-row.flex-grow.logs-content', [
       tableLogsContent(model),
-      tableLogsScrollMap(model),
+      !model.safariNav ? tableLogsScrollMap(model) : null
```
graduta, line 66: Rather than returning `null` you can rewrite to only add table when `safariNav` evaluates to false
`!model.safariNav && tableLogsScrollMap(model)`

InfoLogger/public/Model.js
```diff
@@ -77,6 +77,10 @@ export default class Model extends Observable {
     // Model can change very often we protect router with callRateLimiter
     // Router limit: 100 calls per 30 seconds max = 30ms, 2 FPS is enough (500ms)
     this.observe(callRateLimiter(this.updateRouteOnModelChange.bind(this), 500));
+
+    this.safariNav = navigator.vendor && navigator.vendor.indexOf('Apple') > -1 &&
+      navigator.userAgent && navigator.userAgent.indexOf('CriOS') == -1 &&
```
graduta, line 82: As you are checking for `String` in all 3 you can use the `includes` method

graduta (changes requested):
I think a more elegant solution can be obtained.

Rather than hiding the canvas, the issue is that the canvas is placed under the scrollbar on Safari. 
Simply add the `canvas` in a `div` element and it will appear back on Safari

Edit: This solves the missing sidebar issue but the scrollbar is still not clickable

graduta:
Closing in favour of [!942](https://github.com/AliceO2Group/WebUi/pull/942)

PR #1040, opened 2022-01-28
[OGUI-1069] Move notification logic to Kafka consumer

awegrzyn:


Framework/Backend/services/notification.js
```diff
@@ -69,81 +69,25 @@ class KafkaConnector {
    * @param {string} message message to be sent
    * @returns {Promise}
    */
-  async _send(topic, message) {
+  async send(tag, title, url, extra) {
```
graduta, line 72: If the method is to be used within this class only, please keep the naming convention to use `_`

graduta (approved):

Framework/Backend/services/notification.js
awegrzyn: Nope, this is to be used from outside, eg. by Bookkeeping

PR #1074, opened 2022-02-08
[OGUI-1075] Adapt tag parsing to changes in Consul service

awegrzyn:


QualityControl/lib/api.js
```diff
@@ -150,10 +150,9 @@ function errorHandler(err, res, status = 500) {
 function getTagsFromServices(services) {
   const prefix = model.queryPrefix;
   const tags = [];
-  for (const serviceName in services) {
-    if (services[serviceName] && services[serviceName].Tags && services[serviceName].Tags.length > 0) {
-      const tagsToBeAdded = services[serviceName].Tags;
-      tagsToBeAdded.filter((tag) => tag.startsWith(prefix))
+  for (const service in services) {
```
graduta: I would rewrite a bit this part due to:
* now catalog has a simpler structure to use;
* we currently do not check for duplicates
```
const tagsSet = new Set(Object.values(services)
  .filter((service) => service?.length > 0)
  .flat());

const tags = [...tagsSet]
  .filter((tag) => tag?.startsWith(prefix))
  .map((tag) => ({name: tag}));

return tags
```
Or as one assignment only:
```
const tags = [...new Set(Object.values(services)
  .filter((service) => service?.length > 0)
  .flat())
].filter((tag) => tag?.startsWith(prefix))
  .map((tag) => ({name: tag}));

return tags;
```

graduta (changes requested):
Tests in `mocha-consul.service.js` need to be updated. We could use this mocked reply that contains all possible cases:
```
const services = {
    test: ['qc', 'qc/2', 'qc/3'],
    test2: ['qc', 'qc3', 'abc', null, undefined],
    test3: undefined,
    test4: null,
    test5: [],
    test6: 'string'
}
```
which should reply with:
```
[
  { name: 'qc' },
  { name: 'qc/2' },
  { name: 'qc/3' },
  { name: 'qc3' }
]
```

Moreover, this is a change in `Framework` and one in `QC`

Please split the pull request changes in 2
1. For the framework plus a release for it
2. For QC changes plus the bumped version of the `aliceo2/webui` framework

awegrzyn:
Definitely, this is not yet complete.

QualityControl/lib/api.js
awegrzyn: I don't think this gonna work as you're filtering on tags not service names

graduta: Are you referring to the one assignment only?
Basically what is happening is that we create a set which is flattened, containing only `tags` at this point. 
Out of that set we create an array and we filter the tags that do not start with QCPrefix

This was the initial requirement

Now that we have Nomad, we can filter on services as well but for that we do not have a standard naming convention from QC

awegrzyn: I was referring to `filter((tag) => tag?.startsWith(prefix))`, the prefix is for service names not for tags (see my proposed solution).

graduta: Yes, but beforehand the prefix was filtering tags. I believe we should use both because if a user puts something else in a tag, we do not want to send it to QC as QC will only filter objects in the CCDB that start with the prefix. 
So, I believe the filter has to be there for tags especially to check if there are any undefined or null values

QualityControl/lib/api.js
```diff
@@ -149,13 +149,10 @@ function errorHandler(err, res, status = 500) {
...
-      const tagsToBeAdded = services[serviceName].Tags;
-      tagsToBeAdded.filter((tag) => tag.startsWith(prefix))
-        .forEach((tag) => tags.push({name: tag}));
-    }
-  }
+  const tags = Object.entries(services)
+    .filter(([key]) => key.startsWith(prefix))
+    .map(([, value]) => value)
+    .flat()
+    .reduce((accu, tag) => accu.concat({name: tag}), []);
```
graduta: Rather than using an accumulator to concatenate I think it is better to use `map`
`  .map((tag) => ({name: tag}))`

QualityControl/lib/api.js
awegrzyn: Okay, indeed, I think you're right.

QualityControl/lib/api.js
awegrzyn: Yes, indeed, it's more elegant

graduta (approved):
Kind reminder that you also need to bump the web-ui version in this PR and the version of QC

PR #1095, opened 2022-02-15
[OGUI-798] Use token scopes for authorisation

awegrzyn:


Framework/Backend/http/server.js
```diff
@@ -443,17 +443,14 @@ class HttpServer {
   /**
    * Provides access level number for JWT token depending on users' role
    * @param {object} details - user details
-   * @return {number} - access level based on role
+   * @return {object} - access roles
```
graduta: Could you add a bit more specific type?
I guess we expect `Array<String>`

Framework/Backend/http/server.js
```diff
@@ -443,17 +443,14 @@ class HttpServer {
...
+    let scope = [];
     if (details.hasOwnProperty('resource_access')) {
-      const roles = details.resource_access[Object.keys(details.resource_access)[0]].roles;
-      if (roles.includes('admin')) {
-        accessLevel = 2;
-      }
+      scope = details.resource_access[Object.keys(details.resource_access)[0]].roles;
     }
-    return accessLevel;
+    return scope.join(',');
```
graduta, line 453: Can it happen that scope is `null` or `undefined` or not an array?
In that case we need to add a catch

awegrzyn: I guess it could but this is coming from official docs and has never been changed. In such case we should cover for all other parameters as well, eg. `cern_person_id` could change to `cern_id`.

graduta: I meant that calling a method such as `.join` on a `null` or `undefined` will trow an exception

graduta (approved):

PR #1103, opened 2022-02-16
[OGUI-1084] Adapt view according to assigned SSO roles

awegrzyn:


Control/public/services/DetectorService.js
```diff
@@ -26,6 +26,9 @@ export default class DetectorService extends Observable {
   constructor(model) {
     super();
     this.model = model;
+    this.roles = this.model.session.access
```
graduta: In the constructor usually we initialize with empty values and then any logic is applied in the `init` method.
Could you please define roles as empty array and move the logic to init method?

Control/public/Model.js
```diff
@@ -38,6 +38,7 @@ export default class Model extends Observable {
 
     this.session = sessionService.get();
     this.session.personid = parseInt(this.session.personid, 10); // cast, sessionService has only strings
+    this.session.admin = this.session.access.includes('admin') ? true : false;
```
graduta: If this is a boolean value, we should name it following the standard `isAdmin`

Control/public/common/detectorModal.js
```diff
@@ -32,20 +32,24 @@ const detectorsModal = (model) =>
         model.detectors.listRemote.match({
           NotAsked: () => null,
           Loading: () => h('.w-100.text-center', loading(2)),
-          Success: (data) => detectorsList(model, data),
+          Success: (data) => detectorsList(model, model.session.admin ?
```
graduta: I think it would be cleaner if we move the detector logic on what detectors are sent in the `detectorList` and we keep the `match` statement simpler.

Control/public/common/detectorModal.js
```diff
@@ -32,20 +32,24 @@ const detectorsModal = (model) =>
...
           Failure: (_) => h('.w-100.text-center.danger', [
-            iconCircleX(), ' Unable to load list of detectors. Use GLOBAL View'
+            iconCircleX(), ' Unable to load list of detectors.'
           ])
         })
       ]),
       h('.w-100.pv3.f3.flex-row', {style: 'justify-content:center;'},
         h('.w-50.flex-column.dropdown#flp_selection_info_icon', [
-          h(`button.btn.btn-default.w-100`, {
+          (model.session.admin || model.detectors.roles.includes('GLOBAL'))
```
graduta: Is this supposed to be similar to a ternary operator?
If condition is met we go with `GLOBAL`, otherwise with `GUEST`?
If yes, can we use the ternary operator?

Or even better, we can define an individual node for the button and we send as parameter `GUEST` or `GLOBAL` as they are pretty much the same

graduta (changes requested):

graduta:
We also have to take into consideration how the `GUEST` account will behave on the rest of the pages.
e.g. 
* `newEnv` page filters the detectors list and variables on the detector selection
* same for `cru`, `flps`, `tasks` pages

awegrzyn:
Tests will follow...

Control/public/Model.js
```diff
@@ -94,22 +93,22 @@ export default class Model extends Observable {
    * @returns {object} User's role 
    */ 
   setRole() {
-    let role = this.Roles.Guest;
+    let role = ROLES.Guest;
```
graduta: I think this is a `get` rather than a `set` method. Also there is no need for variable `role` to be defined. 
Is it only us adding `det-` prefix? If not maybe it would be good to make the role and prefix uppercase before comparison.

It can be written with return statements:
```
getRole() {
		if (this.session.access.includes('admin')) {
      return ROLES.Admin;
    } else if (this.session.access.includes('global')) {
      return ROLES.Global;
    } else if (this.session.access.some((role) => role.toUpperCase().startsWith(PREFIX.SSO_DET_ROLE.toUpperCase()))) {
      return ROLES.Detector;
    }
    return ROLES.Guest;
  }
```

Control/public/Model.js
```diff
@@ -86,6 +88,34 @@ export default class Model extends Observable {
...
+    }
+    if (this.session.access.includes('admin')) {
+      role = ROLES.Admin;
+    }
+    return role;
+  }
+
+  /** 
+   * Evaluate whether action is allowed for given role
+   * @param {ROLES} target role
```
graduta: I believe ESLINT will complain a bit here. The standard for documenting a parameter includes the parameter name
Something like
```
 * @param {ROLES} role - target role
 ```

File: Control/public/common/detectorModal.js [hunk 7]
```diff
@@ -58,11 +58,13 @@ const detectorsModal = (model) =>
...
  */
 const detectorsList = (model, list) =>
-  list.map((detector) => h('.w-25.pv3.text-center.f3',
-    h('button.btn.btn-default.w-70', {
-      id: `${detector}ViewButton`,
-      onclick: () => model.setDetectorView(detector)
-    }, detector)
-  ));
+  list
+    .filter(detector => model.detectors.authed.includes(detector) || model.isAllowed(ROLES.Global))
```
graduta, line 62: Missing brackets around parameter
`.filter((detector) =>`

awegrzyn: is it required?

graduta (approved):

PR #1119, opened 2022-02-25
[OGUI-847] Per-detector lock

awegrzyn:


Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+        throw new Error(`Lock ${entity} is already hold by ${this.lockedByName[entity]} (id ${this.lockedBy[entity]})`);
+      }   
+      this.lockedBy[entity] = req.session.personid;
+      this.lockedByName[entity] = req.session.name;
+      log.info(`Lock ${entity} taken by ${req.session.name}`);
+      this.broadcastLockState();
+      res.status(200).json({ok: true});
+    } catch (error) {
+      log.error(`Unable to lock by ${req.session.name}: ${error}`);
+      res.status(403).json({message: error.message});
```
graduta: I would recommend using the `errorHandler` from utility module. This will also take care of trace of error. 
```suggestion
     errorHandler(`Unable to force lock by ${req.session.name}: ${error}`), res, 403, 'lockservice');
```

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+   * Method to try to acquire lock
+   * @param {Request} req
+   * @param {Response} res
+   */
+  lockDetector(req, res) {
+    try {
+      if (!('name' in req.body)) {
+        throw new Error('Unspecified lock entity');
+      }
+      const entity = req.body.name;
```
graduta: You can directly check for `name`
```suggestion
      const entity = req.body?.name;
      if (!entity) {
        throw new Error('Unspecified lock entity');
      } 
```

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+  } 
+
+  /** 
+   * Method to try to acquire lock
+   * @param {Request} req
+   * @param {Response} res
+   */
+  lockDetector(req, res) {
+    try {
+      if (!('name' in req.body)) {
```
graduta: I am a bit confused here. By `Name` we refer to the `username` or full name which might contain spaces which is not suitable to be a key

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+      this.broadcastLockState();
+      res.status(200).json({ok: true});
+    } catch (error) {
+      log.error(`Unable to lock by ${req.session.name}: ${error}`);
+      res.status(403).json({message: error.message});
+    }
+  }
+
+  /** 
+   * Method to try to release lock
```
graduta: Please update documentation accordingly

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+   * @param {Request} req
+   * @param {Response} res
+  */
+  forceUnlock(req, res) {
+    try {
+      if (!('name' in req.body)) {
+        throw new Error('Unspecified lock entity');
+      }
+      const entity = req.body.name;
```
graduta:
```suggestion
    const entity = req.body?.name;
      if (!entity) {
        throw new Error('Unspecified lock entity');
      } 
```

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+        throw new Error(`Insufficient permission`);
+      }   
+      delete this.lockedBy[entity];
+      delete this.lockedByName[entity];
+      log.info(`Lock ${entity} forced by ${req.session.name}`);
+      this.broadcastLockState();
+      res.status(200).json({ok: true});
+    } catch (error) {
+      log.error(`Unable to force lock by ${req.session.name}: ${error}`);
+      res.status(403).json({message: error.message});
```
graduta: I would recommend using the `errorHandler` from utility module. This will also take care of trace of error. 
```suggestion
     errorHandler(`Unable to force lock by ${req.session.name}: ${error}`), res, 403, 'lockservice');
```

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+   * Method to try to release lock
+   * @param {Request} req
+   * @param {Response} res
+   */
+  unlockDetector(req, res) {
+    try {
+      if (!('name' in req.body)) {
+        throw new Error('Unspecified lock entity');
+      }
+      const entity = req.body.name;
```
graduta:
```suggestion
      const entity = req.body?.name;
      if (!entity) {
        throw new Error('Unspecified lock entity');
      }
```

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+      if (!(entity in this.lockedBy)) {
```
graduta, line 118: Do we want to throw an error if the lock is already released? In the end this is the expected behavior from the user; the action ends with the lock not being taken

Control/lib/services/Lock.js
```diff
@@ -0,0 +1,137 @@
...
+        throw new Error(`${entity} owner is ${this.lockedByName[entity]} (id ${this.lockedBy[entity]})`);
+      }
+      delete this.lockedBy[entity];
+      delete this.lockedByName[entity];
+      log.info(`Lock ${entity} released by ${req.session.name}`);
+      this.broadcastLockState();
+      res.status(200).json({ok: true});
+    } catch (error) {
+      log.error(`Unable to give away lock to ${req.session.name}: ${error}`);
+      res.status(403).json({message: error.message});
```
graduta:
```suggestion
      errorHandler(`Unable to give away lock to ${req.session.name}: ${error}`, res, 403, 'lockservice');
```

Control/public/common/detectorHeader.js
```diff
@@ -30,7 +31,9 @@ const detectorHeader = (model) => {
         h('.ph2', `Detector View: ${selected}`),
         h('a.f6.actionable-icon', {
           onclick: () => model.resetDetectorView('')
```
graduta: You can add some padding to the main edit icon rather than a new element with 4px 
```suggestion
        h('a.f6.actionable-icon.ph2', {
          onclick: () => model.resetDetectorView('')
```

Control/public/lock/Lock.js
```diff
@@ -29,6 +29,41 @@ export default class Lock extends Observable {
     this.padlockState = RemoteData.notAsked(); // {lockedBy, lockedByName}
   }
 
+  /**
+   * Provides name of lock holder
+   * @param {string} name Lock entity
```
graduta: I think here we need a bit of an example as to not confuse what is what

Control/public/lock/Lock.js
```diff
@@ -60,49 +95,43 @@ export default class Lock extends Observable {
    * Ask server to get the lock of Control
    * Result of this action will be an update by WS
    */
-  async lock() {
-    this.padlockState = RemoteData.loading();
-    this.notify();
-
-    const {result, ok} = await this.model.loader.post(`/api/lock`);
+  async lock(entity) {
+    const {result, ok} = await this.model.loader.post(`/api/lock`, {name: entity});
```
graduta, line 99: I can see below in `forceLock` we still make use of `this.padlockState`
Is there any particular reason for here to be removed?

Control/public/lock/Lock.js
```diff
@@ -60,49 +95,43 @@ export default class Lock extends Observable {
...
    * Ask server to release the lock of Control
    * Result of this action will be an update by WS
    */
-  async unlock() {
-    this.padlockState = RemoteData.loading();
-    this.notify();
-
-    const {result, ok} = await this.model.loader.post(`/api/unlock`);
+  async unlock(entity) {
+    const {result, ok} = await this.model.loader.post(`/api/unlock`, {name: entity});
```
graduta, line 126: I can see above in `forceLock` we still make use of `this.padlockState`
Is there any particular reason for here to be removed?

Control/public/lock/lockPage.js
```diff
@@ -0,0 +1,80 @@
...
+ * or submit itself to any jurisdiction.
+*/
+
+import {h} from '/js/src/index.js';
+import {detectorHeader} from '../common/detectorHeader.js';
+import {detectorLockButton} from './lockButton.js';
+import {ROLES} from './../workflow/constants.js';
+
+/**
+ * @file Header of the Task Page that displays the title and 2 clean operations
```
graduta: Documentation missleading

Control/public/lock/lockPage.js
```diff
@@ -0,0 +1,80 @@
...
+    h('h4', 'Locks')
+  ]), 
+  model.detectors.selected === 'GLOBAL' && h('.flex-row.text-right', {
+    style: 'position: absolute; right: 0px;'
+  })  
+];
+
+/**
+ * Content
+ * Show loading or error on other cases
```
graduta: Documentation is not inline with expected code

Control/public/workflow/panels/flps/detectorsPanel.js
```diff
@@ -36,37 +37,48 @@ export default (workflow) => {
 /**
  * Display an area with selectable elements representing detectors
  * @param {Array<string>} list
- * @param {Object} workflow
+ * @param {Object} model
```
graduta, line 39: I think the order is and was even before wrong in here

Control/test/lib/mocha-control-service.js
```diff
@@ -23,33 +23,27 @@ const AssertionError = require('assert').AssertionError;
 
 describe('Control Service test suite', () => {
   describe('Check Constructor', () => {
-    it('should throw error due to missing PadLock dependency', () => {
```
graduta, line 26: As these should comeback with the future ticket maybe it helps if we skip the for now rather than deleting
`it.skip`

Control/test/public/page-new-environment-mocha.js
```diff
@@ -133,40 +133,13 @@ describe('`pageNewEnvironment` test-suite', async () => {
     assert.deepStrictEqual(selectedWorkflow.classList, {0: 'w-90', 1: 'menu-item', 2: 'w-wrapped', 3: 'selected'});
   });
 
-  it('should throw error when `Create` button is clicked due to `Control is not locked`', async () => {
```
graduta, line 136: Is the `Create` button not disabled by default anymore?
If it is maybe we can check that is disabled as not lock is taken

graduta (changes requested):

Control/lib/services/Lock.js
awegrzyn: Yes, indeed I agree, exceptions are dedicated for situations that are not expected, but in such case we shall remove almost all `throws` from Lock service.

Control/test/public/page-new-environment-mocha.js
awegrzyn: I think it could be disabled until at least 1 detector is not selected, but I don't see any relation with the locks.

Control/test/lib/mocha-control-service.js
awegrzyn: This will need to be totally reimplemented anyway

PR #1165, opened 2022-03-16
[OGUI-1112] Display list of create env requests

awegrzyn:


Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+    this.requestList = {};
+  }
+
+  /**
+   * Add AliECS request list to "cache", remove it from the "cache" once response comes from AliECS
+   * @param {object} req
+   * @param {object} res
+   */
+  async add(req, res) {
+    const index = Object.keys(this.requestList).length;
```
graduta: I think here we are exposing ourselves to a potential bug. Lets assume we have 2 environments in creation. id is "0" and "1".
Then environment "0" is successfully created and removed from requests lists. We then have in our map { "1": {}}
Another user attempts to create an environment and this index will be assigned "1". Thus, the existing request will be overwritten. 
As you are using "indexes" I would recommend you use an array and then simply add to the end of it.

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+    this.requestList[index] = {
+      id: index,
+      detectors: req.body.detectors,
+      workflow: req.body.workflowTemplate,
+      date: new Date(),
+      owner: req.session.name,
+      personid: req.session.personid,
+      failed: false
+    };
+    res.json({ok: 1});
```
graduta, line 48: Is there any particular reason for this response? Particularly `1`?
Is that the index of request? We can also simply not return a buddy but 200 only if we are not interested in the payload.

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+    log.debug('Added request to cache, ID: ' + index);
+    try {
+      await this.ctrlService.executeCommandNoResponse('NewEnvironment', req.body);
+      log.debug('Auto-removed request, ID: ' + index);
+      delete this.requestList[index];
+    } catch(error) {
+      log.debug('Request failed, ID: ' + index);
```
graduta: I think here we would like to store in infologger the error, right?
Might be better to use the errorLogger from utils

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Remove request from "cache".
+   * @param {object} req
+   * @param {object} res
+   */
+  remove(req, res) {
+    const index = req.body.id;
+    log.debug('User removed request, ID: ' + index);
+    if (index in this.requestList) {
+      delete this.requestList[index];
```
graduta: If it is a JSON, the delete method is safe for non-existing indexes. No need to have the if clause.

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return this.getAll(req, res);
```
graduta, line 71: Missing return tag in method documentation

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * Get all the requests from the "cache"
+   * @param {object} req
+   *  @param {object} res
```
graduta: Extra white space and missing return statement
Also, do we want to return the result of the response call?

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+const {Log} = require('@aliceo2/web-ui');
+const log = new Log(`${process.env.npm_config_log_label ?? 'cog'}/controlrequests`);
+
+/**
+ * Handles AliECS create env requests
```
graduta, line 20: I would suggest we keep the naming convention of Control.
I can see that 2 lines above the log will use `controlrequest` as label. The class itself is called `RequestHandler` while in the api file it is called `AliECSRequestHandler`

I would suggest we rename it `EnvironmentController` or `ControlEnvironmentControlleer` as it deals with requests for controlling an environment

Control/lib/control-core/RequestHandler.js
```diff
@@ -0,0 +1,86 @@
...
+   * @param {object} ctrlService - Handle to Control service
+   */
+  constructor(ctrlService) {
+    this.ctrlService = ctrlService;
+    this.requestList = {};
+  }
+
+  /**
+   * Add AliECS request list to "cache", remove it from the "cache" once response comes from AliECS
+   * @param {object} req
```
graduta: The convention followed is to specify the object type where it is known. In this case is `Request` and `Response` rather than `object`

Control/lib/api.js
```diff
@@ -70,6 +73,10 @@ module.exports.setup = (http, ws) => {
   ctrlProxy.methods.forEach(
     (method) => http.post(`/${method}`, coreMiddleware, (req, res) => ctrlService.executeCommand(req, res))
   );
+  http.post('/NewEnvironmentRequest', coreMiddleware, (req, res) => aliecsReqHandler.add(req, res));
+  http.post('/GetEnvironmentRequests', coreMiddleware, (req, res) => aliecsReqHandler.getAll(req, res));
```
graduta: Is there a reason for which we keep the gRPC naming? If not, I would not continue to follow gRPC convention for our own request. I would suggest that we follow the NodeJS standard similar to below:
```
http.post('environment')
http.get('environments')
http.delete('environment/:id')
```
I believe our frontend loader does not support 'delete' so for now for delete we keep
`http.post('environment/:id')`

Control/public/environment/environmentsPage.js
```diff
@@ -61,10 +68,72 @@ export const content = (model) => h('.scroll-y.absolute-fill.text-center', [
...
+/**
+ * Show list of create env request to AliECS core as a table
+ * @param {Object} model
+ * @param {array} requests List of requested stored in the backend
+ * @param {Date} now Current date according to backend
+ */
+const showRequests = (model, requests, now) =>
+  (requests && requests.length > 0)
+    ? requestsTable(model, requests, now)
+    : h('h3.m4', ['No requests found.']);
```
graduta: When looking at the environments page and always raeading "No requests found.", it is a bit confusing as a user.

I would suggest we do not show anything if there is no creation happening.

Control/public/Model.js
```diff
@@ -190,9 +190,11 @@ export default class Model extends Observable {
     switch (this.router.params.page) {
       case 'environments':
         this.environment.getEnvironments();
+        this.environment.getEnvironmentRequests();
         this.frameworkInfo.getIntegratedServicesInfo();
         this.environment.refreshInterval = setInterval(() => {
           this.environment.getEnvironments();
+          this.environment.getEnvironmentRequests();
```
graduta, line 197: Currently the production setup is set to 2 minutes and 1 second for a refresh. 
For now, I think we would need a different timer for pending requests refresh.
We cannot reduce the existing ones as "getEnvironments" would spam AliECS with too many requests.

Control/public/environment/environmentsPage.js
```diff
@@ -61,10 +68,72 @@ export const content = (model) => h('.scroll-y.absolute-fill.text-center', [
...
 
+/**
+ * Calculate duration from two dates and pretify result
+ * @param {Date} now Current date
+ * @param {Date} startDate Date when a request was cached
+ */
+const durationDisplay = (now, startDate) => {
```
graduta: While the idea of having a timer is good, I think we should not compare only on request refresh. 
We should have a frontend timer which refreshes itself after 1 second. Reason behind is that currently on my own setup I have a 2 minutes refresh period which means for me it simply displays 0s for 2 minutes

graduta (changes requested):

Control/public/environment/environmentsPage.js
```diff
@@ -78,6 +147,7 @@ const environmentsTable = (model, list) => {
   ];
   return h('table.table', [
     h('thead', [
+      h('tr.table-primary',  h('th', {colspan: 12}, 'Created environments')),
```
graduta: I believe "Active Environments" is the preferred naming we received as feedback

Control/lib/control-core/RequestHandler.js
awegrzyn: Indeed, good catch, but I think the array is not a good structure for this purpose, I'd just use `Object.keys(obj).pop() + 1 || 0` as index.

graduta: I still think this is not good enough. For example
```
const requests = {}

let index = Object.keys(requests).pop() || 0;
// Even though you specify 0, JS will store it as a String Key
requests[index] = {name: 'some-request'}; // we add a request

index = Object.keys(requests).pop() || 0; // then we want to add another request
// this index will be a String '0' so it will once again override
```

This happens because it will return the last inserted Key and not the actual index of it

Control/public/Model.js
awegrzyn: This will be addressed by get envs cache.

Control/public/environment/environmentsPage.js
awegrzyn: Okay, then in such case I'd just display created timestamp, and leave the timer as future improvement.

graduta: I agree

Control/lib/control-core/RequestHandler.js
awegrzyn: I think RemoteData expects JSON, so `200` might not be enough

Control/lib/control-core/RequestHandler.js
awegrzyn: uhh, indeed, I've tested this now: `parseInt(Object.keys(obj).pop()) + 1 || 0;`

graduta (approved):

PR #1176, opened 2022-03-22
[OGUI-1115] Cache env list and use WS for updates

awegrzyn:


Control/lib/control-core/EnvCache.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+class EnvCache {
+
+  /**
+   * @param {object} ctrlService - Handle to Control service
+   */
+  constructor(ctrlService) {
+    this.ctrlService = ctrlService;
+    this.cache = {};
+    this.refreshInterval = setInterval(() => this.refresh(), 1000);
```
graduta: I still believe we should check if previous request completed. 
There is no point in flooding AliECS with hundreds of requests if it is hanging. It will only come back to us as request to address it.

graduta (approved):

Control/lib/control-core/EnvCache.js
awegrzyn: Okay, @teo what would be reasonable time to wait for `GetEnvironments` to complete?

awegrzyn: @graduta Indeed, I'm also not sure what to do when `GetEnvironments` timeouts? Do we clear envs from cache? Maybe it's better to keep them but clear the list only on AliECS core restart (is there a way to detect it?).

graduta: Hi @awegrzyn . From my point of view I would propose we:
1. keep the list so that shifters can easily copy-paste the env id so that they search in infologger
2. Display an informative message to the user that AliECS hangs
3. We do not allow the user to go to "More"of an environment as AliECS is hanging anyway
4. If AliECS is back then we get a new list of Envs and Cache is updated

gPRC.js will send back a standard "DEADLINE_EXCEEDED" message if timeout is reached but other than that we cannot tell if something is wrong. As we also have lengthy times for creating an environment, I would say is not ok to assume anything until we get the timeout message

teo: > Okay, @teo what would be reasonable time to wait for `GetEnvironments` to complete?

Not sure, a few seconds? It should reply immediately or almost.

awegrzyn: ok, thx

graduta (approved):

Control/public/environment/environmentsPage.js
```diff
@@ -101,6 +101,7 @@ const requestsTable = (model, requests) =>
         )),
         h('td', {style: 'text-align: center;'}, item.owner),
         h('td', {style: 'text-align: center;'}, new Date(item.date).toLocaleString()),
+        h('td', {style: 'text-align: center;font-weight: bold;'}, item.failed ? 'FAILED' : 'ONGOING'),
```
graduta, line 104: Instead of ongoing I would propose you add the loading spinner to get another request out of the way;
```suggestion
        h('td', {style: 'text-align: center;font-weight: bold;'}, item.failed ? 'FAILED' : loading(2)),
```

At the top of the file, import the loading spinner as:
`import loading from './../common/loading.js';`

awegrzyn: One request was to add "state" (ONGOING/FAILED) and another to add timer.

PR #1195, opened 2022-03-29
[OGUI-1127] Fix detector display in env request table

awegrzyn:


Control/public/environment/environmentsPage.js
```diff
@@ -98,7 +98,7 @@ const requestsTable = (model, requests) =>
     ]),
     h('tbody', [
       requests.map(item => h('tr', {style: {background: item.failed ? 'rgba(214, 38, 49, 0.2)' : '' }}, [
-        h('td', {style: 'text-align: center;'}, item.detectors),
+        h('td', {style: 'text-align: center;'}, item.detectors.join(' ')),
```
graduta: Could you please follow the current example from the "Active Environments" table and display `-` in case of no detectors (for ex QC workflows)
Moreover, maybe adapt that one to using `join` rather than `map` or the other way around

graduta (changes requested):

graduta (approved):

PR #1197, opened 2022-03-30
[OGUI-1131] Add env ID to env request table

awegrzyn:
This is ugly workaround to get environment ID from error message details. This should be improved to use protobuf deserialisation (we need to change protobuf to do so).

### DIFF
--- a/Control/lib/control-core/GrpcProxy.js
+++ b/Control/lib/control-core/GrpcProxy.js
@@ -70,12 +70,18 @@ class GrpcProxy {
       return new Promise((resolve, reject) => {
         this.client[methodName](args, options, (error, response) => {
           if (error) {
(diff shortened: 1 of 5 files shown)


Control/lib/control-core/GrpcProxy.js
```diff
@@ -70,6 +70,13 @@ class GrpcProxy {
       return new Promise((resolve, reject) => {
         this.client[methodName](args, options, (error, response) => {
           if (error) {
+            if (methodName === 'NewEnvironment' && error.metadata?.internalRepr?.has('grpc-status-details-bin')) {
```
graduta: I see there are potentially lots of places where there might be missing fields which will lead to exceptions being thrown.
I would suggest we wrap it in a `try/catch` just to be sure

graduta (changes requested):

graduta (approved):

PR #1207, opened 2022-04-01
[OGUI-1135] Properly deserialise Env ID from grpc error details

awegrzyn:


graduta (approved):

graduta (changes requested):
Please add back the `NewEnvironment` check and have a look at LGTM errors

awegrzyn:
> have a look at LGTM errors

The LGTM error refers to `google-protobuf` generate code

graduta:
> > have a look at LGTM errors
> 
> The LGTM error refers to `google-protobuf` generate code

Yes, but this is a file we do not generate every time. I do not think it will create problems if we remove the unused variables.

graduta (approved):

PR #1232, opened 2022-04-18
[OGUI-1147] Add mandatory author to notifications

awegrzyn:


Framework/Backend/services/notification.js
```diff
@@ -65,23 +65,30 @@ class NotificationService {
 
   /**
    * Sends a message to selected topic
-   * @param {string} topic Kafka topic
-   * @param {string} message message to be sent
+   * @param {string} tag Bookkeeping tag
```
graduta: Is the tag `string` or `array` or can it be both?
I can see it in the parameters as string but the code checks if  it is an array

graduta (approved):

Framework/Backend/services/notification.js
awegrzyn: yes, it can be both, I'll correct jsdoc

graduta (approved):

PR #1451, opened 2022-07-21
[OGUI-1220] Add button to request Web Notification permission

awegrzyn:


awegrzyn:
@graduta @martinboulais could you review? Do you think there's better place for that button?

Control/public/Model.js
```diff
@@ -297,14 +297,23 @@ export default class Model extends Observable {
...
   }
 
+  /**
+   * Check whether it makes sense to ask for notification permission
+   */
+  checkBrowserNotificationPermissions() {
+    return (Notification.permission === 'denied' ||
+      Notification.permission === 'default') &&
+      window.location.protocol == "https:";
+  }
```
martinboulais: 2 small details: if it is not too much work maybe you can put a different name, here I would expect it to return true if the notification permissions **is** granted
Also for consistency the third comparison may use strict comparison

martinboulais:
For me the user menu is fine for the notifications enabling

Control/public/Model.js
```diff
@@ -297,14 +297,25 @@ export default class Model extends Observable {
...
 
+  /**
+   * Defines policy when user can trigger request for enabling notifications
+   * This is used to display/hide enable button
+   * @returns {bool} True when notifications are not enabled and HTTPS is used
+   */
+  checkBrowserNotificationPermissions() {
+    return (Notification.permission === 'denied' ||
+      Notification.permission === 'default') &&
+      window.location.protocol === "https:";
```
graduta, line 316: As per the API, the `Notification` is only provided by the browser in `HTTPS` protocol so there is no need to check on it.

Control/public/Model.js
```diff
@@ -297,14 +297,25 @@ export default class Model extends Observable {
   }
 
   /**
-   * After pages load check if the user enabled notifications
+   * Request notification permission
    */
-  checkBrowserNotificationPermissions() {
+  requestBrowserNotificationPermissions() {
```
graduta, line 302: I would recommend you make use of the `check` method below to see if notifications are granted or not as you already have it.

Control/public/Model.js
```diff
@@ -297,14 +297,25 @@ export default class Model extends Observable {
...
     if (Notification.permission !== 'granted') {
       Notification.requestPermission();
     }
   }
 
+  /**
+   * Defines policy when user can trigger request for enabling notifications
+   * This is used to display/hide enable button
+   * @returns {bool} True when notifications are not enabled and HTTPS is used
```
graduta: `bool` is of type `boolean` in JS
```suggestion
   * @returns {boolean} True when notifications are not enabled and HTTPS is used
```

graduta (changes requested):

Control/public/Model.js
awegrzyn: Yes, but is being used to define whether button is displayed or not

Control/public/common/appHeader.js
```diff
@@ -41,6 +41,8 @@ const loginButton = (model) => h('.dropdown', {class: model.accountMenuEnabled ?
     h('p.m3.mv2.text-ellipsis', `Welcome ${model.session.name}`, h('sup', model.session.role)),
     model.session.personid === 0 // anonymous user has id 0
       && h('p.m3.gray-darker', 'You are connected as anonymous, no authentification needed for this application.'),
+    model.checkBrowserNotificationPermissions() &&
+      h('a.menu-item', {onclick: () => model.requestBrowserNotificationPermissions()}, 'Enable notifications'),
```
graduta: Once the button is clicked, I believe it would be better from a UX point of view to:
* display a web-ui notification to inform the user that browser notifications have been enabled; Or even better a browser-notification
* close the drop-down menu

graduta (changes requested):

Control/public/common/appHeader.js
awegrzyn: ad 1) Once clicked, user is notified by browser, so I guess it's good enough.
Screenshot 2022-07-29 at 15 17 19

ad 2) Yes agreed.

graduta (approved):

PR #1747, opened 2023-01-17
[OGUI-1301] Migrate to jsonwebtoken v9 and refactor O2 token class

graduta:
- [x] PR labels are selected

PR which:
* Migrates to v9 of `jsonwebtoken` library;
* Refactors `token.js` class into `O2WebToken.js` 
* Bumps minor version of `web-ui` framework


Framework/Backend/websocket/server.js
```diff
@@ -109,18 +109,19 @@ class WebSocket {
...
-        client.on('message', (message) => this.onmessage(message, client));
-        client.on('close', () => this.onclose(client));
-        client.on('pong', () => client.isAlive = true);
-        client.on('error', (err) => this.log.error(`Connection ${err.code}`));
-      }, (error) => {
-        this.log.debug(`${error.name} : ${error.message}`);
-        client.close(1008);
-      });
+    try {
+      const decoded = this.http.o2WebToken.verify(token);
```
martinboulais: I think the try should wrap as less code as possible, so here only this line. Then in the catch you return after the close, and you continue manipulating client after

martinboulais (changes requested):

Framework/Backend/websocket/server.js
graduta: Good suggestion, updated accordingly.

martinboulais (approved):

PR #1786, opened 2023-02-13
[OGUI-1307] Adds controllers for broadcasting statuses to clients

graduta:
- [x] PR labels are selected

PR which:
* Adds controllers for status GET requests
* Adds controller for broadcasting messages to the subscribed clients
* Adds service for setting up intervals to update status information
* Splits status service methods in 2 so that they are reused in intervals and controller


Control/lib/api.js
```diff
@@ -107,28 +119,23 @@ module.exports.setup = (http, ws) => {
...
   // Consul
-  const validateService = consulConnector.validateService.bind(consulConnector);
-  http.get('/consul/flps', validateService, (req, res) => consulConnector.getFLPs(req, res));
-  http.get('/consul/crus', validateService, (req, res) => consulConnector.getCRUs(req, res));
-  http.get('/consul/crus/config', validateService, (req, res) => consulConnector.getCRUsWithConfiguration(req, res));
-  http.get('/consul/crus/aliases', validateService, (req, res) => consulConnector.getCRUsAlias(req, res));
-  http.post('/consul/crus/config/save', validateService, (req, res) => consulConnector.saveCRUsConfiguration(req, res));
+  const validateService = consulController.validateService.bind(consulController);
+  http.get('/consul/flps', validateService, (req, res) => consulController.getFLPs(req, res));
+  http.get('/consul/crus', validateService, (req, res) => consulController.getCRUs(req, res));
```
martinboulais: Why do you use a lambda here and do not simply bind the method?

Control/lib/control-core/ControlService.js
```diff
@@ -109,7 +109,7 @@ class ControlService {
       try {
         vars = !vars ? {} : vars;
         if (!vars.hosts) {
-          vars.hosts = await this.consulConnector.getFLPsList();
+          vars.hosts = await this.consulController.getFLPsList();
```
martinboulais, line 112: I think that controllers should never be used in services but only the other way around, hese I think you should have a consulService

Control/lib/api.js
```diff
@@ -12,17 +12,21 @@
...
-const StatusService = require('./services/StatusService.js');
+const {StatusService} = require('./services/Status.service.js');
+const {Intervals} = require('./services/Intervals.service.js');
 
-// connectors
-const NotificationService = require('@aliceo2/web-ui').NotificationService;
-const ConsulConnector = require('./connectors/ConsulConnector.js');
+// web-ui services
+const {NotificationService, ConsulService} = require('@aliceo2/web-ui');
+const {ConsulController} = require('./controllers/Consul.controller.js');
```
martinboulais: I think that this should go l.21

Control/lib/controllers/Status.controller.js
```diff
@@ -0,0 +1,153 @@
...
+   * @param {Response} res - HTTP Response object
+   * @returns {undefined}
+   */
+  async getConsulStatus(_, res) {
+    try {
+      const componentStatus = await this._statusService.getConsulAsComponent();
+      res.status(200).json(componentStatus);
+    } catch (error) {
+      const message = 'Unable to retrieve status of CONSUL';
+      this.log.errorMessage(message, {level: 99, facility: 'cog/status-ctrl', system: 'GUI'});
```
martinboulais: I think that the level should be a constant somewhere (no magic numeric values), and I would even suggest to factorize the whole configuration that seems to be the same in that class

Control/lib/controllers/WebSocket.controller.js
```diff
@@ -0,0 +1,56 @@
...
+ * - GUI information (status, host, port)
+ * - AliECS information
+ * - AliECS Integrated Services information
+ * - AliECS Apricot
+ * - Consul
+ * - Grafana
+ */
+class WebSocketController {
+  /**
+   * @param {WebSocket} - object to be used to send messages to clients
```
martinboulais: You forgot the parameter name

Control/lib/services/Intervals.service.js
```diff
@@ -0,0 +1,54 @@
...
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+/**
+ * Class which deals with setting up intervals for sending messages to clients via WebSocket
```
martinboulais: I think this docblock is a bit outdated

Control/lib/services/Status.service.js
```diff
@@ -0,0 +1,283 @@
...
+
+    this.CONSUL_KEY = 'CONSUL';
+    this.GRAFANA_KEY = 'GRAFANA';
+    this.NOTIFICATION_SYSTEM_KEY = 'NOTIFICATION_SYSTEM';
+
+    this.ALIECS_SERVICES = 'ALIECS_SERVICES';
+    this.ALIECS_CORE = 'ALIECS_CORE';
+    this.APRICOT_KEY = 'APRICOT';
+
+    this.NOT_CONFIGURED = 'This service was not configured';
```
martinboulais: I think those can be module constants, to avoid them being changed

Control/lib/utils.js
```diff
@@ -40,31 +40,36 @@ function errorLogger(err, facility = 'utils') {
 }
 
 /**
-  * Util to get JSON data (parsed) from server
+  * Util to get JSON data (parsed) following a GET request
   * @param {string} host - hostname of the server
   * @param {number} port - port of the server
   * @param {string} path - path of the server request
+  * @param {JSON} options - specific request options (e.g range of accepted status code)
```
martinboulais, line 47: parameter is named options here but params bellow

Control/lib/utils.js
```diff
@@ -40,31 +40,36 @@ function errorLogger(err, facility = 'utils') {
...
   * @return {Promise.<Object, Error>} JSON response
   */
-function httpGetJson(host, port, path) {
+function httpGetJson(hostname, port, path, params = undefined) {
+  const options = params ?? {
+    statusCodeMin: 200,
+    statusCodeMax: 299,
+    rejectMessage: 'Non-2xx status code: '
+  };
```
martinboulais: I think it is error prone to rename params to options here

martinboulais (changes requested):
Nice job 👍🏻

Control/lib/api.js
graduta: Updated. Initially I was not planning on touching what is not related to the intervals, but with the renaming I can also add the bind

Control/lib/control-core/ControlService.js
graduta: Yes, but this change needs a bigger refactor which is planned separately for Control, Apricot and Consul.

Control/lib/controllers/Status.controller.js
graduta: Yes, there are a couple of utility methods that deal with this but they did not get updated following the changes of the logger. 
I created OGUI-1308 for that

graduta: For now I updated the values to `const` ones

Control/lib/utils.js
graduta: Removed params and kept only options

martinboulais (approved):

PR #1854, opened 2023-03-21
[OGUI-1339] Add QC and TRG cards summary

graduta:
- [x] PR labels are selected

PR which:
* groups tasks by categories: ALL, FLP, QC Nodes, CTP Readout
* groups FLP tasks by their detector
* expands on `miniCard` to also provide a `miniCardTitle`
* increases the size of Monitoring panels
* moves `runNumber` card on the same line as `General Information` card

### DIFF
--- a/Control/public/environment/Environment.js
+++ b/Control/public/environment/Environment.js
@@ -272,9 +272,9 @@ export default class Environment extends Observable {
    * @param {Array<TaskInfo>} tasks - list of tasks belonging to an environment
    * @returns {{tasks: Array<TaskInfo, machines: number}} - tasks that belong to QC Nodes and number of unique hosts
    */
-  _getQcTasks(tasks) {
+  _getQcTasks(tasks = []) {
     const machines = new Set();
-    const qcTasks = tasks.filter(({deploymentInfo: {hostname}}) => {
+    const qcTasks = tasks.filter(({deploymentInfo: {hostname = ''} = {}}) => {
       if (hostname.match(QC_NODES_NAME_REGEX)) {
         machines.add(hostname)
         return true;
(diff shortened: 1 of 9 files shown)


Control/public/environment/Environment.js
```diff
@@ -265,6 +267,79 @@ export default class Environment extends Observable {
...
+   */
+  _getQcTasks(tasks = []) {
+    const machines = new Set();
+    const qcTasks = tasks.filter(({deploymentInfo: {hostname = ''} = {}}) => {
+      if (hostname.match(QC_NODES_NAME_REGEX)) {
+        machines.add(hostname)
+        return true;
+      }
+      return false;
+    });
```
martinboulais: I think it may be error prone to add side-effect in the filter lambda. I suggest to use the filter first then use map to get machines, or simply use a `for ... of ...` and build the 2 arrays from there.

Control/public/environment/Environment.js
```diff
@@ -265,6 +267,79 @@ export default class Environment extends Observable {
...
+        .filter(({deploymentInfo: {hostname = ''} = {}}) => {
+          const keyDetector = Object.keys(hostsByDetectors)
+            .filter((detector) => hostsByDetectors[detector].includes(hostname))[0];
+          if (includedDetectors.includes(keyDetector)) {
+            machines.add(hostname);
+            return true;
+          }
+          return false;
+        })
+      return {tasks: tasksFiltered, machines: machines.size}
```
martinboulais: Same thing as before, I think that filter should not have side-effect. I suggest to combine all of this in one `for ... of ...` (plus it avoid to loop several times)

Control/public/environment/Environment.js
```diff
@@ -265,6 +267,79 @@ export default class Environment extends Observable {
...
+          const keyDetector = Object.keys(hostsByDetectors)
+            .filter((detector) => hostsByDetectors[detector].includes(hostname))[0];
+          if (includedDetectors.includes(keyDetector)) {
+            return false;
+          }
+          machines.add(hostname);
+          return true;
+        }
+        )
+      return {tasks: tasksFiltered, machines: machines.size}
```
martinboulais: I think this can be factorized, it looks a lot like the function before

martinboulais (changes requested):

Control/public/environment/Environment.js
graduta: Forgot, the chain iterates multiple times. 🤦 
I was thinking of streams when I wrote it. I will update with `for...of`

Control/public/environment/Environment.js
graduta: Refactored all 3 functions in a for...of

Control/public/environment/Environment.js
graduta: Done

martinboulais (approved):

PR #1860, opened 2023-03-22
[OGUI-1341] Add information on components used in the environment

graduta:
- [x] PR labels are selected

PR which:
* adds a new card with information on components used (taken from General Configuration panel)
* factorizes the ODC status and style to be reused between active-envs page and env-details page

### DIFF
--- a/Control/public/common/utils.js
+++ b/Control/public/common/utils.js
@@ -12,6 +12,8 @@
  * or submit itself to any jurisdiction.
 */
 
+import {ODC_STATE_COLOR} from "./constants/stateColors";
+
 /**
  * Method to display specific JSON fields in a particular way
  * @param {Object} item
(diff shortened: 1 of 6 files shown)


Control/public/common/utils.js
```diff
@@ -86,13 +87,20 @@ const getTaskShortName = (taskName) => {
...
-      return odc.State;
+      const {State: state} = JSON.parse(environment.integratedServicesData['odc']);
+      const styleClass = state === 'RUNNING'
+        ? 'success'
+        : (state === 'READY'
+          ? 'primary'
+          : (state === 'ERROR'
+            ? 'danger'
+            : ''));
+      return {state, styleClass};
```
martinboulais: Nested ternary are often hard to read, I suggest to use an object or a map

martinboulais (changes requested):

Control/public/common/utils.js
graduta: Updated to use STATE_COLORS similar to AliECS

martinboulais (approved):

PR #2044, opened 2023-07-18
[OGUI-1243] Skeleton for simplified environment creation page

graduta:
- [x] PR labels are selected

PR which:
* adds new page skeleton for environment creation page in a simplified manner
* imports `jsonFetch` standard from Bookkeeping 
* adds component data for workflow information (default information as per AliECS)


Control/public/pages/EnvironmentCreation/EnvironmentCreation.model.js
```diff
@@ -0,0 +1,68 @@
...
+  constructor(model) {
+    super();
+
+    this._creationModel = null;
+    /**
+     * @type {Model}
+     */
+    this._model = model;
+
+    this._currentWorkflow = RemoteData.notAsked();
```
martinboulais, line 34: You do not set it to `loading`, I am in favor of setting it here

Control/public/pages/EnvironmentCreation/EnvironmentCreation.page.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {Model} model - the global model
+ * @return {vnode} - main component for the creation page of an environment
+ */
+export const EnvironmentCreationPage = (model) => {
+  const {envCreationModel: {currentWorkflow}} = model;
+  return h('', [
+    detectorHeader(model),
+
+    h('.ph2.w-100.text-right', h('a', {
+      style: 'cursor: pointer',
```
martinboulais: You have to set this style on `a` ? Isn't it by default?

martinboulais (changes requested):

Control/public/pages/EnvironmentCreation/EnvironmentCreation.page.js
graduta: Unfortunately yes, because of: https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Frontend/css/src/bootstrap.css#L230

Control/public/pages/EnvironmentCreation/EnvironmentCreation.model.js
graduta: I forgot to set it, indeed.
I would propose I set it in the `initPage` though

Control/public/pages/EnvironmentCreation/EnvironmentCreation.page.js
graduta: Yes, I should put correct `href` and that will solve it.
Thanks for spotting it!

martinboulais (approved):

PR #2049, opened 2023-07-24
[OGUI-1385] Save a list of object paths in-memory to improve loading times of ObjectTree

graduta:
- [x] PR labels are selected

PR which:
* adds an Intervals service which allows other services to register for repeated actions;
* object service registers refreshCache to intervals;
* Number of minutes on which the interval should refresh is configurable;
* Cache prefix which lets the server know what paths should be saved is also configurable.


QualityControl/lib/QCModel.js
```diff
@@ -21,6 +21,7 @@ import { CcdbService } from './services/CcdbService.js';
 import { QcObjectService } from './services/QcObject.service.js';
 import { UserService } from './services/UserService.js';
 import { JsonFileService } from './services/JsonFileService.js';
+import { Intervals } from './services/Intervals.service.js';
```
martinboulais: Why is this not named IntervalsService?

QualityControl/lib/services/CcdbService.js
```diff
@@ -236,6 +238,22 @@ export class CcdbService {
...
+   */
+  get cachePrefix() {
+    return this._cachePrefix;
+  }
+
+  /**
+   * Return the number of ms for the interval for refreshing the cache
+   * @returns {number} - number of ms
+   */
+  get cacheRefresh() {
```
martinboulais: maybe `cacheRefreshFrequency` or `cacheRefreshRate`?

QualityControl/lib/services/Intervals.service.js
```diff
@@ -0,0 +1,50 @@
...
+   * Setup initial request of data and interval for retrieving and updating cache of objects paths from CCDB
+   * @param {number} cacheRefresh - (ms) on how often the cache should be refreshed
+   * @returns {void}
+   */
+  _initializeQcObjectInterval(cacheRefresh = 60 * 1000) {
+    this._logger.debug('Cache - objects - has been initialized');
+    this._qcObjectService.refreshCache();
+    this._intervals.push(setInterval(() => this._qcObjectService.refreshCache(), cacheRefresh));
+  }
+}
```
martinboulais: In my opinion, the interval service should not know about the qc object service and same the other way around. Instead, I would provide a method to register a function to call every X milliseconds, and call this instead of `initializeIntervals`.

martinboulais (changes requested):

QualityControl/lib/QCModel.js
graduta: Missed this, updated

QualityControl/lib/services/CcdbService.js
graduta: Yes, `cacheRefreshRate` is better

QualityControl/lib/services/Intervals.service.js
graduta: Updated to decouple the connection between the 2

martinboulais (approved):

PR #2052, opened 2023-07-25
[OGUI-1386] Add services versions to about page

graduta:
- [x] PR labels are selected

PR which:
* extracts `StatusService` and `StatusController` as they were part of the same file.
* renames API routes to follow convention.
* adds CCDB version retrieval as per CCDB monitor points.
* adds QC version retrieval as per QC recommendation.
* adds QCG version retrieval from `package.json`
* adapts tests of `ObjectService` and `ObjectController` following migration from CommonJS to ES6
* removes `hosts` and `ports` from about page as QCG is meant to go public

### DIFF
--- a/QualityControl/lib/services/Status.service.js
+++ b/QualityControl/lib/services/Status.service.js
@@ -12,10 +12,13 @@
  * or submit itself to any jurisdiction.
  */
 
+import { promisify } from 'node:util';
+import { exec } from 'node:child_process';
+
 import { Log } from '@aliceo2/web-ui';
-import { execSync } from 'node:child_process';
 
 const QC_VERSION_EXEC_COMMAND = 'yum info o2-QualityControl | awk \'/Version/ {print $3}\'';
+const execPromise = promisify(exec);
 
 /**
  * Service for retrieving information and status of QCG dependencies
@@ -61,7 +64,7 @@ export class StatusService {
   async retrieveFrameworkInfo() {
     return {
       qcg: this.retrieveOwnStatus(),
-      qc: this.retrieveQcVersion(),
+      qc: await this.retrieveQcVersion(),
       ccdb: await this.retrieveDataServiceStatus(),
       consul: {
         status: await this.retrieveOnlineServiceStatus(),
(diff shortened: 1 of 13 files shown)


QualityControl/lib/services/Status.service.js
```diff
@@ -0,0 +1,136 @@
...
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { Log } from '@aliceo2/web-ui';
+import { execSync } from 'node:child_process';
```
martinboulais: Isn't the async version better to use in express route? 
```js
import {exec: execCallback} from 'node:child_process';
const exec = util.promisify(execCallback);
```

QualityControl/lib/services/Status.service.js
```diff
@@ -0,0 +1,136 @@
...
+   * Method to execute QC version retrieval command
+   * @returns {string} - version of QC deployed on the system
+   */
+  retrieveQcVersion() {
+    let version = 'Not part of an FLP deployment';
+    if (this._config.qc?.enabled) {
+      try {
+        version = execSync(QC_VERSION_EXEC_COMMAND, { timeout: 3000 }).toString().trim();
+      } catch (error) {
+        this._logger.errorMessage(error, { level: 99, system: 'GUI', facility: 'status-service' });
```
martinboulais, line 115: You still return `ok` here?

martinboulais (changes requested):

QualityControl/lib/services/Status.service.js
graduta: Yes, for now it is better to return `ok` because QCG only benefits of 2 states. `green`(success) or `red` (error) and there is no `grey(invalid/not-configured)` like in AliECS GUI. This will come at a later stage.
This is to prevent oncallers being asked to fix an error that it actually does not exist

QualityControl/lib/services/Status.service.js
graduta: Of course, thank you for pointing it out.
Updated

martinboulais (approved):

PR #2117, opened 2023-09-08
[OGUI-1409] Add bkp service for past runs retrieval

graduta:
- [x] PR labels are selected

PR which:
* adds a new HTTP service to retrieve runs information from Bookkeeping


Control/lib/adapters/RunSummaryAdapter.js
```diff
@@ -0,0 +1,54 @@
...
+      startTime,
+      endTime,
+      detectors = [],
+    } = run;
+    return  {
+      runNumber,
+      environmentId,
+      definition,
+      calibrationStatus,
+      runType: runType?.name ?? '-',
```
martinboulais: I think except for display, null value should stay null, because it carries more information than a string

graduta: Yes, you are right, in case of missing run_type name I want to be able to throw error in some cases.
Updated accordingly

martinboulais (approved):

PR #2164, opened 2023-10-10
[OGUI-1414] Retrieve initial data needed for quering and building calibration page

graduta:
- [x] PR labels are selected

PR which:
* extracts constants used in KV store retrieval of information so that they can be reused by services
* adds new `CalibrationRunService` which will be the high-level service used for building data needed for the Calibration Page
* moves the storing of `runTypes` from `BookkeepingService` so that it is kept as a low service
* adds new enum file for storing `logLevels`


Control/lib/services/Bookkeeping.service.js
```diff
@@ -33,40 +33,29 @@ class BookkeepingService {
...
-   */
-  async init() {
-    this._runTypes = await this._getRunTypes();
-  }
-
   /**
    * Given a definition, a type of a run and a detector, fetch from Bookkeeping the last RUN matching the parameters
    * @param {String} definition - definition of the run to query
-   * @param {String} type - type of the run to query
+   * @param {Number} type - type of the run to query as per BKP mapping
```
martinboulais: I would put typeId if it's the id

Control/lib/services/CalibrationRun.service.js
```diff
@@ -0,0 +1,102 @@
...
+      /**
+       * @type {Object<String, Array<String>>}
+       * @example 
+       * { "XYZ": ["CALIB1", "CALIB2"], "ABC": ["XCALIB"] }
+       */
+      this._calibrationPerDetectorMap = await this._apricotService.getRuntimeEntryByComponent(COG, CALIBRATION_MAPPING);
+    } catch (error) {
+      const err = grpcErrorToNativeError(error);
+      this._logger.errorMessage(`Unable to load calibration mapping due to: ${err}`,
+        {level: 5, system: 'GUI', facility: 'calibration-service'}
```
martinboulais: Don't we have constants for level? We should

martinboulais (changes requested):

Control/lib/services/Bookkeeping.service.js
graduta: Updated, I missed it when migrating from String to Number
Thank you!

Control/lib/services/CalibrationRun.service.js
graduta: Yes, we should and the plan is to add them in the `webui/framework` with the new logger library.
For now, I have added them as constants

martinboulais (approved):

PR #2179, opened 2023-10-27
[OGUI-1433] Update BookkeepingFilter for more generic one

graduta:
- [x] PR labels are selected

During development of CalibrationPage it has been needed to adapt BookkeepingFilter multiple times. Thus, this PR:
* adds an adapter for building a Bookkeeping accepted filter based on given object
* adapts the service that uses it and tests accordingly

### DIFF
--- a/Control/lib/adapters/external/BookkeepingFilterAdapter.js
+++ b/Control/lib/adapters/external/BookkeepingFilterAdapter.js
@@ -1,0 +1,89 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
+ * distributed under the terms of the GNU General Public License v3 (GPL
+ * Version 3), copied verbatim in the file "COPYING".
+ *
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * @class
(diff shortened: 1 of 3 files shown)


Control/lib/adapters/external/BookkeepingFilterAdapter.js
```diff
@@ -0,0 +1,89 @@
...
+   */
+  static toString(filterMap) {
+    const {detectors, runTypes, definitions, calibrationStatuses, limit = 1} = filterMap;
+    let filter = `page[limit]=${limit}`;
+
+    if (detectors) {
+      let operator = 'and';
+      let values = '';
+      if (typeof detectors === 'object') {
+        if (detectors?.operator === 'and' || detectors?.operator === 'or') {
```
martinboulais: I think you don't need optional chaining here because of the `typeof`

Control/lib/adapters/external/BookkeepingFilterAdapter.js
```diff
@@ -0,0 +1,89 @@
...
+          operator = detectors.operator;
+        }
+        if (detectors.values) {
+          values = detectors.values;
+        }
+      } else {
+        values = detectors;
+      }
+      values = BookkeepingFilterAdapter._parseValueIntoString(values);
```
martinboulais: I personally dislike reusing the same variable for different usage, for example conversion from array to string. It's harder to understand what type we have at a given point

martinboulais (changes requested):

Control/lib/adapters/external/BookkeepingFilterAdapter.js
graduta: Added new const with suggestive name

martinboulais (approved):

PR #2204, opened 2023-11-21
[OGUI-1442] Use new PUT endpoint for controlling environment

graduta:
- [x] PR labels are selected

* Update front-end to use the new PUT endpoint for controlling an environment
* Update method to accept specified parameters rather than an undescribed object

### DIFF
--- a/Control/public/utilities/jsonFetch.js
+++ b/Control/public/utilities/jsonFetch.js
@@ -11,47 +11,34 @@
  * or submit itself to any jurisdiction.
  */
 
-import { fetchClient } from '/js/src/index.js';
+import {fetchClient} from '/js/src/index.js';
 
 /**
- * @source {Bookkeeping} - https://github.com/AliceO2Group/Bookkeeping/blob/main/lib/public/utilities/fetch/jsonFetch.js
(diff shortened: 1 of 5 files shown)


Control/public/utilities/jsonFetch.js
```diff
@@ -11,47 +11,34 @@
...
+  } catch (error) {
+    return Promise.reject({message: 'Connection to server failed, please try again'});
   }
-  if (result === null || result) {
-    return result;
+  try {
+    const result = response.status === 204 // case in which response is empty
+      ? null
+      : await response.json();
+    return response.ok
```
martinboulais, line 38: It's weird to check for `ok` after `json` no?

martinboulais (changes requested):

Control/public/utilities/jsonFetch.js
graduta: Yes, indeed it might not be clear. 
I wanted to avoid having to do `response.json()` twice because you still need to call `response.json` to extract the `message` field of the error even if the status is not ok

In this scenario status not ok is represented by HTTP status codes out of [200,400) range

martinboulais (approved):

PR #2219, opened 2023-12-11
[OGUI-1448] Refactoring of RemoteData

martinboulais:
#### I DON'T have JIRA ticket
- [ ] test are added

* Use ES6 sub-classes
* Deprecated methods that should not be used anymore
* Improve doc blocks using generics (typedef until typescript support)

### DIFF
--- a/Framework/Frontend/js/src/RemoteData.js
+++ b/Framework/Frontend/js/src/RemoteData.js
@@ -132,6 +132,14 @@ export class RemoteData {
   }
 
   /**
+   * Returns the kind of the remote data
+   * @deprecated use {@see RemoteData#match} or {@see RemoteData#apply}
+   */
+  get kind() {
+    throw new Error('Abstract function call');
+  }
+
+  /**
    * Test if current kind is a `NotAsked`
    * @return {boolean}
    * @deprecated use {@see RemoteData#match} or {@see RemoteData#apply} instead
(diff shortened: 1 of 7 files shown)


Framework/Frontend/js/src/RemoteData.js
```diff
@@ -54,7 +67,40 @@ class RemoteData {
...
+   * Return a new remote data (shallow copy) build from transformation on the current data or error
+   *
+   * @template NP the type of the transformed success data
+   * @template NE the type of the transformed error
+   *
+   * @param {Partial<{Success: (payload: P) => NP, Failure: (error: E) => NE}>} transformation the transformation to apply to
+   *     concrete data transform from the current remote data to the new one
+   * @return {RemoteData<NP, NE>} the resulting remote data
+   */
+  map(transformation) {
```
graduta: Similar to the ObservableData, I would be in favour of naming this function `apply` as you `Apply transformation`

graduta (changes requested):
`RemoteData` class is a core element of our interfaces, and a concept that is not always easily grasped by students.

Please make sure you add examples in this class. e.g. Other is a bit ignored but more importantly please update the documentation used by the students with real like examples: https://github.com/AliceO2Group/WebUi/blob/dev/Framework/docs/guide/async-calls.md#remotedata

Framework/Frontend/js/src/RemoteData.js
martinboulais: Done

graduta (approved):

PR #2252, opened 2024-02-01
[OGUI-1460] Add API endpoint for requesting to refresh calibration configuration as Global or Admin

graduta:
- [x] PR labels are selected

PR which:
* adds an endpoint for allowing certain roles of users to request a reload of configurations stored in-memory for the calibration pages

### DIFF
--- a/Control/lib/api.js
+++ b/Control/lib/api.js
@@ -97,7 +97,7 @@ module.exports.setup = (http, ws) => {
 
   const bkpService = new BookkeepingService(config.bookkeeping ?? {});
   const runService = new RunService(bkpService, apricotService, cacheService);
-  runService.load();
+  runService.loadStaticConfigurations();
   const runController = new RunController(runService, cacheService);
 
   const notificationService = new NotificationService(config.kafka);
--- a/Control/lib/api.js
+++ b/Control/lib/api.js
@@ -49,7 +49,7 @@ const O2_CONTROL_PROTO_PATH = path.join(__dirname, './../protobuf/o2control.prot
 const O2_APRICOT_PROTO_PATH = path.join(__dirname, './../protobuf/o2apricot.proto');
 
 const {Role} = require('./common/role.enum.js');
-const {meetsRoleCriteria} = require('./middleware/role.middleware.js');
+const {minimumRoleMiddleware} = require('./middleware/minimumRole.middleware.js');
 
 if (!config.grpc) {
   throw new Error('Control gRPC Configuration is missing');
@@ -130,7 +130,7 @@ module.exports.setup = (http, ws) => {
   http.get('/workflow/configuration', workflowController.getWorkflowConfiguration.bind(workflowController));
 
   http.get('/runs/calibration/config', [
-    meetsRoleCriteria(Role.GLOBAL)
+    minimumRoleMiddleware(Role.GLOBAL)
   ], runController.refreshCalibrationRunsConfigurationHandler.bind(runController));
   
   http.get('/runs/calibration', runController.getCalibrationRunsHandler.bind(runController))
--- a/Control/lib/api.js
+++ b/Control/lib/api.js
@@ -48,6 +48,9 @@ const path = require('path');
 const O2_CONTROL_PROTO_PATH = path.join(__dirname, './../protobuf/o2control.proto');
 const O2_APRICOT_PROTO_PATH = path.join(__dirname, './../protobuf/o2apricot.proto');
 
+const {Role} = require('./common/role.enum.js');
+const {meetsRoleCriteria} = require('./middleware/role.middleware.js');
+
 if (!config.grpc) {
   throw new Error('Control gRPC Configuration is missing');
 }
@@ -94,7 +97,7 @@ module.exports.setup = (http, ws) => {
 
   const bkpService = new BookkeepingService(config.bookkeeping ?? {});
   const runService = new RunService(bkpService, apricotService, cacheService);
-  runService.init();
+  runService.load();
   const runController = new RunController(runService, cacheService);
 
   const notificationService = new NotificationService(config.kafka);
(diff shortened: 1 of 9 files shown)


Control/lib/api.js
```diff
@@ -94,7 +97,7 @@ module.exports.setup = (http, ws) => {
 
   const bkpService = new BookkeepingService(config.bookkeeping ?? {});
   const runService = new RunService(bkpService, apricotService, cacheService);
-  runService.init();
+  runService.load();
```
martinboulais: Why do you have a constructor and a load function?

Control/lib/common/role.enum.js
```diff
@@ -0,0 +1,44 @@
...
+});
+
+/**
+ * Method to check if the provided access as per WebUI server is stronger than the minimum required role
+ * @param {String} role - admin, global, det-, guest
+ * @param {Role} minimumRole - minimum role that user should have
+ * @return {Boolean}
+ */
+const isWithinRole = (roleName, minimumRole) => {
+  const isDetectorRole = roleName.toLocaleUpperCase().startsWith('DET-');
```
martinboulais: I would create a variable to store the role in upper case, maybe reasigning the parameter variable

Control/lib/common/role.enum.js
```diff
@@ -0,0 +1,44 @@
...
+  ADMIN: 1,
+  GLOBAL: 10,
+  DETECTOR: 30, // roles for this level will be in format of `det-<name>` e.g. `det-its`
+  DEFAULT_ROLE: 50,
+  GUEST: 100
+});
+
+/**
+ * Method to check if the provided access as per WebUI server is stronger than the minimum required role
+ * @param {String} role - admin, global, det-, guest
```
martinboulais: roleName is the parameter. Thing is, you have `roleName` and `minimumRole`, I would rather use `role` and `minimumRole` or `roleName` and `minimumRoleName`. 
I would prefer `currentRole` than simply role though.

Control/lib/common/role.enum.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Role} minimumRole - minimum role that user should have
+ * @return {Boolean}
+ */
+const isWithinRole = (roleName, minimumRole) => {
```
martinboulais: The name is not clear for me, I would name it `isRoleSufficient` or `doesRoleMatchMinimumRequired` or `doesRoleFitRequirement` something like that

Control/lib/controllers/Run.controller.js
```diff
@@ -61,6 +62,23 @@ class RunController {
...
+  /**
+   * API - GET endpoint to request a re-init of the run service which will update calibration configurations
+   * The handler also must check that the user making the request is allowed for such operation
+   * @param {Request} _ - HTTP Request object
+   * @param {Response} res - HTTP Response object
+   * @returns {void}
+   */
+  async refreshCalibrationRunsConfigurationHandler(_, res) {
+    try {
+      await this._runService.load();
```
martinboulais: For me it's not the role of the controller to load the service

Control/lib/middleware/role.middleware.js
```diff
@@ -0,0 +1,45 @@
...
+const {isWithinRole} = require('./../common/role.enum.js');
+const {UnauthorizedAccessError} = require('./../errors/UnauthorizedAccessError.js');
+const {updateExpressResponseFromNativeError} = require('./../errors/updateExpressResponseFromNativeError.js');
+
+/**
+ * Method to receive a minimum role that needs to be met by owner of request and to return a middleware function
+ * @param {Role} minimumRole - minimum role that should be fulfilled by the requestor
+ * @return {function(req, res, next): void} - middleware function
+ */
+const meetsRoleCriteria = (minimumRole) => {
```
martinboulais: I would prefer a more explicit name, such as `minimumRoleLevelMiddleware`

If you have only one export, the file name should be this export (if you keep the file name, export should be `roleMiddleware` which is good too)

martinboulais (changes requested):

Control/lib/api.js
graduta: Constructors cannot await for services. 
Moreover, you do not always start loading needed data from constructor. Usually you have the initializer (constructor) and a load function that loads in-memory what data you need. 
Something in line with `constructor` and `ngOnInit` from Angular

Control/lib/common/role.enum.js
graduta: I agree. Updated

Control/lib/common/role.enum.js
graduta: The variable names were chosen specifically for this:
`roleName` - refers to the name of the role as a `String`
`role` - is the local variable which will store the `Role` as per the provided string
`minimumRole` - is the same, of type `Role`

graduta: While for the parameter missspell I updated to `roleName`. Thank you for pointing it out

Control/lib/common/role.enum.js
graduta: Agree

Control/lib/controllers/Run.controller.js
graduta: The purpose of the controller here is to handle the request of an admin/global user that requests the reload of the configuration which is covered by the `load` function. 
What would you propose otherwise?

Control/lib/middleware/role.middleware.js
graduta: I think `roleMiddleware` is not enough to suggest that a minimum role should be achieved.
Updated to `minimumRoleMiddleware`

Control/lib/controllers/Run.controller.js
martinboulais: I would say that the service should load what it needs when you call its functions

graduta: The service already loads what it needs from the in-memory service. 
But there are very rare occurrences when `admins` want to refresh a configuration that was changed in Consul. That is why we need to `re-load` the configuration

martinboulais (dismissed):

PR #2254, opened 2024-02-05
[OGUI-1462] Add reload button for calibration static configuration setup

graduta:
- [x] PR labels are selected

PR which:
* adds a button to the front-end which sends a request to retrieve the static calibration configuration setup;
* button is displayed only for admin/global roles


Control/public/pages/CalibrationRuns/CalibrationRuns.model.js
```diff
@@ -59,6 +59,26 @@ export class CalibrationRunsModel extends Observable {
     this.notify();
   }
 
+
+  /**
+   * Request a reset of the calibration page including the calibration configurations and reload
+   * the calibration runs if the reset was successful
+   * @return {void}
+   */
+  async requestReload() {
```
martinboulais: For me it feels weird to name it `requestReload` I would directly name it `reload`, `refresh`

martinboulais (changes requested):

Control/public/pages/CalibrationRuns/CalibrationRuns.model.js
graduta: I thought about that but then I used `requestReload` as:
- not any user has access to reload, and this is a request verified by the server and only if successful it will start the process.
- reload/refresh I thought it might incline the developer thinking it is a refresh or reload of the page.

Perhaps `refreshCalibrationConfigurations` instead?

martinboulais: Yes I prefer `refreshCalibrationConfigurations`

> not any user has access to reload, and this is a request verified by the server and only if successful it will start the process.
In that case the function should return and the button should not be available no?

graduta: Renamed

martinboulais (approved):

PR #2494, opened 2024-07-19
[OGUI-1532] Create enums for log level and severity

martinboulais:
- [X] PR labels are selected

### DIFF
--- a/Framework/Backend/log/InfoLoggerMessage.js
+++ b/Framework/Backend/log/InfoLoggerMessage.js
@@ -10,10 +10,10 @@
  * In applying this license CERN does not waive the privileges and immunities
  * granted to it by virtue of its status as an Intergovernmental Organization
  * or submit itself to any jurisdiction.
-*/
+ */
 
-const {LogLevel} = require("./LogLevel");
-const {LOG_SEVERITIES, LogSeverity} = require("./LogSeverity");
+const {LogLevel} = require('./LogLevel');
+const {LOG_SEVERITIES, LogSeverity} = require('./LogSeverity');
 
 /**
  * TypeDefinition for InfoLoggerMessage Object
@@ -30,7 +30,7 @@ class InfoLoggerMessage {
     this._severity = LogSeverity.Info;
     this._level = LogLevel.Developer;
     this._system = 'GUI';
-    this._facility = 'gui'
+    this._facility = 'gui';
     this._partition = undefined;
     this._run = undefined;
     this._errorSource = undefined;
@@ -62,7 +62,7 @@ class InfoLoggerMessage {
   getComponentsOfMessage() {
     const components = [
       `-oSeverity=${this._severity}`, `-oLevel=${this._level}`,
-      `-oSystem=${this._system}`, `-oFacility=${this._facility}`
+      `-oSystem=${this._system}`, `-oFacility=${this._facility}`,
     ];
     if (this._partition) {
       components.push(`-oPartition=${this._partition}`);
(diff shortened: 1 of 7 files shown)


Framework/Backend/log/InfoLoggerMessage.js
```diff
@@ -35,14 +38,13 @@ class InfoLoggerMessage {
...
+   * @param {object} logJson - object with all or partial InfoLoggerMessage fields
    * @returns {InfoLoggerMessage}
    */
   static fromJSON(logJson) {
     const log = new InfoLoggerMessage();
-    log._severity = logJson.severity && ['Info', 'Error', 'Fatal', 'Warning', 'Debug'].includes(logJson.severity) ?
-      logJson.severity : 'Info';
-    log._level = Number.isInteger(logJson?.level) ? logJson.level : 11;
+    log._severity = logJson.severity && LOG_SEVERITIES.includes(logJson.severity) ? logJson.severity : LogSeverity.Info;
+    log._level = Number.isInteger(logJson?.level) ? logJson.level : LogLevel.Developer;
```
graduta, line 47: Should it not attempt to check if the provided value can be converted to a Number? Just in case a developer provides a string representation of a number?

Framework/Backend/log/LogLevel.js
```diff
@@ -0,0 +1,26 @@
...
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * Enumeration of Log levels defining the audience
```
graduta, line 16: Please add link to InfoLogger definition of levels

Framework/Backend/log/InfoLoggerSender.js
```diff
@@ -15,19 +15,21 @@
...
   /**
-   * @param {winston.instance} winston - local winston instance object
+   * @param {Logger} winston - local winston instance object
    */
-  constructor(winston, label = '') {
+  constructor(winston) {
     this._isConfigured = false;
     this.winston = winston;
-    this.label = label;
+    const label = 'infologger';
```
graduta: I am not sure that this should default to 'infologger' without option to change it or provide your own label for winston.
With this change my logs will look like this:
```
2024-07-19T09:18:21.706Z [cog/server] info: Server listening on port 8080
2024-07-19T09:18:21.707Z [infologger] debug: InfoLogger executable not found
```
I think it should still be `[cog/infologgersender]` or some user preferred label.

graduta (changes requested):

Framework/Backend/log/InfoLoggerMessage.js
martinboulais: For me, in library we should expect users to provide proper type when it's properly typed (users means developpers using library, not end-users of course)

Framework/Backend/log/LogLevel.js
martinboulais: Done

Framework/Backend/log/InfoLoggerSender.js
martinboulais: Do you instanciate `InfologgerSender` in other places than Logger?
Because if not, previous behavior was using `this.label` from static function which was `undefined` if I am not mistaken

graduta: I do not think it was undefined but this is because there was a hidden default value in the WinstonWrapper which tricked me

graduta (approved):

PR #2624, opened 2024-10-10
Add kafka messages consumer to Framework

martinboulais:
#### I DON'T have JIRA ticket
- [X] test are added

Added a class to listen to ECS kafka messages and automatically parse them using protobuf


Framework/Backend/protobuf/getWebUiProtoIncludeDir.js
graduta, line 1: Licence please

Framework/Backend/kafka/KafkaMessagesConsumer.js
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Call every registered listeners by passing the given message to it
+   *
+   * @param {T} message the message to pass to listeners
+   * @return {void}
+   */
+  async _handleEvent(message) {
+    for (const listener of this._listeners) {
+      try {
+        await listener(message);
```
graduta: What do you think in case of having lots of listeners, to use `.catch()` and remove the await so that listeners will not await in sequence

graduta (changes requested):

Framework/Backend/kafka/KafkaMessagesConsumer.js
martinboulais: Indeed, fixed

Framework/Backend/protobuf/getWebUiProtoIncludeDir.js
martinboulais: Added

graduta (approved):

PR #2629, opened 2024-10-16
[OGUI-1556] Allow modification of existing layouts via JSON panel

mariscalromeroalejandro:
- [X] PR labels are selected

* Add option to layout view to edit the layout schema via JSON
* Added tests to this solution


QualityControl/public/layout/view/header.js
```diff
@@ -61,10 +61,24 @@ const toolbarViewMode = (model) => {
...
+            h('.dropdown-menu', { style: {
+              right: '0',
+              left: 'auto',
+            } }, [
+              h('p.m3.mv2.text-ellipsis', [
+                h('a.menu-item', { title: 'Edit via GUI', onclick: () => model.layout.edit() }, 'Edit via GUI'),
+                h('a.menu-item', {
+                  title: 'Edit via JSON',
+                  onclick: () => openJSONEditModal(model),
+                }, 'Edit via JSON'),
```
graduta, line 75: After clicking on the `Edit via JSON` and the modal appears, I expect the dropdown from the `edit` button to disappear as well.

QualityControl/public/layout/view/header.js
```diff
@@ -61,10 +61,24 @@ const toolbarViewMode = (model) => {
...
-          h('button.btn.btn-primary', {
-            onclick: () => model.layout.edit(),
+          h('.dropdown', {
             title: 'Edit layout',
-          }, iconPencil()),
+            class: model.layout.editMenuOpen ? 'dropdown-open' : '',
+          }, [
+            h('button.btn.btn-primary', { onclick: () => model.layout.toggleEditMenu() }, iconPencil()),
+            h('.dropdown-menu', { style: {
+              right: '0',
```
graduta: This style properties should be defined in a CSS class in `app.css` or even better reuse the existing ones such as `.dropdown`. See how `info` button of the objects are using this.

QualityControl/public/layout/view/header.js
```diff
@@ -61,10 +61,24 @@ const toolbarViewMode = (model) => {
...
+              left: 'auto',
+            } }, [
+              h('p.m3.mv2.text-ellipsis', [
+                h('a.menu-item', { title: 'Edit via GUI', onclick: () => model.layout.edit() }, 'Edit via GUI'),
+                h('a.menu-item', {
+                  title: 'Edit via JSON',
+                  onclick: () => openJSONEditModal(model),
```
graduta:
```suggestion
                  onclick: () => {
                    model.isUpdateVisible = true;
                  }
```

Functionality is defined in the model and not in the component. You should call directly from your onclick function

QualityControl/test/setup/seeders/qcg-mock-data-template.json
```diff
@@ -1,139 +1,139 @@
...
- "users": [
-  {
-   "id": 0,
-   "name": "Anonymous",
-   "username": "anonymous"
-  }
- ]
-}
\ No newline at end of file
+    "layouts": [
```
graduta, line 2: As a standard in WebUI, we use `2 space` and not `4 space` for tabbing. Please update your VS code setting and revert this file change

QualityControl/test/public/pages/layout-show.test.js
```diff
@@ -158,72 +159,5 @@ export const layoutShowTests = async (url, page, timeout = 5000, testParent) =>
...
-  await testParent.test(
-    'should show normal sidebar after Cancel click',
-    { timeout },
-    async () => {
-      const cancelButtonPath = 'header > div > div:nth-child(3) > div > button:nth-child(2)';
-      await page.locator(cancelButtonPath).click();
-      await page.waitForSelector('nav .menu-title', { timeout: 5000 });
-    },
-  );
+  await editLayoutTests(page, timeout, testParent);
```
graduta: Only `test-index` should import other test files. Moreover the `editLayoutTests` are still part of the page layoutShow. Thus, the naming of the test file should provide us with clear information on that.
I would recommend for this scenario to keep them in the same file otherwise I cannot track which test was updated/added/removed.

QualityControl/public/layout/Layout.js
```diff
@@ -727,4 +739,95 @@ export default class Layout extends Observable {
       this.selectTab(this._tabIndex);
     }
   }
+
+  /**
+   * Returns the updated layout in a formatted JSON string.
+   * @returns {string} The updated JSON string representing the layout.
+   */
+  getUpdatedLayout() {
```
graduta: What does `getUpdatedLayout` refer to? What does it update? I see you are using it to display the initial value in the modal. It would be good to improve on the name of the function to suggest that. Something along the lines: `getLayoutInEditJSONStructure` and in the method doc, you mention it removes the IDs from tabs, objects and layout

QualityControl/public/layout/panels/editModal.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param {Model} model - root model of the application
+ * @returns {vnode} - virtual node element
+ */
+export default (model) => h('.o2-modal', [
+  h('.o2-modal-content', [
+    h('.p2.text-center.flex-column', [
+      h('h4.pv1', 'Edit JSON file of a layout'),
+      h('', h('textarea.form-control.w-100', {
+        rows: 15,
+        oninput: (e) => model.layout.checkLayoutToUpdate(e.target.value),
```
graduta, line 28: I think the validation should happen only when the user clicks to update the layout. Otherwise, if they make changes manually, it will become quite frustrating to constantly see the error message popping up

QualityControl/public/layout/Layout.js
```diff
@@ -727,4 +739,95 @@ export default class Layout extends Observable {
...
+      this.changeUpdateStatus(RemoteData.failure(error.message || error));
+    }
+    this.notify();
+  }
+
+  /**
+   * Merges skeleton to the layout
+   * @returns {object} item updated with user changes
+   */
+  getUpdatedItem() {
```
graduta: I would recommend this method you remove and integrate it where it is used above

QualityControl/public/layout/Layout.js
```diff
@@ -727,4 +739,95 @@ export default class Layout extends Observable {
...
+    this.changeUpdateStatus(RemoteData.Success());
+    this.toggleUpdatePanel();
+  }
+
+  /**
+   * Sets the status of the layout update
+   * @param {RemoteData} status new layout update status
+   */
+  changeUpdateStatus(status) {
+    this.model.services.update = status;
```
graduta: There is no `update` service in QCG. Where does this come from?

QualityControl/public/layout/Layout.js
```diff
@@ -727,4 +739,95 @@ export default class Layout extends Observable {
...
+    return {
+      ...updatedLayout,
+      id: this.item.id,
+    };
+  }
+
+  /**
+   * Saves the updated item and manages the update status.
+   */
+  saveUpdatedItem() {
```
graduta: Why do you duplicate the `save` method instead of using the existing `save` one?

graduta (changes requested):
Please have a look at the comments and let me know where it is not clear. We can then have a go together in pair-programming to address the issues

QualityControl/public/layout/Layout.js
graduta: I am referring to the fact that you declared a new method `saveUpdatedItem` instead of directly using the `save` one

QualityControl/public/layout/panels/editModal.js
mariscalromeroalejandro: The behavior of this modal has been copied from the import schema via JSON modal

graduta (approved):

PR #2665, opened 2024-11-19
[OGUI-1568] Filter issues filtering objects on edit mode layout show

mariscalromeroalejandro:
- [x] PR labels are selected

PR which resolves issues when loading objects in edit mode for the layout show page:
* fixes virtual table reliance on model. Now acts as a reusable component which recieves objects that are to be displayed.
* adapts objectTree and objectTreeSidebar usage of virtual table to pass array of objects to display
* Adds tests to validate the fix and ensure functionality.

### DIFF
--- a/QualityControl/public/layout/view/header.js
+++ b/QualityControl/public/layout/view/header.js
@@ -62,7 +62,7 @@ const toolbarViewMode = (model) => {
         }, iconShareBoxed()),
         model.session.personid == layoutItem.owner_id && [
           h('button.btn.btn-primary', {
-            onclick: async () => await model.layout.edit(),
+            onclick: async () => model.layout.edit(),
             title: 'Edit layout',
           }, iconPencil()),
           h('button.btn.btn-danger', {
--- a/QualityControl/public/layout/view/header.js
+++ b/QualityControl/public/layout/view/header.js
@@ -62,7 +62,7 @@ const toolbarViewMode = (model) => {
         }, iconShareBoxed()),
         model.session.personid == layoutItem.owner_id && [
           h('button.btn.btn-primary', {
-            onclick: async () => model.layout.edit(),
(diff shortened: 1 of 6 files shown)


QualityControl/public/layout/view/header.js
```diff
@@ -62,7 +62,7 @@ const toolbarViewMode = (model) => {
         }, iconShareBoxed()),
         model.session.personid == layoutItem.owner_id && [
           h('button.btn.btn-primary', {
-            onclick: () => model.layout.edit(),
+            onclick: async () => await model.layout.edit(),
```
graduta: This is an event trigger, there is no need to await for it as nothing else is done after.
It is enough to just fire the call and forget about it.

QualityControl/public/object/virtualTable.js
```diff
@@ -29,30 +29,38 @@ export default function virtualTable(model, location = 'main') {
   return h('.flex-grow.flex-column', {
   }, [
     location !== 'side' && tableHeader(),
-    h(
-      '.scroll-y.animate-width',
-      tableContainerHooks(model),
+    model.object.searchResult.length === 0 ? tableEmpty() :
```
graduta: Keep in mind that `virtualTable` is a reused component by other plages.
If you look one line above, you will see that the `tableHeader` is added already under a condition which checks the page location. 

Now because you add another table, on object tree page, we will have two tables.

The standard is that the `No objects found for this search` message in case of no data matching, is that it should be added as a row and not as a header (check screenshot to see how it looks with these changes on the object tree page)

image

graduta (changes requested):
Hi, great start for fixing this issue. Once you update the creation of new table issue, the PR will be good to go!

graduta (approved):

PR #2667, opened 2024-11-20
[OGUI-1578] Added vscode debugging configuration in launch.js

Houwie7000:
- [x] PR labels are selected

Added vscode debugging configuration in launch.js
.gitignore exception made for .vscode/launch.js.
Added --inspect parameter to npm run dev command to enable node debugging from the browser.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,23 +1,23 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "Run dev",
+            "name": "[ilg] run dev",
             "request": "launch",
             "runtimeArgs": [
                 "run-script",
                 "dev"
             ],
-            "localRoot": "${workspaceFolder}\\InfoLogger\\",
+            "localRoot": "${workspaceFolder}/InfoLogger/",
             "runtimeExecutable": "npm",
             "skipFiles": [
                 "<node_internals>/**"
             ],
             "outputCapture": "std",
-            "type": "node"
+            "type": "node",
+            "windows": {
+                "localRoot": "${workspaceFolder}\\InfoLogger\\"
+            }
         }
     ]
 }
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
graduta: I believe these comments can be removed

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
...
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Run dev",
```
graduta: As the file for all configurations sits at the root of the directory, it means it covers all 4 GUIs present in this repository. Thus, I would like for the `name` filed of the configurations to be a bit more specific. For example, the one you added is for InfoLogger so the name should inform the developer about it. 
Perhaps a more sensible name would be: `[ilg] run dev`
`[ilg]` stands for InfoLogger GUI and it is a common label used within the team to refer to this project

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
...
+            "request": "launch",
+            "runtimeArgs": [
+                "run-script",
+                "dev"
+            ],
+            "localRoot": "${workspaceFolder}\\InfoLogger\\",
```
graduta: I see that you have used the `Windows` path for localRoot. This of course will not work for unix based operating systems.
Could you please create a configuration for windows and one for unix based systems in which the forward slash is used for the path?
Ideally, the name of the configuration should also reflect this

graduta (changes requested):
Hi, Jasper! Thank you for adding a configuration to quickly start a server in dev mode. Please see my comments on some changes that I think need to be done before approving this PR. 

One more general comment, please make sure to add a new line at the end of the files when you commit. This is checked by the lint process by also by GitHub if you check the files in  your PR changes.(attaching here a small screenshot as well)
image

Houwie7000:
I've resolved the feedback, the config now works on *nix like platforms and on Windows.

graduta (approved):

PR #2683, opened 2024-12-02
[OGUI-1580] add middleware for detector matching user role

pepijndik:
Middleware for checking if a user has a matching detector role
* Added User belongsToDetector

* DetectorOwnership middleware
* DetectorOwnershipMiddleware tests

- [x] PR labels are selected


Control/lib/middleware/detectorOwnership.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+  const { detectorId } = req.params;
+  const { name, username, personid, access } = req.session || {};
+
+  if (!detectorId || !access) {
+    return res.status(400).json({ message: 'Invalid request: missing information' });
+  }
+  
+  try {
+    const u = new User(username, name, personid, access);
+    console.log(`Checking locks for detector ${detectorId}`);
```
graduta: Hi, @pepijndik . We do not make use of `console` in this project but instead we use the `Logger` defined in the WebUI framework.
Nevertheless, in this case there is no need to log anything here in case of success.

EDIT: I have seen there are some `console.errors` in lockOwnership middleware and I confirm those are not correct and we should not be using it like this. I apologise if this is what tricked you into using it.

Control/lib/middleware/detectorOwnership.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+const detectorOwnershipMiddleware = (req, res, next) => {
+  const { detectorId } = req.params;
+  const { name, username, personid, access } = req.session || {};
+
+  if (!detectorId || !access) {
+    return res.status(400).json({ message: 'Invalid request: missing information' });
+  }
+  
+  try {
+    const u = new User(username, name, personid, access);
```
graduta: Please refrain from using single-letter variables and instead use more descriptive ones

Control/lib/middleware/detectorOwnership.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+    console.log(`Checking locks for detector ${detectorId}`);
+    if (!u.belongsToDetector(detectorId)) {
+      return res
+        .status(403)
+        .json({ message: `User ${name} does not have ownership of the lock for detector ${detectorId}` });
+    }
+  
+    next(); // Proceed if lock ownership is verified
+  } catch (error) {
+    console.error(`Error checking locks:`, error);
```
graduta: The same as above, we do not use the `console` but instead the web-ui logger.

Control/lib/middleware/detectorOwnership.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (!detectorId || !access) {
+    return res.status(400).json({ message: 'Invalid request: missing information' });
+  }
+  
+  try {
+    const u = new User(username, name, personid, access);
+    console.log(`Checking locks for detector ${detectorId}`);
+    if (!u.belongsToDetector(detectorId)) {
+      return res
```
graduta: Rather than straight using the `res`, I would recommend you use the ` updateExpressResponseFromNativeError` with an `UnauthorizedError`

Control/lib/middleware/detectorOwnership.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+        .status(403)
+        .json({ message: `User ${name} does not have ownership of the lock for detector ${detectorId}` });
+    }
+  
+    next(); // Proceed if lock ownership is verified
+  } catch (error) {
+    console.error(`Error checking locks:`, error);
+    res.status(500).json({ message: 'Internal server error' });
```
graduta: Similar to above, please make use of ` updateExpressResponseFromNativeError`

Control/test/lib/middleware/mocha-detectorOwnership.middleware.test.js
```diff
@@ -0,0 +1,100 @@
...
+    const next = sinon.stub();
+
+    detectorOwnershipMiddleware(req, res, next);
+
+    assert.ok(res.status.calledWith(400));
+    assert.ok(res.json.calledWith({ message: 'Invalid request: missing information' }));
+    assert.ok(next.notCalled);
+  });
+
+  it('should integrate with an Express route and add detectorId', async () => {
```
graduta: This test does not follow the standard and moreover, it defines a new server just for the purpose of this middleware. In time, it can create confusion for developers. I would recommend to remove it.

I would encourage you to have a look at `api-get-locks.test.js` for example if you would like to see how to add tests to the API and make use of the `request` library to test the server API endpoint.
Of course, this would go into the api test directory

graduta (changes requested):

Control/lib/middleware/detectorOwnership.middleware.js
```diff
@@ -0,0 +1,33 @@
...
+  const { detectorId } = req.params;
+  const { name, username, personid, access } = req.session || {};
+
+  if (!detectorId || !access) {
+    updateExpressResponseFromNativeError(res, new UnauthorizedAccessError('Invalid request: missing information')); 
+  }
+  
+  try {
+    const user = new User(username, name, personid, access);
+    if (!user.belongsToDetector(detectorId)) {
```
graduta: Hi @pepijndik . Very good first step for the PR!
There is one thing that we did not discuss and I did not explain in the ticket. We should take into consideration the scenario in which:
- a user with a role higher than `Detector` is requesting the lock. In this scenario, they will not have the `det-<name>` permission but instead they will have a higher role. In this case, we should ignore the check of this middleware function

I am thinking before this `if` condition, you have another one in which you check that the role of the user is sufficient to go `next()`. Have a look at the function `isRoleSufficient`. You should check that the current role of user is at least `Role.GLOBAL`. If so, then you do not need to check if the user belongs to the detector

graduta (changes requested):

graduta (approved):

PR #2684, opened 2024-12-02
[OGUI-1455] Logging live mode filters/criteria

Houwie7000:
- [x] PR labels are selected

### DIFF
--- a/Framework/Backend/websocket/server.js
+++ b/Framework/Backend/websocket/server.js
@@ -240,7 +240,7 @@ class WebSocket {
    * @param {object} client - disconnected client
    */
   onclose(client) {
-    this.logger.info(`ID ${client.id} Client disconnected`);
+    this.logger.debugMessage(`ID ${client.id} Client disconnected`);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


Framework/Backend/websocket/server.js
```diff
@@ -133,6 +133,42 @@ class WebSocket {
     client.on('error', (err) => this.logger.error(`Connection ${err.code}`));
   }
 
+  /**
+   * Make criteria more readable.
+   * This code is a close copy of InfoLogger/public/logFilter/LogFilter.js LN 101 toObject()
+   * @param {object} criteria - criteria to be minified
+   * @returns {object} minimal filter object
+   */
+  minifyCriteria(criteria) {
```
graduta: @Houwie7000  as this is a utility function, I think it is better if we move it to an independent JS module in an utility folder.
As currently there is no such folder, please create one and add this function to its own module. Then call it from there in the WS server.
The tests associated to this function should match the folder structure as well

Framework/Backend/websocket/server.js
```diff
@@ -133,6 +133,42 @@ class WebSocket {
...
+    // Copy everything
+    const criterias = JSON.parse(JSON.stringify(criteria));
```
graduta: Currently this function `minifyCriteria` is called form the WS server directly and there is no `try-catch` associated to it which means if the criteria is in a bad format it will crash the server.
I would recommend you add this function in a try catch and log in case of error that the payload criteria is not a correct format. We should not log the payload (as it is user input) as it might contain an injection attack

graduta (changes requested):

PR #2691, opened 2024-12-05
[OGUI-1472] Run types filter selector for layout show view

mariscalromeroalejandro:
- [X] PR labels are selected

PR which:
* Adds new filter selector for run types in layout show view
* Retrieves run types from bookkeeping
* Creates a filter selector component to be able to reuse it
* Allow user to configure Bookkeeping


QualityControl/README.md
graduta, line 1: This might need to be updated based on new DOCS you have written

QualityControl/lib/api.js
```diff
@@ -44,7 +44,7 @@ export const setup = (http, ws) => {
     statusService,
     userController,
     layoutRepository,
-    jsonFileService,
+    jsonFileService, filterController,
```
graduta: Perhaps better on a new line like the others?

QualityControl/lib/api.js
```diff
@@ -86,4 +86,6 @@ export const setup = (http, ws) => {
   );
 
   http.get('/checkUser', userController.addUserHandler.bind(userController));
+
+  http.get('/runTypes', filterController.getRunTypesHandler.bind(filterController));
```
graduta: Given that they are related to the filter model, I think is best we add something more general which will allow other filter parameters?
`/filter/configuration`. Same for the controller method name

QualityControl/lib/QCModel.js
```diff
@@ -67,18 +69,41 @@ export const setupQcModel = () => {
...
   const intervalsService = new IntervalsService();
 
-  intervalsService.register(
-    qcObjectService.refreshCache.bind(qcObjectService),
-    qcObjectService.getCacheRefreshRate(),
-  );
+  const bookkeepingService = new BookkeepingService(config.bookkeeping);
+  bookkeepingService.retrieveRunTypes();
+
+  const filterController = new FilterController(bookkeepingService);
```
graduta: The filter controller should receive a filter service (high-level service) which in the future will build the filter based on other service as well.
The Bookkeeping service should return the data as stored in BKP and the Filter service should be the one mapping the list of runTypes recieved to a list.

QualityControl/lib/controllers/FilterController.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * HTTP GET endpoint for retrieving a list of run types from Bookkeeping
+   * @param {Request} req - HTTP request
+   * @param {Response} res - HTTP response to provide run types information
+   */
+  async getRunTypesHandler(req, res) {
+    try {
+      if (!this._bkpService) {
```
graduta: We should not throw an error but instead return an empty list of run types. Of course, the logic being in the FIlter Service

QualityControl/lib/services/BookkeepingService.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+export class BookkeepingService {
+  constructor({ url, token, refreshRate }) {
+    this._url = url;
+    const { protocol, hostname, port } = url ? new URL(this._url) : {};
+    this._hostname = hostname;
+    this._port = port;
+    this._protocol = protocol;
+    this._token = token;
+    if (!this._hostname || !this._port || !this._protocol) {
```
graduta: It is important to notice that our URLs are not always in the format `hostname:port`, but sometimes is just a simple URL like `alice-bkp.cern.ch`
Thus, you should not throw an error if you cannot export the port but instead modify the `httpGetJson` method to be able to make a request even if you pass just a URL instead of hostname and port. 
It is standard convention that for HTTPS port is 443 and HTTP is 80

graduta (changes requested):

QualityControl/docs/Configuration.md
```diff
@@ -41,6 +41,14 @@ ccdb: {
 }
 ```
 
+### BOOKKEEPING
+Attribute to define the `Bookkeeping` endpoint.
```
graduta: Missing attribute that encapsulates the attributes below. Perhaps providing an example object can help (with non-existent URL and token)

QualityControl/lib/QCModel.js
```diff
@@ -35,6 +37,7 @@ import { UserRepository } from './repositories/UserRepository.js';
 import { ChartRepository } from './repositories/ChartRepository.js';
 import { initDatabase } from './database/index.js';
 import { SequelizeDatabase } from './database/SequelizeDatabase.js';
+import { FilterService } from './services/FilterService.js';
```
graduta: Please import the FilterService in the services categories (grouped by lines above around line 26)

QualityControl/lib/controllers/FilterController.js
```diff
@@ -0,0 +1,44 @@
...
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * Gateaway class to be used to retrieve data with regard to filters
+ */
+export class FilterController {
+  constructor(filterService) {
+    /**
+     * @type {BookkeepingService}
```
graduta: Wrong type

QualityControl/lib/controllers/FilterController.js
```diff
@@ -0,0 +1,44 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * Gateaway class to be used to retrieve data with regard to filters
+ */
+export class FilterController {
+  constructor(filterService) {
```
graduta, line 23: Missing docs for constructor

QualityControl/lib/services/FilterService.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * High level service that composes, processes and maps data from the bookkeeping service
+ */
+export class FilterService {
+  /**
+   * Creates an instance of FilterService to map and expose data from the bookkeeping service.
+   * @param {BookkeepingService} bookkeepingService - Low level data provider fetching raw data from the BKP source
+   */
+  constructor(bookkeepingService) {
+    assert(bookkeepingService, 'Bookkeeping Service is required');
```
graduta: Given that you have the check in controller, there is no point in keeping it here as assert will throw an error if missing. Remember, filter service will incorporate multiple components and bookkeeping will not be a critical one

QualityControl/test/lib/controllers/FiltersController.test.js
```diff
@@ -0,0 +1,60 @@
...
+  suite('Creating a new FiltersController instance', async () => {
+    test('should successfully initialize FiltersController', async () => {
+      const bkpService = new BookkeepingService({});
+      doesNotThrow(() => new FilterController(bkpService));
+    });
+  });
+
+  suite('getFilterConfigurationHandler', async () => {
+    test('should successfully retrieve run types from Bookkeeping service', async () => {
+      const bkpService = {
```
graduta: I believe this is the filter service now which you can mock with class definition with SInon so that is even more self-explanatory

QualityControl/test/lib/controllers/FiltersController.test.js
```diff
@@ -0,0 +1,60 @@
...
+        status: sinon.stub().returnsThis(),
+        json: sinon.stub(),
+      };
+      const req = {};
+      const filterController = new FilterController(bkpService);
+      await filterController.getFilterConfigurationHandler(req, res);
+      ok(res.status.calledWith(200), 'Response status was not 200');
+      ok(res.json.calledWith({ runTypes: bkpService.runTypes }), 'Run types were not sent back');
+    });
+    test('should return an empty array if bookkeeping service is not defined', async () => {
```
graduta, line 48: Could you add a test in case BKP is configured by replies with error?

QualityControl/test/public/pages/layout-show.test.js
```diff
@@ -34,6 +34,62 @@ export const layoutShowTests = async (url, page, timeout = 5000, testParent) =>
     },
   );
 
+  // await testParent.test(
```
graduta: Will this be enabled? If not, it should be removed

QualityControl/public/common/filters/filterTypes.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+const FILTER_TYPE = {
```
graduta: I believe the convention we are aiming for is:
```
const FilterType = {
  INPUT: 'input',
  DROPDOWN: 'dropdownSelector',
};
```

QualityControl/public/common/filters/filter.js
```diff
@@ -0,0 +1,134 @@
...
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { FILTER_TYPE } from './filterTypes.js';
+import { h } from '/js/src/index.js';
+import { RemoteData } from '/js/src/index.js';
```
graduta: h and RemoteData can be imported together

QualityControl/public/layout/Layout.js
```diff
@@ -801,4 +804,39 @@ export default class Layout extends Observable {
...
+  };
+
+  /**
+   * Method to allow the addition/update/removal of key;value pairs in filter object
+   * @param {string} key - key to look for in filter object
+   * @param {any} value - value to update for given key; if none, entry is removed from object
+   * @returns {undefined}
+   */
+  setFilterValue = (key, value) => {
+    if (value) {
```
graduta: If value is a number and contains value `0` this will not pass and delete the filter. You should specifically check for undefined/null. In this way with values of 0 or empty strings will pass

QualityControl/lib/services/BookkeepingService.js
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { httpGetJson } from '../utils/utils.js';
+import { LogManager } from '@aliceo2/web-ui';
+
+const logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'bkp'}/service`);
```
graduta: For loggers, we try to add a label to specify which services it is. Thus instead of `/service`, we would use `/bkp-service`

QualityControl/public/layout/view/panels/filtersConfig.js
```diff
@@ -0,0 +1,50 @@
...
+    type: FILTER_TYPE.INPUT,
+    queryLabel: 'RunNumber',
+    placeholder: 'RunNumber (e.g. 546783)',
+    id: 'runNumberLayoutFilter',
+    inputType: 'number',
+  },
+  {
+    type: FILTER_TYPE.DROPDOWN,
+    queryLabel: 'RunType',
+    placeholder: 'RunType (e.g. PHYSICS)',
```
graduta: I think given that is a dropdown, we should use `ANY` as placeholder.
Placeholders on dropdown elements are not displayed like on the input ones with a light grey but instead it looks like selected option.

QualityControl/public/Model.js
```diff
@@ -87,6 +88,8 @@ export default class Model extends Observable {
     this.ws.addListener('close', this.handleWSClose.bind(this));
 
     this.initModel();
+
+    this.services.filter.initFilterService();
```
graduta: I believe the services are currently initiliazed in handleLocationChange. It would be good if you initialize that one there as well https://github.com/AliceO2Group/WebUi/pull/2691/files#diff-8829298f0249abbf3900a4d93c5d0d84fc5bb4e38bfcd19be06a9d0b42657522R177

QualityControl/public/layout/view/panels/filters.js
```diff
@@ -12,71 +12,70 @@
...
+      onInputCallback,
+      onEnterCallback,
+    };
+
+    switch (type) {
+      case FILTER_TYPE.INPUT:
+        filterElement = filterInput({ ...commonConfig, type: inputType });
+        break;
+      case FILTER_TYPE.DROPDOWN:
+        filterElement = dynamicSelector({ ...commonConfig, options, onChangeCallback }); break;
```
graduta: Missing new line before `break`

graduta (changes requested):
A few minor comments but after that looks good

QualityControl/public/common/filters/filter.js
```diff
@@ -0,0 +1,134 @@
...
+ * @param {Function} config.onInputCallback - Callback to be triggered on the input event.
+ * @param {Function} config.onEnterCallback - Callback to be triggered when the Enter key is pressed.
+ * @param {string} [config.type='text'] - The type of the filter element (e.g., 'text', 'number').
+ * @param {string} [config.width='w-20'] - The CSS class that defines the width of the filter.
+ * @returns {vnode} - A virtual node element representing the filter input.
+ */
+const filterInput = (config) => {
+  const { queryLabel, placeholder, id, filterMap, onInputCallback, onEnterCallback, type, width = 'w-20' } = config;
+
+  return h(`${width}`, [
```
graduta, line 81: You are missing the `.` in front of the class. 
Like you wrote it, it will create an HTML element w-20 
```html
<w-20></w-20>
```
(the same goes for the dropdown selector

QualityControl/test/lib/controllers/FiltersController.test.js
mariscalromeroalejandro: this test case is checked by filter service. The controller should not be aware of this

graduta (approved):

PR #2702, opened 2024-12-20
[OGUI-1593] Preserve layout tab in path in Layout Show View After Navigating Back from Object View

mariscalromeroalejandro:
- [x] PR labels are selected

* Adds tab information of an specific object id from backend
* Updates 'back to layout' button to add tab to path

### DIFF
--- a/QualityControl/lib/services/JsonFileService.js
+++ b/QualityControl/lib/services/JsonFileService.js
@@ -217,7 +217,7 @@ export class JsonFileService {
       for (const tab of layout.tabs) {
         for (const object of tab.objects) {
           if (object.id === id) {
-            return { object, layoutName: layout.name };
+            return { object, layoutName: layout.name, tab: tab.name };
           }
         }
       }
--- a/QualityControl/lib/services/JsonFileService.js
+++ b/QualityControl/lib/services/JsonFileService.js
@@ -217,7 +217,7 @@ export class JsonFileService {
       for (const tab of layout.tabs) {
         for (const object of tab.objects) {
           if (object.id === id) {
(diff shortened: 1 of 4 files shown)


QualityControl/lib/services/JsonFileService.js
```diff
@@ -217,7 +217,7 @@ export class JsonFileService {
       for (const tab of layout.tabs) {
         for (const object of tab.objects) {
           if (object.id === id) {
-            return { object, layoutName: layout.name };
+            return { object, layoutName: layout.name, tab: tab.name };
```
graduta: I would recommend you name the variable a bit more specific similar to the others. `tab` would refer to the `tab object` itself. In your case (similar to layoutName), you only pass the `tab.name`. Thus, a better naming would be `tabName`

QualityControl/public/pages/objectView/components/header.js
```diff
@@ -87,3 +87,13 @@ function getCopyURLToClipboardButton(model) {
...
   ));
 }
+
+/**
+ * Extract tab name from the selected object
+ * @param {RemoteData} selected - contains the selected object
+ * @returns {string} - name of the tab where the object is placed or empty string if not applicable
+ */
+function getTabFromObject(selected) {
+  const tab = selected?._payload?.tab ?? '';
```
graduta: `_payload` is the private attribute of the RemoteData object `selected`. You should call it with its native getter `payload`

graduta (changes requested):
Good job! Please have a look at recommendations I added

graduta (approved):

PR #2713, opened 2025-01-08
[OGUI-1594] Remove double conversion of grpc error and adapt to usage of errors from web-ui

graduta:
- [x] PR labels are selected

PR which:
* removes CustomError Classes from Control codebase and reuses them from web-ui
* fixes a bug in which EnvironmentService was trying to convert an error from a gRPC error


Control/lib/controllers/Environment.controller.js
pepijndik, line 1: Update the test so that the codeCOv check passes

mariscalromeroalejandro (approved):

PR #2725, opened 2025-01-21
[OGUI-1565] Display infologger link to each task in env-details

pepijndik:
The request of this ticket is to add another button (and such create a btn-group out of the two buttons). This new button should be labeled "ILG" and work in a similar fashion to existing infologgerbutton which opens infologger. But extra, it should add one more parameter, that is pid with the value of the pid of the task that it is in line with. (PID can be seen in ECS GUI table as well)

- [x] PR labels are selected


pepijndik (commented):
Looks good, no comments ;-)

graduta (approved):

PR #2734, opened 2025-01-29
[OGUI-1604] Add middleware to 'edit layout' requests

mariscalromeroalejandro:
- [x] PR labels are selected

This PR adds checks when editing layouts using `PATCH` and `PUT` requests:
* Makes sure that Json File Service is provided.
* Checks that the layout ID is included and actually exists in the system.
* For `PUT` requests, ensures that only the owner can update their layout.
* Verifies that the request contains a body.


QualityControl/lib/middleware/requestBody.middleware.js
graduta, line 1: I would say there is no need for this middleware. If body is not provided the controller check of LayoutDTO will throw an error with a descriptive enough message about it

graduta (changes requested):
Hi @mariscalromeroalejandro ,
The PR is in the right direction, good job!
Given that we move these checks to middleware function, and cleaning the tests for controller on such criteria, it means we are left exposed on API being tested against these scenarios.

Please have a look at the API tests from Control https://github.com/AliceO2Group/WebUi/tree/dev/Control/test/api and add some API tests for the PUT and PATCH `layout:id` paths as well

QualityControl/lib/middleware/requestBody.middleware.js
mariscalromeroalejandro: the function `LayoutDto.validateAsync(req.body)` is supposed to throw an error if req.body is missing from the request but it is not failing as expected

QualityControl/test/api/layouts/api-put-layout.test.js
```diff
@@ -0,0 +1,48 @@
...
+        .put(`?token=${ADMIN_TEST_TOKEN}`)
+        .expect(404, {
+          message: 'layout (test) not found',
+          status: 404,
+          title: 'Not Found',
+        });
+    });
+
+    test('should return a 404 error if the layout id is not provided', async () => {
+      await request('localhost:8080/api/layout/')
```
graduta: Hi @mariscalromeroalejandro ,

I had a look at your test as you have asked me. Please find below my comment:

For your test:
```
 await request('localhost:8080/api/layout/')
        .put(`?id=${null}&token=${ADMIN_TEST_TOKEN}`)
        .expect(404, {
          message: 'layout id not provided',
          status: 404,
          title: 'Not Found',
        });
```
You are building the request like this with `id` passed as `?id`. When passed like this, this is called a `query parameter` and it is taken from `req.query`.
But your api route defined in `api.js` does not expect a query parameter but instead an ID directly in the URL.
```
http.put(
    '/layout/:id',
)
```
This type of parameter is called simply parameter and is taken from `req.params`

So your `null` for the test should be in the URL of the request. That is why for now, it is failing with URL not found, because you are not passing anything to it. 
So you have two options now:
- test that if there is nothing passed, the error message is page not found
- pass `null`/`undefiend` and check accordingly, e.g. below

e.g. 
```
    test('should return a 404 error if the layout id is not provided', async () => {
      await request(`localhost:8080/api/layout/${undefined}`)
        .put(`?token=${ADMIN_TEST_TOKEN}`)
        .expect(404, {
          message: 'layout (undefined) not provided',
          status: 404,
          title: 'Not Found',
        });
    });
```

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -126,47 +124,35 @@ export class LayoutController {
...
+      const layouts = await this._dataService.listLayouts({ name: layoutProposed.name });
+      const layoutExistsWithName = layouts.every((layout) => layout.id !== layoutProposed.id);
+      if (layouts.length > 0 && layoutExistsWithName) {
+        updateAndSendExpressResponseFromNativeError(
+          res,
+          new InvalidInputError(`Proposed layout name: ${layoutProposed.name} already exists`),
+        );
+        return;
+      }
+      const layout = await this._dataService.updateLayout(id, layoutProposed);
```
graduta: This method will return a `layoutId`. Could you please update the name of the variable accordingly?
```suggestion
      const layoutId = await this._dataService.updateLayout(id, layoutProposed);
```

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -229,33 +215,22 @@ export class LayoutController {
...
-      try {
-        const layoutUpdated = await this._dataService.updateLayout(id, layout);
-        res.status(201).json(layoutUpdated);
-      } catch {
-        updateAndSendExpressResponseFromNativeError(res, new Error(`Unable to update layout with id: ${id}`));
-        return;
-      }
+    let layout = {};
+    try {
+      layout = await LayoutPatchDto.validateAsync(req.body);
```
graduta, line 217: Here I see that we do not have a check on missing `req.body` as we do for `putLayoutHandler`.
Do you think it should also be added?

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -229,33 +215,22 @@ export class LayoutController {
...
+    } catch {
+      updateAndSendExpressResponseFromNativeError(
+        res,
+        new InvalidInputError('Invalid request body to update layout'),
```
graduta: Moreover, I see that we do not provide the error message from the `vaidateAsync` like we do for PUT.
Perhaps this can also be updated. In this way we will have a convention

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -229,33 +215,22 @@ export class LayoutController {
...
+      );
+      return;
+    }
+    try {
+      const layoutUpdated = await this._dataService.updateLayout(id, layout);
```
graduta: Similar to the other method, might be good to be more explicit with the naming:
```suggestion
      const layoudId = await this._dataService.updateLayout(id, layout);
```

QualityControl/test/api/layouts/api-put-layout.test.js
```diff
@@ -0,0 +1,48 @@
...
+        .put(`?id=${null}&token=${ADMIN_TEST_TOKEN}`)
+        .expect(404, {
+          message: 'layout id not provided',
+          status: 404,
+          title: 'Not Found',
+        });
+    });
+
+    test.skip('should return a 401 error if the user is not the owner of the layout', async () => {
```
graduta: The following two tests are missing.

graduta (changes requested):
Direction of the PullRequest is good. There are a few comments I have added plus, please add also API tests for PATCH

QualityControl/lib/controllers/LayoutController.js
mariscalromeroalejandro: In the following push, I've also removed the req.body check in putLayoutHandler because the error is now triggered directly in the validateAsync method if no body is provided in the request. Therefore, an additional check isn't necessary.

QualityControl/lib/controllers/LayoutController.js
mariscalromeroalejandro: Done

QualityControl/lib/controllers/LayoutController.js
mariscalromeroalejandro: Done

QualityControl/lib/controllers/LayoutController.js
mariscalromeroalejandro: Done

QualityControl/test/api/layouts/api-patch-layout.test.js
```diff
@@ -0,0 +1,74 @@
...
+    test('should return a 404 error if the id of the layout does not exist', async () => {
+      await request(`${URL_ADDRESS}/api/layout/`)
+        .patch(`?token=${OWNER_TEST_TOKEN}`)
+        .expect(404, {
+          error: '404 - Page not found',
+          message: 'The requested URL was not found on this server.',
+        });
+    });
+
+    test('should return a 404 error if the id is not provided', async () => {
```
graduta: I believe you miss-copied the description of the test.
You are saying in this test `should return a 404 error if the id is not provided` but then in the test, the id is provided `test` 

Then in the test above you are saying `should return a 404 error if the id of the layout does not exist` but you are not providing any id to check

Probably you swapped by mistake between the two

graduta (changes requested):

QualityControl/test/api/layouts/api-patch-layout.test.js
mariscalromeroalejandro: Sure, it was a mistake. Done.

graduta (dismissed):

QualityControl/test/api/layouts/api-put-layout.test.js
```diff
@@ -0,0 +1,81 @@
...
+        .expect(404, {
+          message: 'layout (test) not found',
+          status: 404,
+          title: 'Not Found',
+        });
+    });
+
+    test('should return a 404 error if the layout id is not provided', async () => {
+      await request(`${URL_ADDRESS}/api/layout/`)
+        .put(`?id=${null}&token=${OWNER_TEST_TOKEN}`)
```
graduta: Where is the query parameter `id` used? As far as I can tell the id/name of the layout is taken from the req url param list

graduta (changes requested):

QualityControl/test/api/layouts/api-put-layout.test.js
mariscalromeroalejandro: Yes. The id and token are included in the URL as query parameters, not in the request body. They are still part of the query string in the URL

graduta: But the API PUT route is not expecting a query parameter (the ones after `?`)
It is expecting a URL parameters (the one after `api/layout/`

PR #2743, opened 2025-02-07
[OGUI-1054] Status routes convention

mariscalromeroalejandro:
- [X] PR labels are selected
- [X] FLP integration tests were ran successful

This PR:
- Removes Consul information from requests to /status.
- Modifies the /status/framework route to /status/:component, allowing status queries for specific components.
- Updates the frontend of the about page to support the new routes for fetching the service status.


QualityControl/lib/config/publicConfigProvider.js
```diff
@@ -38,6 +38,8 @@ function buildPublicConfig(config) {
     REFRESH_MIN_INTERVAL: config?.consul?.refreshRate?.min || 10,
     REFRESH_MAX_INTERVAL: config?.consul?.refreshRate?.max || 120,
     CONSUL_SERVICE: config.consul ? true : false,
+    ABOUT_COMPONENTS: ['qcg', 'qc', 'ccdb'],
```
graduta: Given that we have an enum for this, would it not be better to use that instead of hardcoding the values here?

QualityControl/lib/api.js
```diff
@@ -58,7 +59,12 @@ export const setup = (http, ws) => {
   );
 
   http.get('/status/gui', statusController.getQCGStatus.bind(statusController), { public: true });
-  http.get('/status/framework', statusController.getFrameworkInfo.bind(statusController), { public: true });
+  http.get(
+    '/status/:component',
+    statusComponentMiddleware,
+    statusController.getFrameworkInfo.bind(statusController),
```
graduta: I believe the method name here has two issues:
- name is not representative, you are retrieving the status of a component which is defined in the URL
- for controllers, let's try to keep the convention of also ending in `handler` as name

QualityControl/lib/middleware/status/statusComponent.middleware.js
```diff
@@ -0,0 +1,37 @@
...
+ * Returns 400 if the component is missing or invalid, otherwise calls next()
+ * @param {Express.Request} req - HTTP Request
+ * @param {Express.Response} res - HTTP Response
+ * @param {Express.Next} next - HTTP Next (check pass)
+ * @returns {void}
+ */
+export const statusComponentMiddleware = (req, res, next) => {
+  try {
+    if (!req.params.component) {
+      throw new InvalidInputError('Component parameter is missing');
```
graduta, line 30: If there is a missing `component` parameter, the URL will not be matched on this API route. 
Thus, this if will never be triggered and it is unused code

QualityControl/public/frameworkInfo/FrameworkInfo.js
```diff
@@ -31,19 +31,28 @@ export default class FrameworkInfo extends Observable {
...
+  async getFrameworkInfo(components) {
     this.item = RemoteData.loading();
     this.notify();
 
-    const { result, ok } = await this.model.loader.get('/api/status/framework');
-    if (!ok) {
-      this.item = RemoteData.failure(result.message);
+    try {
+      const results = {};
+      await Promise.all(components.map(async (component) => {
```
graduta: Do I understand correctly that these are just migration changes and the changes for front-end will come in a future PR?
Because in these changes, the status cards will still be loaded all at one time compared to the requirement of the ticket to load independently

graduta (changes requested):
I think `components` is not a descriptive enough name, perhaps `integrated services` or something like this might be better. What do you think?

QualityControl/lib/config/publicConfigProvider.js
mariscalromeroalejandro: You’re right! I have a bad memory, and I had forgotten that I had already created an enum for the integrated services

QualityControl/lib/middleware/status/statusComponent.middleware.js
mariscalromeroalejandro: done

graduta (approved):

PR #2744, opened 2025-02-10
[OGUI-1612] Keep grpc call in workflowtemplate.service independently and pass informative message to user

graduta:
- [x] PR labels are selected

PR which:
* updates front-end so that informative error message in case of failure to retrieve workflow information is displayed to the user rather than a hardcoded one
* updates the backend service of WorkflowTemplate to keep the gRPC call independently from other potential errors

### DIFF
--- a/Control/lib/services/WorkflowTemplate.service.js
+++ b/Control/lib/services/WorkflowTemplate.service.js
@@ -38,8 +38,18 @@ class WorkflowTemplateService {
    * @throws
    */
   async getDefaultTemplateSource() {
-    const {repos: repositories} = await this._coreGrpc['ListRepos']();
-    const defaultRepository = repositories.find((repository) => repository.default);
+    /**
+     * @type {grpc.core.ListReposReply}
+     */
+    let repositoriesReply = {};
+    try {
+      repositoriesReply = await this._coreGrpc['ListRepos']();
+      
+    } catch (error) {
+      throw grpcErrorToNativeError(error);
+    }
+    const { repos = [] } = repositoriesReply;
+    const defaultRepository = repos.find((repository) => repository.default);
 
     if (!defaultRepository) {
       throw new NotFoundError(`Unable to find a default repository`);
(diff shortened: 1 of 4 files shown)


Control/lib/services/WorkflowTemplate.service.js
```diff
@@ -59,34 +69,36 @@ class WorkflowTemplateService {
...
-        return mappings.sort(({label: labelA}, {label: labelB}) => labelA < labelB ? -1 : 1);
-      }
-
-      return [];
+      mappingsString = await this._apricotGrpc.getRuntimeEntryByComponent(RUNTIME_COMPONENT, RUNTIME_KEY);
     } catch (error) {
       throw grpcErrorToNativeError(error);
     }
+    const mappings = JSON.parse(mappingsString);
+    if (Array.isArray(mappings)) {
```
martinboulais: I personally prefer to explicitely catch specific cases in ifs then continue the process, it avoids nested ifs if you need to add more cases (`if (!Array.isArray(mappings)) {return []}`)

Control/lib/services/WorkflowTemplate.service.js
```diff
@@ -59,34 +69,36 @@ class WorkflowTemplateService {
...
+      return mappings.sort(({ label: labelA }, { label: labelB }) => labelA < labelB ? -1 : 1);
```
martinboulais: You might prefer to use `localeCompare`?

martinboulais (changes requested):

Control/lib/services/WorkflowTemplate.service.js
graduta: Good catch

Control/lib/services/WorkflowTemplate.service.js
graduta: You are right, long-term it would make the code easier to read

martinboulais (approved):

PR #2754, opened 2025-02-17
[OGUI-1567] ilg link to bkp

graduta:
- [x] PR labels are selected

PR which:
* is raised on behalf of @Houwie7000 
* adds a new endpoint to request the configuration of infologger external services
* adds a button which allows users to get redirected to BKP on-call entry form with available log data pre-filled


InfoLogger/public/Model.js
```diff
@@ -38,6 +39,9 @@ export default class Model extends Observable {
 
     this.loader = new Loader(this);
     this.loader.bubbleTo(this);
+    this.configurationService = new ConfigurationService(this);
+    this.configurationService.bubbleTo(this);
+    this.configurationService.load();
```
mariscalromeroalejandro, line 44: load is async, shouldn't it wait until is completed?

InfoLogger/public/services/externalRouting/getBkpLogEntryFormUrl.js
```diff
@@ -0,0 +1,56 @@
...
+
+  let bkpPreparedUrl = configuration.payload?.bookkeeping.url;
+  if (!bkpPreparedUrl) {
+    return '';
+  }
+
+  const bkpUrlParameters = {
+    page: BKP_PAGE,
+    templateKey: BKP_TEMPLATE_KEY,
+    ...logItem.run !== null && { runNumbers: [logItem.run] },
```
mariscalromeroalejandro: Let say that `run` is set to `undefined` or `logItem` does not have the `run` property, in this case `longItem.run !== null` will be true and `runNumbers` will be set to `undefined` in both cases, is this the behavior we want?

Instead of checking only for null, it can be good to check that run is neither nulll nor undefined before including it in runNumbers: 
`...(logItem.run != null ? { runNumbers: [logItem.run] } : {})
` using != for loose comparison

InfoLogger/test/lib/controller/mocha-config-controller.js
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  describe('getConfigurationHandler', () => {
+    it('should successfully return configuration', async () => {
+      const controller = new ConfigController(configMock);
+      await controller.getConfigurationHandler(null, resMock);
+
+      assert.ok(resMock.status.calledWith(200));
+      assert.ok(resMock.json.calledWith({
+        bookkeeping: { url: 'http://bookkeeping.com' },
```
mariscalromeroalejandro: hardcoded URL, variable instead?

mariscalromeroalejandro (changes requested):

InfoLogger/public/Model.js
graduta: It is not recommended to wait here, due to the fact that this is the model constructor initializing and in case for some reason the endpoint is not responding or timing out, it means the whole application will be stuck and not loaded due to this

InfoLogger/test/lib/controller/mocha-config-controller.js
graduta: Updated

InfoLogger/public/services/externalRouting/getBkpLogEntryFormUrl.js
graduta: Yes, you are correct, I should simply compare if it exists like with the other fields given that runNumber will never be `0` case in which `...logItem.run` would return false

PR #2766, opened 2025-02-27
[OGUI-1162] Add support for env ecs topic

graduta:
- [x] PR labels are selected

PR which:
* adds support for processing events on the environment topic
* moves to using the web-ui framework provided consumer
* extracts topics and consumer groups `const`(s) in `enum` objects
* updates kafka to retry on failure to connect


Control/lib/kafka/adapters/environmentEventAdapter.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {Event.proto} eventMessage - the event message to adapt
+ * @param {int64.proto} eventMessage.timestamp - the timestamp of the event
+ * @param {Ev_EnvironmentEvent.proto} eventMessage.environmentEvent - the environment event to adapt
+ * @return {EnvironmentEvent} - the adapted event message without the timestampNano field
+ */
+exports.environmentEventAdapter = ({ timestamp, environmentEvent }) => {
+  const {
+    environmentId: id,
+    state, runNumber, error, message, transition, transitionStep, transitionStatus, vars, lastRequestUser
+  } = environmentEvent;
```
mariscalromeroalejandro, line 25: If any of the properties are missing, the code can throw an Error. Maybe you can give them default values

Control/lib/kafka/enums/topics.enum.js
```diff
@@ -0,0 +1,19 @@
...
+ *  Version 3), copied verbatim in the file "COPYING".
+ *
+ *  See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ *  In applying this license CERN does not waive the privileges and immunities
+ *  granted to it by virtue of its status as an Intergovernmental Organization
+ *  or submit itself to any jurisdiction.
+ */
+
+exports.Topics = Object.freeze({
```
mariscalromeroalejandro, line 18: documentation is missing

Control/lib/kafka/enums/topics.enum.js
```diff
@@ -0,0 +1,19 @@
...
+  ENVIRONMENT: ['aliecs.environment'],
+  INTEGRATED_SERVICE: {
```
mariscalromeroalejandro, line 20: `ENVIRONMENT` is an array, while `INTEGRATED_SERVICES` is an object, maybe in terms of consistency it can be changed

Control/lib/kafka/enums/consumerGroups.enum.js
```diff
@@ -0,0 +1,19 @@
...
+ *  Version 3), copied verbatim in the file "COPYING".
+ *
+ *  See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ *  In applying this license CERN does not waive the privileges and immunities
+ *  granted to it by virtue of its status as an Intergovernmental Organization
+ *  or submit itself to any jurisdiction.
+ */
+
+exports.ConsumerGroups = Object.freeze({
```
mariscalromeroalejandro, line 18: documentation is missing

mariscalromeroalejandro (changes requested):

Control/lib/kafka/adapters/environmentEventAdapter.js
graduta: The code will not throw an error but instead the variable will have the value `undefined`. This is the normal behaviour of JSON destructuring: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment 
In our case, we also want to be able to tell when a value is missing so that we can report to ECS.

Control/lib/kafka/enums/topics.enum.js
graduta: Indeed, fixed!

Control/lib/kafka/enums/topics.enum.js
graduta: It cannot be. The integrated service, while for now a single value will contain multiple sub-topics which in turn will contain arrays.
An array is needed when passing the topics to follow to KafkaConsumer

Control/lib/kafka/enums/consumerGroups.enum.js
graduta: Indeed, fixed!

Control/lib/kafka/adapters/environmentEventAdapter.js
mariscalromeroalejandro: Understood. I think a good practice could be to tell ECS explicitly that a value is missing instead of leaving it undefined, because one thing is to not provide the value and another thing is provide it as 'undefined' 

`state='MISSING'`, for example

mariscalromeroalejandro (approved):

Control/lib/kafka/adapters/environmentEventAdapter.js
graduta: We do not tell to ECS directly that a value is undefined. But the dataflow is:
ECS -> Kafka -> GUI
That is, in the GUI we can immediately tell from logging and display that the value is undefined. The GUI on client side will take care of this.
Moreover, it would not be good practice for the GUI to add `MISSING` as state, as this does not exist in the ECS state machine and it will create confusion for the shifters

PR #2768, opened 2025-02-27
[OGUI-1162] Add support for task ecs topics

graduta:
- [x] PR labels are selected

PR which:
* adds support for processing event messages on task topic
* adds TaskEvent adapter
* adds TaskEvent definition
* adds back declaration of `async` as this is mandatory due to the fact that the KafkaConsumer calls `catch` on it as it expects a promise
* adds check for timestamp before attempting to convert in all events adapters


Control/lib/kafka/adapters/taskEventAdapter.js
```diff
@@ -0,0 +1,45 @@
...
+    state,
+    status,
+    hostname,
+    className,
+    traits,
+    environmentId,
+    path
+  } = taskEvent;
+  return {
+    timestamp: timestamp.toNumber(),
```
mariscalromeroalejandro: check the timestamp before? it will break if it´s null

mariscalromeroalejandro (changes requested):

Control/lib/kafka/adapters/taskEventAdapter.js
graduta: Good catch, updated

mariscalromeroalejandro (approved):

PR #2780, opened 2025-03-07
[OGUI-1619] Move CRUD logic of layouts to repository

mariscalromeroalejandro:
- [x] PR labels are selected

This PR moves the CRUD logic related to layouts into a dedicated repository:
- Creates Chart Repository, Layout Repository and User Repository
- Now controllers use these repositories instead of JSONFileService
- UserService transitions to UserController
- Adds tests for new repositories, for user controller and for JSONFileService


QualityControl/lib/database/repositories/LayoutRepository.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * LayoutRepository class to handle CRUD operations for Layouts.
+ */
+export default class LayoutRepository {
```
martinboulais: I usually avoid every default exports (makes the code much harder to maintain, because you can't rely on the class name which might (and probably will) be changed when importing).
You can do a quick google search to have more detailed discussion about this topic.

QualityControl/lib/database/repositories/LayoutRepository.js
```diff
@@ -0,0 +1,44 @@
...
+  findLayoutById(data, layoutId) {
```
martinboulais: Two points here: 
1- You should add docblock with at least the type of parameter expected. In the repositories you can expect the proper data to be provided (like if you expect an int, you do not have to check if you have a proper int here and not something like `hello world`, that is the job of controllers to check user request) but you need to hint developer about what is the data expected.
2- The interface should not change when you will move to sequelize (that's the purpose). Here if you expect the data, you directly make the repository depend on the underlying database system => See my comments on JsonFileService file

QualityControl/lib/database/repositories/LayoutRepository.js
```diff
@@ -0,0 +1,44 @@
...
+    return data.layouts.find((layout) => layout.id === layoutId);
```
martinboulais: My personal preference is to use `null` when meaning `no data`, not undefined (as it is the case here). 
That is my personal opinion, the best is to stick to the project's conventions.

QualityControl/lib/database/repositories/LayoutRepository.js
```diff
@@ -0,0 +1,44 @@
...
+
+  findLayoutByName(data, layoutName) {
+    return data.layouts.find((layout) => layout.name === layoutName);
+  }
+
+  createLayout(data, newLayout) {
+    data.layouts.push(newLayout);
+  }
+
+  updateLayout(layout, newData) {
```
martinboulais: In this case I would suggest to have a function `updateLayoutById`, taking as parameter the id of the layout and the new data (I personally name this kind of variable "patch").
The reason is here is that you expect the user to actually have fetched the layout somehow, and because it is based on reference you might override key properties, such as the layout ID (which I suppose should not be allowed)

QualityControl/lib/database/repositories/LayoutRepository.js
```diff
@@ -0,0 +1,44 @@
...
+    Object.assign(layout, newData);
+  }
+
+  deleteLayout(data, layout) {
+    const index = data.layouts.indexOf(layout);
```
martinboulais: Do you know what happens if trying to delete a layout that do not exist?
Here too, I would rather name the function `deleteLayoutById` and not rely on a reference.

martinboulais (changes requested):

QualityControl/lib/database/repositories/LayoutRepository.js
mariscalromeroalejandro: I used the default export because in this case, I would only be exporting individual repositories

QualityControl/lib/services/JsonFileService.js
```diff
@@ -43,8 +40,8 @@ export class JsonFileService {
    */
   async _syncFileAndInternalState() {
     await this._readFromFile();
-    await this._writeToFile();
-    logger.info(`Preferences will be saved in ${this.pathname}`);
+    this.writeToFile();
```
martinboulais: Why removing the await here?

QualityControl/lib/database/repositories/LayoutRepository.js
martinboulais: Yes I know, but even in this case default export is a bad idea.
In calling code you have no guarantee that devs will not name it `HelloRepo` and then it's a pain to look for reference tree.

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -0,0 +1,129 @@
...
+   * @returns {Array<Layout>} - list of layouts as per the filter
+   */
+  listLayouts(filter) {
+    return this._jsonFileService.data.layouts.filter((layout) =>
+      (filter.owner_id === undefined || layout.owner_id === filter.owner_id)
+            && (filter.name === undefined || layout.name === filter.name));
+  }
+
+  /**
+   * Retrieve a layout or undefined
```
martinboulais: `or undefined` should be `or throws`

QualityControl/test/lib/services/json-service.test.js
martinboulais, line 1: I do not know if you intend to push more commits here, but the tests that you removed from here should be added in something like `layout-service.test.js` and `user-service.test.js` (and adapted of course)

QualityControl/test/test-index.js
```diff
@@ -167,6 +163,9 @@ suite('All Tests - QCG', { timeout: FRONT_END_TIMEOUT + BACK_END_TIMEOUT }, asyn
 
     suite('Controllers - Test Suite', async () => {
       suite('LayoutController test suite', async () => await layoutControllerTestSuite());
+      suite('UserController test suite', async () => {
+        // TODO - bring inline with current tests
```
martinboulais: Do you plan to push this in this PR?
If not please remove this, it is dead code and it should be avoided

martinboulais (changes requested):
Very good job!
I have put some minor comments but the right structure is there!

QualityControl/test/lib/services/json-service.test.js
mariscalromeroalejandro: Yes, they will be added in the next commits

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -96,34 +96,54 @@ export default class LayoutRepository {
...
+      for (const tab of layout.tabs) {
+        for (const object of tab.objects) {
+          if (object.id === id) {
+            return { object, layoutName: layout.name, tabName: tab.name };
+          }
+        }
+      }
+    }
+    throw new Error(`Object with ${id} could not be found`);
+  }
```
martinboulais, line 127: This should be in a dedicated repository, I guess "ObjectRepository` or any name we agreed on for the future database

QualityControl/test/lib/repositories/LayoutRepository.test.js
```diff
@@ -0,0 +1,206 @@
...
+        const layoutIdToDelete = '3';
+        const deletedLayoutId = await layoutRepository.deleteLayout(layoutIdToDelete);
+
+        assert.strictEqual(deletedLayoutId, layoutIdToDelete);
+        assert.strictEqual(jsonFileServiceMock.data.layouts.length, 2);
+        assert.deepStrictEqual(
+          jsonFileServiceMock.data.layouts,
+          mockedLayouts,
+        );
+        strictEqual(deletedLayoutId, layoutIdToDelete);
```
martinboulais, line 168: You should try to be consistent in your usage of `assert.something` and `something`, in this case `strictEqual`

QualityControl/test/lib/services/JsonFileService.test.js
```diff
@@ -0,0 +1,102 @@
...
+      );
+    });
+
+
+
+
+
+
+
+  });
```
martinboulais: Formatting to fix

martinboulais (changes requested):

QualityControl/lib/repositories/LayoutRepository.js
martinboulais: Please also move the tests in a dedicated file

martinboulais (approved):

PR #2811, opened 2025-04-10
[OGUI-1630] Handle gRPC client lost connection and retry

graduta:
- [x] PR labels are selected

PR which:
* renames the `GrpcProxy` into `GrpcServiceClient`
* adds first connection attempt and retry if it fails. This is needed in case the main gRPC server is not yet up and running in the farm
* adds monitoring to the channel activity and reconnect attempt if the channel fails - this is needed for when the main server gets updated and we do not want to restart the GUI for it
* Add new library for injecting stubs into file for testing
* adds tests for `GrpcServiceClient`


Control/lib/control-core/GrpcServiceClient.js
```diff
@@ -0,0 +1,297 @@
...
+  
+      const address = `${config.hostname}:${config.port}`;
+      const packageDefinition = protoLoader.loadSync(path, {longs: String, keepCase: true, arrays: true});
+      const octlProto = grpcLibrary.loadPackageDefinition(packageDefinition);
+      const protoService = octlProto[this._package][this._label];
+  
+      const credentials = grpcLibrary.credentials.createInsecure();
+      const options = {'grpc.max_receive_message_length': 1024 * 1024 * this._maxMessageLength}; // MB
+
+      this.client = new protoService(address, credentials, options);
```
mariscalromeroalejandro, line 79: handle error instantiating (try catch) or not existing protoService

martinboulais (dismissed):

Control/lib/control-core/GrpcServiceClient.js
```diff
@@ -0,0 +1,297 @@
...
+    if (!path) {
+      this._logger.errorMessage('Missing path for gRPC API declaration');
+      isValid = false;
+    }
+    if (!config.label) {
+      this._logger.errorMessage('Missing service label for gRPC API');
+      isValid = false;
+    }
+    if (!config.package) {
+      this._logger.errorMessage('Missing service label for gRPC API');
```
mariscalromeroalejandro: Missing package, not service label

Control/lib/control-core/GrpcServiceClient.js
```diff
@@ -0,0 +1,297 @@
...
+    };
+    return methodName;
+  }
+
+  /**
+   * Checks if configuration provided for gRPC Connection is valid
+   * @param {JSON} config
+   * @param {String} path - location of gRPC file containing API
+   */
+  _isConfigurationValid(config, path) {
```
mariscalromeroalejandro, line 135: what if config is null?

Control/lib/control-core/GrpcServiceClient.js
```diff
@@ -0,0 +1,297 @@
...
+          && currentState !== grpcLibrary.connectivityState.CONNECTING
+        ) {
+          this._handleConnectionStatus(
+            new ServiceUnavailableError(`Channel state changed to ${grpcLibrary.connectivityState[currentState]}`),
+            channel.getTarget()
+          );
+          this._establishConnectionWithRetry(channel.getTarget());
+          return;
+        }
+        channel.watchConnectivityState(currentState, Date.now() + 10000, checkState);
```
mariscalromeroalejandro: parametrize timeout instead of a hardcoded value (1000)

mariscalromeroalejandro (changes requested):

Control/lib/control-core/GrpcServiceClient.js
graduta: If config is null, we will never reach this line. It is not an optional but a mandatory parameter.
The server will not start at all without a gRPC config: https://github.com/AliceO2Group/WebUi/blob/dev/Control/lib/api.js#L70

Control/lib/control-core/GrpcServiceClient.js
graduta: Could you please explain what you are referring to?

mariscalromeroalejandro: if there is any error trying to instantiate the service, it is not caught

mariscalromeroalejandro (approved):

Control/lib/control-core/GrpcServiceClient.js
graduta: As far as I know, it is not good practice for constructors to do that and instead is better to let the error propagate to the caller

PR #2814, opened 2025-04-11
[OGUI-1592] Add warning messages if some CRUs are missing configuration

graduta:
- [x] PR labels are selected

PR which:
* adds warning messages at detector panel and host panel levels if one of their CRUs is missing configuration. 
* there was already a CRU level warning message but this is not clearly seen
* Screenshots were added to the jira ticket


Control/public/common/warningComponent.js
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+import {h, iconMedicalCross} from '/js/src/index.js';
+
+/**
+ * A component which displays a medical cross icon and a warning message given message
+ * @param {string} warningMessage - The message to be displayed
+ * @returns {vnode}
+ */
+export const warningMessageOnMissingCruConfig = (warningMessage = '', classOptions = []) => {
```
martinboulais: The name can be more generic, I think that the file name is good

martinboulais: classOptions missing docblock. Why is it classOptions if it contains only classes string? I would rather name it `classes`

Control/public/configuration/utils/isValidCruConfig.js
```diff
@@ -0,0 +1,23 @@
...
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+/**
+ * Check if the configuration given host has a valid configuration for each of its CRUs
+ * Validation is done by checking if the configuration contains at least one link
+ * @param {object<string, <string, {config: object, info: object}>} cruDataOfHost - Map of CRUs by host
+ */
+export const isValidCruConfig = (cruDataOfHost) =>
```
martinboulais: If this function is used only in one file, I would be in favor of making it private in that module, not exported from its own file

Control/public/app.css
```diff
@@ -116,3 +116,4 @@
 .grid-container { display: grid; gap: 1  rem; grid-template-columns: repeat(auto-fill, minmax(150px, 1fr)); }
 .bg-yellow { background-color: #e6e200; }
 .italic { font-style: italic; }
+.left-align { display: flex; margin-left: auto; }
```
martinboulais: I think that `display: flex` should not be part of a class named `left-align`, and I am pretty sure you do not need it there

martinboulais (changes requested):

Control/public/common/warningComponent.js
graduta: Indeed, initially I had it in the same module and exported after. Fixed

graduta: Updated

Control/public/configuration/utils/isValidCruConfig.js
graduta: I was thinking that it might be needed for a refactor on the CRUs model, but until then I will move it to the module

Control/public/app.css
graduta: Indeed, thank you!

martinboulais (approved):

PR #2818, opened 2025-04-11
[OGUI-1640] Handle error when ccdb version is unable to be retrieved

mariscalromeroalejandro:
- [x] PR labels are selected

This PR fixes an issue where the ccdb service would throw an error if it couldn’t provide the requested version:
* Now, it returns "Unknown version" instead of an error
* Tests have been updated to check for this behavior.

### DIFF
--- a/QualityControl/lib/services/ccdb/CcdbService.js
+++ b/QualityControl/lib/services/ccdb/CcdbService.js
@@ -89,10 +89,8 @@ export class CcdbService {
       throw new Error(`Unable to connect to CCDB due to: ${error}`);
     }
     try {
-      const monitorData = serviceInfo[CCDB_MONITOR] && serviceInfo[CCDB_MONITOR][this._hostname];
-      const version = Array.isArray(monitorData) && monitorData[0]?.value
-        ? monitorData[0].value
-        : 'unknown version';
+      const monitorData = serviceInfo?.[CCDB_MONITOR]?.[this._hostname] ?? [];
+      const version = monitorData[0]?.value ?? 'unknown version';
       return { version };
     } catch (error) {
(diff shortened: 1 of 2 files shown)


QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -89,7 +89,10 @@ export class CcdbService {
       throw new Error(`Unable to connect to CCDB due to: ${error}`);
     }
     try {
-      const version = Object.values(serviceInfo[CCDB_MONITOR])[0][0]?.value ?? '-';
+      const monitorData = serviceInfo[CCDB_MONITOR] && serviceInfo[CCDB_MONITOR][this._hostname];
```
graduta, line 92: I think these 2 lines are not easy to read and can be improved. What do you think of:
```suggestion
      const monitorData = serviceInfo?.[CCDB_MONITOR]?.[this._hostname] ?? [];
      const version = monitorData[0]?.value ?? 'unknown version';
```

QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -89,7 +89,10 @@ export class CcdbService {
...
+      const version = Array.isArray(monitorData) && monitorData[0]?.value
```
graduta: I think these 2 lines are not easy to read and can be improved. What do you think of:
```suggestion
      const monitorData = serviceInfo?.[CCDB_MONITOR]?.[this._hostname] ?? [];
      const version = monitorData[0]?.value ?? 'unknown version';
```

graduta (changes requested):

graduta (approved):

PR #2832, opened 2025-04-24
[OGUI-1619] Database creation and sequelize migrations

mariscalromeroalejandro:
- [x] PR labels are selected

This PR sets up the necessary environment to create the database and run the migrations to create the required tables. It also includes the creation of models and their relationships.

### DIFF
--- a/QualityControl/lib/config/database.js
+++ b/QualityControl/lib/config/database.js
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+ * See http://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file 'COPYING'.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * Returns a normalized database configuration object.
+ * @param {object} config - Custom config defined in the config file.
+ * @returns {object} Normalized database configuration.
+ */
+export function getDbConfig(config) {
+  return {
+    host: process?.env.DATABASE_HOST || 'localhost',
+    port: 3306,
+    username: 'cern',
+    password: 'cern',
+    database: 'qcg',
+    charset: 'utf8mb4',
+    collate: 'utf8mb4_general_ci',
+    timezone: '+00:00',
+    logging: false,
+    retryThrottle: 5000,
+    ...config,
+  };
+};
--- a/QualityControl/lib/config/database.js
+++ b/QualityControl/lib/config/database.js
@@ -19,16 +19,15 @@
  */
 export function getDbConfig(config) {
   return {
-    host: process?.env.DATABASE_HOST || 'localhost',
-    port: 3306,
-    username: 'cern',
-    password: 'cern',
-    database: 'qcg',
-    charset: 'utf8mb4',
-    collate: 'utf8mb4_general_ci',
-    timezone: '+00:00',
-    logging: false,
-    retryThrottle: 5000,
-    ...config,
+    host: config.host || 'localhost',
+    port: config.port || 3306,
+    username: config.username || 'cern',
+    password: config.password || 'cern',
+    database: config.database || 'qcg',
+    charset: config.charset || 'utf8mb4',
+    collate: config.collate || 'utf8mb4_general_ci',
+    timezone: config.timezone || '+00:00',
+    logging: config.logging ?? false,
+    retryThrottle: config.retryThrottle || 5000,
   };
 };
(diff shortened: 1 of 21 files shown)


QualityControl/lib/config/database.js
```diff
@@ -0,0 +1,34 @@
...
+    port: 3306,
+    username: 'cern',
+    password: 'cern',
+    database: 'qcg',
+    charset: 'utf8mb4',
+    collate: 'utf8mb4_general_ci',
+    timezone: '+00:00',
+    logging: false,
+    retryThrottle: 5000,
+    ...config,
```
graduta: I think here it would be better if we use the format for each field of:
```
username: config.username ?? 'cern`
```
So that we are explicit to what we bring from the configuration file. Otherwise, unknown attributes may be passed through the use of `...config`

QualityControl/lib/database/SequelizeDatabase.js
```diff
@@ -0,0 +1,131 @@
...
+
+import { LogManager } from '@aliceo2/web-ui';
+import { Sequelize } from 'sequelize';
+import pkg from 'umzug';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+import { createUmzug } from './umzug.js';
+import { getDbConfig } from '../config/database.js';
+import models from './models/index.js';
+const { SequelizeStorage } = pkg;
```
graduta: Given that we only use this one from `pkg`, would it not work to import as:
```js
import { SequalizeStorage } from 'umzug';
```

QualityControl/lib/database/SequelizeDatabase.js
```diff
@@ -0,0 +1,131 @@
...
+ * Sequelize implementation of the Database.
+ */
+export class SequelizeDatabase {
+  constructor(config) {
+    this._logger = LogManager.getLogger('qcg/database');
+
+    if (!config) {
+      this._logger.warnMessage('No configuration provided for SequelizeDatabase. Using default configuration.');
+    }
+    this.dbConfig = getDbConfig(config);
```
graduta: Is it not a private field? If so, please use convention `this._dbConfig` for naming

QualityControl/lib/database/SequelizeDatabase.js
```diff
@@ -0,0 +1,131 @@
...
+      host,
+      port,
+      charset,
+      collate,
+      timezone,
+      logging,
+    } = this.dbConfig;
+
+    const __filename = fileURLToPath(import.meta.url);
+    this.__dirname = dirname(__filename);
```
graduta: Any reason why this is defined in the constructor if it is used only in the migrate method?

QualityControl/lib/database/SequelizeDatabase.js
```diff
@@ -0,0 +1,131 @@
...
+          define: {
+            underscored: true,
+          },
+        },
+      );
+      this._models = models(this.sequelize);
+      this._logger.infoMessage('Database connection initialized successfully.');
+    } catch (error) {
+      this._logger.errorMessage(`Error initializing database connection: ${error}`);
+      throw new Error('Error initializing database connection');
```
graduta: Is there any reason we catch and throw the error, rather than leaving it to propagate by itself, especially as now you are propagating an error that does not reflect what happened. I would be in favour of removing the try-catch from constructor

QualityControl/lib/database/SequelizeDatabase.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  /**
```
graduta, line 70: I think documentation can be improved here:
- missing doc on retries and how to configure waiting time
- incorrect doc on when an error is thrown. Firstly, error is never thrown due to infinte loop and always being caught

QualityControl/lib/database/SequelizeDatabase.js
```diff
@@ -0,0 +1,131 @@
...
+   */
+  async connect() {
+    const { retryThrottle } = this.dbConfig;
+    let connected = false;
+
+    while (!connected) {
+      try {
+        await this.sequelize.authenticate();
+        connected = true;
+        this._logger.debugMessage('Connected to db successfully');
```
graduta: I would recommend in the log message that you also specify to which host and port and DB we connected to

graduta: Moreover, you should choose a type of log message for success messages that are part of a flow.
For example, in here you chose `debug` for connecting to DB, while a bit below, you chose `info` for migration 

Please update accordingly

QualityControl/lib/database/index.js
```diff
@@ -0,0 +1,27 @@
...
+ * Initializes the database connection and runs migrations.
+ * @param {object} sequelizeDatabase - The Sequelize database instance.
+ * @returns {Promise<void>} A promise that resolves when the database is initialized.
+ */
+export const initDatabase = async (sequelizeDatabase) => {
+  try {
+    await sequelizeDatabase.connect();
+    await sequelizeDatabase.migrate();
+  } catch {
+    process.exit(1);
```
graduta: I do not think this should control the server but instead log the error. 
The QC GUI should still be up and running and the about page should be the one stating that DB connection is down or something bad happened

graduta (changes requested):

graduta (approved):

PR #2834, opened 2025-04-24
[OGUI-1653] Adapt the emitters and cache to new service and controller

graduta:
- [x] PR labels are selected

PR which:
* removes the attributes `vars` and `defaults` from payload of EnvironmentInfol
* removes the depracted `EnvCache` class in favour of the new one based on emitter;
* update controller for env to use new environment service
* updates front-end to use new endpoint


Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -39,6 +40,40 @@ class EnvironmentCacheService {
...
+        Object.assign(cachedEnvironment, environment);
+        cachedEnvironment.events = events;  
+      } else {
+        cachedEnvironment = { ...environment };
+        if (!cachedEnvironment.events) {
+          cachedEnvironment.events = [];
+        }
+      }
+
+      this._environments.set(id, cachedEnvironment);
```
mariscalromeroalejandro: I think it's safer to avoid modifying the existing object directly

```suggestion
      let updatedEnvironment = {};
const existing = this._environments.get(id);

if (existing) {
  updatedEnvironment = {
    ...existing,
    ...environment,
    events: existing.events || [],
  };
} else {
  updatedEnvironment = {
    ...environment,
    events: environment.events || [],
  };
}

this._environments.set(id, updatedEnvironment);

```

mariscalromeroalejandro (changes requested):

Control/lib/services/environment/EnvironmentCache.service.js
graduta: Thank you for the feedback! I have updated the object.assign to create a new target

mariscalromeroalejandro (approved):

PR #2838, opened 2025-04-28
[OGUI-1654] Integrate partidionStateChange from ODC Kafka events

graduta:
- [x] PR labels are selected

PR which:
* integrates events that represent a partition state change from ODC and displays a panel for Glolba and upwards users to view the progress of ODC state changes
* updates cache accordingly and broadcasts ODC state changes to users

### DIFF
--- a/Control/lib/common/broadcastKeys.enum.js
+++ b/Control/lib/common/broadcastKeys.enum.js
@@ -29,6 +29,9 @@ const BroadcastKeys =  Object.freeze({
   },
   ENVIRONMENT_EVENTS: 'ENVIRONMENT_EVENTS',
   ENVIRONMENTS_OVERVIEW: 'ENVIRONMENTS_OVERVIEW',
+  ODC: {
+    ENVIRONMENT_STATE_CHANGE: 'ODC_ENVIRONMENT_STATE_CHANGE',
+  }
 });
 
 exports.BroadcastKeys = BroadcastKeys;
--- a/Control/lib/kafka/adapters/fromEcsIntegratedServiceEventToEvent.js
+++ b/Control/lib/kafka/adapters/fromEcsIntegratedServiceEventToEvent.js
@@ -1,0 +1,50 @@
+/**
+ *  @license
+ *  Copyright CERN and copyright holders of ALICE O2. This software is
+ *  distributed under the terms of the GNU General Public License v3 (GPL
+ *  Version 3), copied verbatim in the file "COPYING".
+ *
+ *  See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ *  In applying this license CERN does not waive the privileges and immunities
+ *  granted to it by virtue of its status as an Intergovernmental Organization
+ *  or submit itself to any jurisdiction.
+ */
+
+const {LogManager} = require('@aliceo2/web-ui');
+
+/**
+ * Adapter for integrated service event type of message received on integrated_service topic
+ * Adapter's role is to attempt to parse the payload into a JSON and update the timestamp field from BigInt to a number
+ * @param {Event - events.proto} eventMessage - the event message to adapt
+ * @param {Ev_IntegratedServiceEvent - events.proto} eventMessage.integratedServiceEvent - the integrated service event to adapt
+ * @param {BigInt} eventMessage.timestamp - the timestamp of the event message
+ * @return {object} - the adapted event message
+ */
+exports.fromEcsIntegratedServiceEventToEvent = ({ integratedServiceEvent, timestamp }) => {
+  const { payload = '{}' } = integratedServiceEvent;
+  const timestamp = fromBigIntToNumber(timestamp);
+
+  let payloadData = {};
+  try {
+    payloadData = JSON.parse(payload);
+  } catch (error) {
+    const logger = LogManager.getLogger('cog/ecs-integrated-service-event-adapter');
+    logger.errorMessage('Failed to parse payload:', error);
+  }
+  return {
+    ...integratedServiceEvent,
+    timestamp,
+    payload: payloadData,
+  };
+};
+
+/**
+ * Convert a BigInt timestamp to a number
+ * @param {BigInt} bigInt - The BigInt timestamp to convert
+ * @return {number} - The converted number timestamp
+ */
+const fromBigIntToNumber = (bigInt) => {
+  const bigIntTimestamp = BigInt(bigInt.toString(10));
+  return Number(bigIntTimestamp);
+};
--- a/Control/lib/kafka/adapters/fromEcsIntegratedServiceEventToEvent.js
+++ b/Control/lib/kafka/adapters/fromEcsIntegratedServiceEventToEvent.js
@@ -23,7 +23,7 @@ const {LogManager} = require('@aliceo2/web-ui');
  */
 exports.fromEcsIntegratedServiceEventToEvent = ({ integratedServiceEvent, timestamp }) => {
   const { payload = '{}' } = integratedServiceEvent;
-  const timestamp = fromBigIntToNumber(timestamp);
+  const timestampNumber = fromBigIntToNumber(timestamp);
 
   let payloadData = {};
   try {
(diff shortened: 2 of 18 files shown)


Control/lib/common/broadcastKeys.enum.js
```diff
@@ -29,6 +29,9 @@ const BroadcastKeys =  Object.freeze({
   },
   ENVIRONMENT_EVENTS: 'ENVIRONMENT_EVENTS',
   ENVIRONMENTS_OVERVIEW: 'ENVIRONMENTS_OVERVIEW',
+  ODC: {
```
mariscalromeroalejandro, line 32: I would do it the other way around, putting ODC inside ENVIRONMENT_STATE_CHANGE (to maintain the structure)

Control/lib/kafka/adapters/fromEcsIntegratedServiceEventToEvent.js
```diff
@@ -0,0 +1,50 @@
...
+exports.fromEcsIntegratedServiceEventToEvent = ({ integratedServiceEvent, timestamp }) => {
+  const { payload = '{}' } = integratedServiceEvent;
+  const timestampNumber = fromBigIntToNumber(timestamp);
+
+  let payloadData = {};
+  try {
+    payloadData = JSON.parse(payload);
+  } catch (error) {
+    const logger = LogManager.getLogger('cog/ecs-integrated-service-event-adapter');
+    logger.errorMessage('Failed to parse payload:', error);
```
mariscalromeroalejandro, line 33: I would set a new value to payloadData so it also contains the info about the error ??

Control/lib/services/Environment.service.js
```diff
@@ -59,21 +59,38 @@ class EnvironmentService {
...
       const { environments } = await this._coreGrpc.GetEnvironments({ showTaskInfos });
-      const detectorsAll = this._apricotGrpc.detectors ?? [];
-      const hostsByDetector = this._apricotGrpc.hostsByDetector ?? {};
       
-      const environmentsInfoList = environments.map(
-        (environment) => EnvironmentInfoAdapter.toEntity(environment, '', detectorsAll, hostsByDetector)
-      );
-      if (shouldUpdateCache) {
-        this._environmentCacheService.environments = environmentsInfoList;
+      const environmentList = [];
```
mariscalromeroalejandro, line 78: I would return if environments are not available `if (!environments || environments.length === 0)`

Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -41,34 +45,62 @@ class EnvironmentCacheService {
...
+  }
 
+  /**
+   * Update an environment in the cache by its id
+   * @param {string} id - the id of the environment to be updated
+   * @param {EnvironmentInfo} environment - the new environment information to be set
+   * @returns {void}
+   */
+  addOrUpdateEnvironment(environment) {
+    const { id } = environment;
```
mariscalromeroalejandro, line 63: return or throw if id not found

Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -41,34 +45,62 @@ class EnvironmentCacheService {
...
+    if (this._environments.has(id)) {
+      const cachedEnvironment = this._environments.get(id);
+      const { events = [] } = cachedEnvironment;
+      const updatedEnvironment = Object.assign({}, cachedEnvironment, environment);
+      updatedEnvironment.events = [...events];
       this._environments.set(id, updatedEnvironment);
-    });
-    this._broadcastService.broadcast(ENVIRONMENTS_OVERVIEW, environments);
+    } else {
+      this._environments.set(id, { ...environment, events: environment.events ?? [] });
```
mariscalromeroalejandro, line 71: I would save `const existingEvents = environment.events ?? []` in a constant and then I will use it like [...existingEvents]

Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -41,34 +45,62 @@ class EnvironmentCacheService {
...
-      }
+  get environments() {
+    return this._environments;
+  }
 
+  /**
+   * Update an environment in the cache by its id
+   * @param {string} id - the id of the environment to be updated
+   * @param {EnvironmentInfo} environment - the new environment information to be set
+   * @returns {void}
```
mariscalromeroalejandro, line 60: what about returning true/false so you know if adding an environment was successful?

Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -41,34 +45,62 @@ class EnvironmentCacheService {
...
+   * Given this information, the method should update the environment in the cache
+   * @param {string} id - the id of the environment to be updated
+   * @param {string} attributePath - the attribute path to be updated in the form of "key1.key2.key3"
+   * @param {string} value - the new value to be set
+   * @returns {void}
    */
-  get environments() {
-    return this._environments;
+  _updateAttributeOfEnvironment(id, attributePath, value) {
+    if (this._environments.has(id)) {
```
mariscalromeroalejandro, line 88: check id not null

Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -94,6 +126,28 @@ class EnvironmentCacheService {
...
+      /**
+       * @private
+       * Event object is as per type returned by fromEcsIntegratedServiceEventToEvent
+       * @param {object} event - the event object containing the payload and environmentId
+       */
+      (event) => {
+        const { payload, environmentId } = event;
+        const environmentUpdated = this._updateAttributeOfEnvironment(
+          environmentId,
+          'hardware.epn.info',
```
mariscalromeroalejandro: can this hardcoded atributte be stored in a constant?

Control/lib/services/environment/EnvironmentCache.service.js
```diff
@@ -94,6 +126,28 @@ class EnvironmentCacheService {
...
+          {
+            state: payload?.state,
+            ddsSessionId: payload?.ddsSessionId,
+            ddsSessionStatus: payload?.ddsSessionStatus,
```
mariscalromeroalejandro: destructuring the properties of the payload object?

mariscalromeroalejandro (changes requested):

Control/lib/common/broadcastKeys.enum.js
graduta: Environment state change is what ODC calls the operation and ODC has multiple operations will be added in the next PRs. Thus, I prefer to keep all ODC operations grouped into this one

Control/lib/kafka/adapters/fromEcsIntegratedServiceEventToEvent.js
graduta: If `payload` is undefined (which happens quite often), then we would display an error that in reality is not an error.
The actual error if there is one is sent by ECS at the same level as payload in attribute `error`. 
Thus, in this case it is better to ignore the lack or improper structure of payload if any

Control/lib/services/environment/EnvironmentCache.service.js
graduta: There is no check on this as the service is not interested to know if it was added or not

Control/lib/services/environment/EnvironmentCache.service.js
graduta: ID is mandatory in the gRPC condition from ECS, thus it cannot be non-existent

Control/lib/services/environment/EnvironmentCache.service.js
graduta: Any particular reason behind it? We would add an extra operation with no benefit. At least readability I do not consider it would be improved

Control/lib/services/environment/EnvironmentCache.service.js
graduta: Any reason for it?
If id is null or undefined, the `has` method will simply not be true

Control/lib/services/environment/EnvironmentCache.service.js
graduta: Yes, good catch

mariscalromeroalejandro (approved):

PR #2839, opened 2025-04-28
[OGUI 1655] update LayoutList REST to only send required fields

NarrowsProjects:
- [x] PR labels are selected

This PR is a replaces the fetch requests from the LayoutListPage to only send the required data rather than the full Layout object. The the required data includes the following properties:
- id {number} - Layout identifier
- name {string} - Layout name
- owner_id {number} - Owner identifier
- owner_name {string} - Owner name
- description {string} - Layout description
- isOfficial {boolean} - Official status flag


QualityControl/lib/controllers/LayoutController.js
```diff
@@ -68,6 +68,28 @@ export class LayoutController {
...
+   */
+  async getLayoutCardsHandler(req, res) {
+    try {
+      const filter = {};
+      if (req.query.owner_id !== undefined) {
+        filter.owner_id = parseInt(req.query.owner_id, 10);
+      }
+      const layoutCards = await this._layoutRepository.listLayoutCards(filter);
+      res.status(200).json(layoutCards);
+    } catch {
```
graduta: if you are not logging or returning the actual errors, we will never know what was the issue.
I would recommend that you catch the error and either:
- send it as response
- log it in debug mode so that can be aware of the issue

QualityControl/lib/api.js
```diff
@@ -51,6 +51,7 @@ export const setup = (http, ws) => {
   http.get('/object', objectController.getObjectContent.bind(objectController));
   http.get('/objects', objectController.getObjects.bind(objectController), { public: true });
 
+  http.get('/layoutcards', layoutController.getLayoutCardsHandler.bind(layoutController));
```
graduta: Rather than removing the existing endpoint `/layouts` I would be more in favour of updating it to something similar to what GraphQL does in which the client specifies which fields should be retrieved.
In this way we can keep the previous compatibility if we will need it while also adding the new feature

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -29,6 +29,34 @@ export class LayoutRepository extends BaseRepository {
             && (filter.name === undefined || layout.name === filter.name));
   }
 
+  /**
+   * Retrieves simplified layout card objects containing only essential properties, optionally filtered
+   * @param {object} filter - Filter criteria with optional keys:
+   *   * owner_id {number} - Filter by owner ID
```
graduta: To be more clear what is being used from the filter, I wold recommend deconstructing the JSON in the method declaration.
Then the docs could also be updated similar to:
```js
* @param {object} filter - Filter criteria with optional keys:
* @param {string} filter.name - Name of the layout that is to be filtered by
* @param {string} filter.owner_id - ID of the owner of the layout to filter by
[...]
listLayoutCards({name = undefined, owner_id = undefiend) {
```

QualityControl/public/layout/Layout.js
```diff
@@ -313,8 +313,8 @@ export default class Layout extends Observable {
     const { isOfficial } = this.model.services.layout.list.payload.find((item) => item.id === id);
 
     await this.model.services.layout.patchLayout(id, { isOfficial: !isOfficial });
-    await this.model.services.layout.getLayouts(this);
-    await this.model.services.layout.getLayoutsByUserId(this.model.session.personid, this);
+    await this.model.services.layout.getLayoutCards(this);
+    await this.model.services.layout.getLayoutCardsByUserId(this.model.session.personid, this);
```
graduta: I believe here we can improve and drop the second promise. (I know it was like this already, but I see no reason not to improve it)
Basically the first promise request will return you all layouts. Thus, you can create your list of layoutsByUserId from that promise.

graduta (changes requested):
Hi @NarrowsProjects good proposal! Could you please have a look at what it would mean to implement the endpoint to accept fields that should be returned? Something similar is done by graphql as an example but do not integrate the entire library just for this. Look at the concept

QualityControl/lib/api.js
NarrowsProjects: I'm not removing an endpoint, I'm adding a new endpoint.
Admittedly though, the layouts endpoint isn't used outside of the data things I changed.

QualityControl/lib/controllers/LayoutController.js
NarrowsProjects: Should I also add this same change to the endpoint above?

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -63,7 +66,18 @@ export class LayoutController {
       }
       const layouts = await this._layoutRepository.listLayouts(filter);
       res.status(200).json(layouts);
-    } catch {
+    } catch (e) {
+      if (e instanceof TypeError && e.message === 'fields parameter must be an array') {
```
graduta: For input validation in QCG, we use `joi` library. Have a look in this file at methods such as:
- `putLayoutHandler`
- `postLayoutHandler`

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -19,14 +19,46 @@ import { BaseRepository } from './BaseRepository.js';
...
+   * @throws {Error} If any specified field does not exist in the layout objects
    */
-  listLayouts(filter = {}) {
-    return this._jsonFileService.data.layouts.filter((layout) =>
-      (filter.owner_id === undefined || layout.owner_id === filter.owner_id)
-            && (filter.name === undefined || layout.name === filter.name));
+  listLayouts({ name, owner_id, fields } = {}) {
+    const { layouts } = this._jsonFileService.data;
+
+    const filter = (layout) =>
```
graduta: Perhaps you can name `filter` to something more descriptive on what is filtering on?

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -19,14 +19,46 @@ import { BaseRepository } from './BaseRepository.js';
...
+      (owner_id === undefined || layout.owner_id === owner_id) &&
+      (name === undefined || layout.name === name);
+
+    const filteredLayouts = layouts.filter(filter);
+
+    if (fields && !Array.isArray(fields)) {
```
graduta: You are already checking in the controller for the user input type

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -19,14 +19,46 @@ import { BaseRepository } from './BaseRepository.js';
...
+    }
+
+    if (!fields || fields.length === 0) {
+      return filteredLayouts;
+    }
+
+    filteredLayouts.forEach((layout) =>{
+      const missingField = fields.find((field) => !(field in layout));
+      if (missingField) {
+        throw new Error(`The following field does not exist for layouts: ${missingField}`);
```
graduta: I believe we should not throw an error if some of the layouts is missing a field. 
I think we should return the layouts given the requested fields on what it exists. For example if a layout is missing a field such as `description`, we should not throw an error but instead return the layout without the `description` attribute

QualityControl/public/services/Layout.service.js
```diff
@@ -37,22 +37,29 @@ export default class LayoutService {
   }
 
   /**
-   * Method to get all layouts shared between users
+   * Method to get all layout cards shared between users
    * @param {Class<Observable>} that - Observer requesting data that should be notified of changes
    * @returns {undefined}
    */
-  async getLayouts(that = this.model) {
+  async getLayoutCards(that = this.model) {
```
graduta: I would not name the method of a `service` to the type of component you display. 
Instead it should be named by the data it retrieves. Thus, I think it is better to stick to `getLayouts`

QualityControl/public/services/Layout.service.js
```diff
@@ -37,22 +37,29 @@ export default class LayoutService {
...
     this.list = RemoteData.loading();
     that.notify();
+    const cardfields = 'id,name,owner_id,owner_name,description,isOfficial';
```
graduta: The same goes in here, rather then `cardFields`, perhaps something like `requestedFields` or `fieldsToRetrieve`, etc.

QualityControl/public/services/Layout.service.js
```diff
@@ -88,6 +95,35 @@ export default class LayoutService {
     that.notify();
   }
 
+  /**
+   * Method to get all layouts as layoutCards by the user's id
+   * @param {string} userId - user id for which to query layouts
+   * @param {Class<Observable>} that - Observer requesting data that should be notified of changes
+   * @returns {undefined}
+   */
+  async getLayoutCardsByUserId(userId, that = this.model) {
```
graduta: Similar comment to above for method name

graduta (changes requested):

QualityControl/public/services/Layout.service.js
```diff
@@ -88,6 +95,35 @@ export default class LayoutService {
...
+      const cardfields = 'id,name,owner_id,owner_name,description,isOfficial';
+
+      const { result, ok } = await this.loader.get(`/api/layouts?owner_id=${userId}&fields=${cardfields}`);
+      if (ok) {
+        const sortedLayouts = result.sort((lOne, lTwo) => lOne.name > lTwo.name ? 1 : -1);
+        this.userList = RemoteData.success(sortedLayouts);
+        this.model.folder.map.get('My Layouts').list = RemoteData.success(sortedLayouts);
+      } else {
+        this.userList = RemoteData.failure(result.error || result.message);
+        this.model.folder.map.get('My Layouts').list = RemoteData.failure(result.error || result.message);
```
mariscalromeroalejandro: You're calling RemoteData.failure(...) twice with the same message. Assign it to a variable

QualityControl/public/services/Layout.service.js
```diff
@@ -88,6 +95,35 @@ export default class LayoutService {
...
+    that.notify();
+
+    if (isNaN(userId)) {
+      this.userList = RemoteData.failure('Provided userId is not a number');
+    } else {
+      const cardfields = 'id,name,owner_id,owner_name,description,isOfficial';
+
+      const { result, ok } = await this.loader.get(`/api/layouts?owner_id=${userId}&fields=${cardfields}`);
+      if (ok) {
+        const sortedLayouts = result.sort((lOne, lTwo) => lOne.name > lTwo.name ? 1 : -1);
```
mariscalromeroalejandro: maybe use `localCompare` to avoid bugs like character casing, accents...

QualityControl/test/lib/repositories/LayoutRepository.test.js
```diff
@@ -46,8 +46,31 @@ export const layoutRepositoryTest = async () => {
     });
 
     suite('list layouts', () => {
+      test('should throw TypeError when fields is not an array', () => {
+        throws(
+          () => layoutRepository.listLayouts({ fields: 'not an array' }),
+          {
+            name: 'TypeError',
+            message: 'fields parameter must be an array',
```
mariscalromeroalejandro: I think here you can use `new TypeError('fields parameter...')` and reduce it to one line (cleaner)

QualityControl/lib/dtos/LayoutDto.js
```diff
@@ -49,3 +77,14 @@ export const LayoutDto = Joi.object({
   displayTimestamp: Joi.boolean().default(false),
   autoTabChange: Joi.number().min(0).max(600).default(0),
 });
+
+export const LayoutsGetDto = Joi.object({
```
mariscalromeroalejandro, line 84: add documentation for this method

QualityControl/lib/controllers/LayoutController.js
graduta: Where do you mean?

NarrowsProjects: It's been a bit since i wrote this, but i most likely meant that the function **Below** this one has the exact same error handeling I tried to implement here.

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -57,14 +60,29 @@ export class LayoutController {
    */
   async getLayoutsHandler(req, res) {
     try {
+      const { owner_id, fields } = await LayoutsGetDto.validateAsync(req.query);
```
graduta: I think it is better to separate in multiple try/catch blocks the potential errors thrown. 
So in this case:
- a try-catch for the JOI validation
- a try-catch for the `listLayouts`

It will mke the code easier to read

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -29,6 +30,8 @@ import {
  * @typedef {import('../repositories/LayoutRepository.js').LayoutRepository} LayoutRepository
  */
 
+const logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'qcg'}/layout`);
```
graduta: The term after ``qcg`/` should describe in very short the class using the logger.
So in this case, it is better to be something like `layout-ctrl` where ctrl is a short of controller.

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -11,22 +11,53 @@
  * or submit itself to any jurisdiction.
  */
 
-import { NotFoundError } from '@aliceo2/web-ui';
+import { LogManager, NotFoundError } from '@aliceo2/web-ui';
 import { BaseRepository } from './BaseRepository.js';
 
+const logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'qcg'}/layout`);
```
graduta: Same in here, we need to tell the logger who the class using it is. `layout-repo`

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -11,22 +11,53 @@
...
+          ${layout.id ?? '<no id found>'}: ${missingField}`);
+      }
+    });
+
+    return filteredLayouts.map((layout) => {
+      const layoutObj = {};
+      fields.forEach((field) => {
+        layoutObj[field] = layout[field];
+      });
+      return layoutObj;
```
graduta: Is there any reason why you are not logging inside here so that we do not iterate twice over all layouts?

QualityControl/public/layout/Layout.js
```diff
@@ -313,8 +313,10 @@ export default class Layout extends Observable {
     const { isOfficial } = this.model.services.layout.list.payload.find((item) => item.id === id);
 
     await this.model.services.layout.patchLayout(id, { isOfficial: !isOfficial });
-    await this.model.services.layout.getLayouts(this);
-    await this.model.services.layout.getLayoutsByUserId(this.model.session.personid, this);
+    const layoutFields = 'id,name,owner_id,owner_name,description,isOfficial';
```
graduta: As you are re-using the layoutFields, I think it would be better to define an enum class in `common` and import it and use it here and in `Model.js`

QualityControl/public/Model.js
```diff
@@ -169,15 +169,16 @@ export default class Model extends Observable {
   async handleLocationChange() {
     this.object.objects = {}; // Remove any in-memory loaded objects
     clearInterval(this.layout.tabInterval);
+    const layoutFields = 'id,name,owner_id,owner_name,description,isOfficial';
```
graduta: See comment in Layout.js for the layoutsFields

QualityControl/public/services/Layout.service.js
```diff
@@ -64,19 +73,25 @@ export default class LayoutService {
...
     if (isNaN(userId)) {
       this.userList = RemoteData.failure('Provided userId is not a number');
     } else {
-      const { result, ok } = await this.loader.get(`/api/layouts?owner_id=${userId}`);
+      const url = `/api/layouts?owner_id=${userId}${fields !== undefined ? `&fields=${fields}` : ''}`;
+
+      const { result, ok } = await this.loader.get(url);
       if (ok) {
-        const sortedLayouts = result.sort((lOne, lTwo) => lOne.name > lTwo.name ? 1 : -1);
+        const layoutComparitor = (layout1, layout2) => layout1.name.localeCompare(layout2.name);
```
graduta: I see that you are implementing two comparitors which aim to sort by name but in two different manners:
Here and a few lines above. 

Best is to extract this in a util function and reuse it from there

graduta (changes requested):
I made a few comments but more importantly, you should pay attention to the call that you make from layout controller to the layout service. 
You are calling it as:
```
const layouts = await this._layoutRepository.listLayouts(filter, fields);
```
But the layout service definition is:
```
listLayouts({ name, owner_id, fields } = {})
```

So the issue is that in the way you are calling it, fields is never used and you still return the entire layout objects.
Have a look at the tests as well, the tests should have caught this mistake

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -11,22 +11,51 @@
...
+   * @param {Array<string>} [options.fields] - Array of field names to include in each returned layout object
+   * @returns {Array<object>} Array of layout objects matching the filters, containing only the specified fields
+   * @throws {TypeError} If fields parameter is provided but is not an array
+   * @throws {Error} If any specified field does not exist in the layout objects
    */
-  listLayouts(filter = {}) {
-    return this._jsonFileService.data.layouts.filter((layout) =>
-      (filter.owner_id === undefined || layout.owner_id === filter.owner_id)
-            && (filter.name === undefined || layout.name === filter.name));
+  listLayouts({ name, owner_id, fields } = {}) {
```
graduta, line 33: You are still missing the correct call of this. In layout controller you call it with `{ fields, filter }` which is not the same as here
Because of this, currently all layouts are returned and displayed in the sidebar as my layouts. 
You can test this by modifing the `db.json` and change the owner_name and owner_id to something different the Anonymous and 0

graduta (changes requested):

graduta (approved):

graduta (changes requested):
You should have `fields` in the API request optional. Otherwise you are introducing a bug. 
To replicate:
- open a layout
- Edit it and save
- check sideber (empty, no layouts of the user anymore)
- in navigation tab you will see that a 400 status code is returned due to missing `fields` on request of retrieving latest version of layouts. 

I would propose that the server returns the minium information on a layout if the fields request parameter is missing
It would be good to also follow the tests we have for the API (Put,Patch) and add tests for GET layouts: https://github.com/AliceO2Group/WebUi/tree/dev/QualityControl/test/api/layouts

QualityControl/test/api/layouts/api-get-layout.test.js
```diff
@@ -0,0 +1,193 @@
...
+import { OWNER_TEST_TOKEN, URL_ADDRESS } from '../config.js';
+import request from 'supertest';
+
+export const apiGetLayoutsTests = () => {
+  suite('GET /layouts', () => {
+    test('should return all layouts when no filters are provided', async () => {
+      await request(`${URL_ADDRESS}/api/layouts`)
+        .get(`?token=${OWNER_TEST_TOKEN}`)
+        .expect(200)
+        .expect((res) => {
```
mariscalromeroalejandro, line 27: it would be more robust and readable to assert the full expected response structure directly

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -94,14 +94,14 @@ export class LayoutController {
     const { id } = req.params;
 
     try {
-      if (!id) {
+      if (!id.trim()) {
         updateAndSendExpressResponseFromNativeError(res, new InvalidInputError('Missing parameter "id" of layout'));
```
mariscalromeroalejandro: I would move this check outside the try catch block as it does not throw and it is safe

QualityControl/test/test-index.js
```diff
@@ -152,7 +153,7 @@ suite('All Tests - QCG', { timeout: FRONT_END_TIMEOUT + BACK_END_TIMEOUT }, asyn
     );
 
     suite('API - Test Suite', async () => {
-      suite('Layout PUT request test suite', async () => apiPutLayoutTests());
+      suite('Layout PUT request test suite', async () => apiGetLayoutsTests());
```
mariscalromeroalejandro: please, change also the description of the suite

QualityControl/test/api/layouts/api-get-layout.test.js
```diff
@@ -44,7 +44,12 @@ export const apiGetLayoutsTests = () => {
             throw new Error('Expected array of layouts');
           }
 
-          deepStrictEqual(res.body, [LAYOUT_MOCK_4, LAYOUT_MOCK_5]);
+          deepStrictEqual(
+            res.body,
+            [LAYOUT_MOCK_4, LAYOUT_MOCK_5],
+            `Expected layouts with owner_id=0 to be ${JSON.stringify([LAYOUT_MOCK_4, LAYOUT_MOCK_5])}
```
mariscalromeroalejandro: I think it is not good to print all the layouts expected and received if it fails, just say that they does not match. This is good for debugging but not as an error message

QualityControl/test/api/layouts/api-get-layout.test.js
```diff
@@ -102,7 +107,12 @@ export const apiGetLayoutsTests = () => {
         .get(`?token=${OWNER_TEST_TOKEN}`)
         .expect(200)
         .expect((res) => {
-          deepStrictEqual(res.body, LAYOUT_MOCK_6);
+          deepStrictEqual(
+            res.body,
+            LAYOUT_MOCK_6,
+            `Expected layout with id '671b8c22402408122e2f20dd' to be ${JSON.stringify(LAYOUT_MOCK_6)}
```
mariscalromeroalejandro: same here

QualityControl/test/lib/controllers/LayoutController.test.js
```diff
@@ -47,39 +48,126 @@ export const layoutControllerTestSuite = async () => {
...
+      };
+      const layoutConnector = new LayoutController(jsonStub);
+
+      await layoutConnector.getLayoutsHandler(req, res);
+
+      ok(res.status.calledWith(400), 'Response status was not 400');
+      ok(res.json.calledOnce, 'Response was not sent');
+
+      const [[responseArg]] = res.json.args;
+      ok(
```
mariscalromeroalejandro: To keep it in sync with the rest of the code, could you please update it to use` ok(res.json.calledWith({ `as well?

QualityControl/test/lib/repositories/LayoutRepository.test.js
```diff
@@ -67,6 +67,18 @@ export const layoutRepositoryTest = async () => {
...
+      test('should return only specified fields when fields array is provided', () => {
+        const fields = ['id', 'name'];
+        const result = layoutRepository.listLayouts({ fields });
+
+        result.forEach((layout) => {
+          const actualKeys = Object.keys(layout);
+          deepStrictEqual(actualKeys.sort(), fields);
+        });
+
+        equal(result.length, jsonFileServiceMock.data.layouts.length);
```
mariscalromeroalejandro:
```suggestion
        strictEqual(result.length, jsonFileServiceMock.data.layouts.length);
```

mariscalromeroalejandro (changes requested):

mariscalromeroalejandro (approved):

PR #2848, opened 2025-04-29
[OGUI-1129] Layouts do not preserve object arrangement

NarrowsProjects:
- [x] PR labels are selected

This PR aimed to fix a visual bug that would alter the LayoutView page object-layout when switching between tabs or switching to the same tab.

### DIFF
--- a/QualityControl/public/layout/Grid.js
+++ b/QualityControl/public/layout/Grid.js
@@ -153,7 +153,7 @@ GridList.prototype = {
           position = this._getItemPosition(item);
 
       this._updateItemPosition(
-        item, this.findPositionForItem(item, {x: position.x, y: 0}));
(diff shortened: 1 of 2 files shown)


graduta (changes requested):
Hi @NarrowsProjects , I believe removing this bit of code will create issues in EDIT mode. 
For example, try to:
- Edit
- Edit via GUI
- use the left sidebar tree to add new objects by double clicking on them. 

You will see that these objects are not added and the panel transforms in a very long panel with a scroll bar. The current behaviour of the tool is that on double click, an object is added to the layout

mariscalromeroalejandro (dismissed):

graduta (changes requested):
Issue still happening (described in group chat)

graduta (approved):

PR #2849, opened 2025-05-02
[OGUI-1657] Migrate to use tree API from CCDB

NarrowsProjects:
- [x] PR labels are selected

In this PR, the 'latest' API from CCDB is swapped out with the /tree API. 

It removes the sort by date option from the ObjectTree page header, removes tests from the previous implementation, and adds new tests.

### DIFF
--- a/QualityControl/lib/services/ccdb/CcdbService.js
+++ b/QualityControl/lib/services/ccdb/CcdbService.js
@@ -12,7 +12,7 @@
  * or submit itself to any jurisdiction.
  */
 
-import { LogManager } from '@aliceo2/web-ui';
+import { FailedDependencyError, LogManager } from '@aliceo2/web-ui';
 import { httpHeadJson, httpGetJson } from '../../utils/utils.js';
 import {
   CCDB_MONITOR, CCDB_VERSION_KEY, CCDB_RESPONSE_BODY_KEYS, CCDB_FILTER_FIELDS, CCDB_RESPONSE_HEADER_KEYS,
@@ -108,7 +108,7 @@ export class CcdbService {
     const { subfolders } = await httpGetJson(this._hostname, this._port, `/tree/${prefix}.*`);
 
     if (!Array.isArray(subfolders)) {
-      throw new Error('Invalid response format from server - expected subfolders array');
+      throw new FailedDependencyError('Invalid response format from server - expected subfolders array');
     }
     return subfolders.map((folder) => ({ path: folder }));
   }
--- a/QualityControl/lib/services/ccdb/CcdbService.js
+++ b/QualityControl/lib/services/ccdb/CcdbService.js
@@ -122,6 +122,22 @@ export class CcdbService {
   }
 
   /**
+   * Returns a list of object paths using the /tree/ endpoint based on a given prefix
+   * @example Equivalent of URL request: `/tree/qc/TPC/object.*`
+   * @param {string} [prefix] - Prefix for which CCDB should search for objects
+   * @returns {Promise.<Array<{PATH}>>} - results of objects query or an empty array
+   * @rejects {Error}
+   */
+  async getObjectsTreeList(prefix = this._PREFIX) {
+    const { subfolders } = await httpGetJson(this._hostname, this._port, `/tree/${prefix}.*`);
+
+    if (!Array.isArray(subfolders)) {
+      throw new Error('Invalid response format from server - expected subfolders array');
+    }
+    return subfolders.map((folder) => ({ path: folder }));
+  }
+
+  /**
    * Retrieve a sorted list of identifications for a specified object which represent different versions of the object;
    * Number of versions defaults to a limit but it can be changed by passing a value
    * @example Equivalent of URL request: `/browse/qc/TPC/object/14324234234234234/id-id-id-id-id/RunNumber=554345`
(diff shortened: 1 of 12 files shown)


QualityControl/public/object/QCObject.js
```diff
@@ -119,25 +119,37 @@ export default class QCObject extends Observable {
...
+  /**
+   * Helper method for sortListByField for sorting strings
+   * @param {string} a - first string to be sorted
+   * @param {string} b - second string to be sorted
+   * @param {string} field - filed by which the sort should be done
+   * @param {number} order - acending (1) or decending (-1)
+   * @returns {undefined}
+   */
+  _compareStrings(a, b, field, order) {
+    return a[field].toUpperCase().localeCompare(b[field].toUpperCase()) * order;
```
mariscalromeroalejandro: If a method is only used once and its implementation is just a single line, consider inlining it directly where it's called

QualityControl/public/object/QCObject.js
```diff
@@ -119,25 +119,37 @@ export default class QCObject extends Observable {
...
+  /**
+   * Helper method for sortListByField for sorting numbers
+   * @param {number} a - first number to be sorted
+   * @param {number} b - second number to be sorted
+   * @param {string} field - filed by which the sort should be done
+   * @param {number} order - acending (1) or decending (-1)
+   * @returns {undefined}
+   */
+  _compareNumbers(a, b, field, order) {
+    return (a[field] - b[field]) * order;
```
mariscalromeroalejandro: same

mariscalromeroalejandro (changes requested):

QualityControl/public/object/QCObject.js
NarrowsProjects: Thanks for the review, but I would argue that  these two helper functions make the code much more readable. But you have more experience as a developer than I do, so If you insist, I'll change it.

QualityControl/lib/services/ccdb/CcdbService.js
graduta, line 1: I had a look on tasks that we will tackle in the next few weeks and I think this method will come in handy soon (retrieving latest)
Thus, I think for the CCDBService, I would recommend that rather than modifying the existing method, to add a new one based on `tree`

QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -98,27 +98,19 @@ export class CcdbService {
...
-      accept: 'application/json',
-      'x-filter-fields': fields.length > 0 ? fields.join(',') : `${PATH},${CREATED},${LAST_MODIFIED}`,
-    };
-    const { objects } = await httpGetJson(this._hostname, this._port, `/latest/${prefix}.*`, headers);
-    return objects;
+  async getObjectsTreeList(prefix = this._PREFIX) {
+    const { subfolders } = await httpGetJson(this._hostname, this._port, `/tree/${prefix}.*`);
+
+    if (!Array.isArray(subfolders)) {
+      throw new Error('Invalid response format from server - expected subfolders array');
```
graduta: In [webui/framework - errors](https://github.com/AliceO2Group/WebUi/tree/dev/Framework/Backend/errors) we have custom errors defined. I think it would be best to have a look at them and throw one from there

QualityControl/public/object/QCObject.js
graduta: I agree with @NarrowsProjects in this scenario, that separating the two methods improves readability. 
But, I would recommend as an improvement to specifically check if the field is indeed a number field. What if instead is a different type?

QualityControl/public/object/QCObject.js
```diff
@@ -119,25 +119,37 @@ export default class QCObject extends Observable {
...
+
+  /**
+   * Helper method for sortListByField for sorting numbers
+   * @param {number} a - first number to be sorted
+   * @param {number} b - second number to be sorted
+   * @param {string} field - filed by which the sort should be done
+   * @param {number} order - acending (1) or decending (-1)
+   * @returns {undefined}
+   */
+  _compareNumbers(a, b, field, order) {
```
graduta: Would it not better be to pass the objects to compare directly? `a[field]` and `b[field]` instead?

graduta (changes requested):

QualityControl/lib/services/ccdb/CcdbService.js
NarrowsProjects: Which one?

InvalidInputError
UnauthorizedAccessError
NotFoundError
TimeoutError
ServiceUnavailableError

None of these fit.

Or should I add a custom error?

QualityControl/public/object/QCObject.js
NarrowsProjects: Fair point

QualityControl/lib/services/ccdb/CcdbService.js
graduta: What do you think about "Failed Dependency"? In this way the response code to the user would be 424 which is our scenario here: The request from QCG failed because of a failed resposne from the CCDB dependency

Right now it is not part of the CustomErrors that we have, but we can add it

NarrowsProjects: Failed dependency makes sense, I'll get to it

graduta (approved):
Good job!

PR #2876, opened 2025-05-14
[OGUI-1662] Object tree breaking when sorting by name in descending order

NarrowsProjects:
- [x] PR labels are selected

This PR attempts to fix the visual bug where sorting the object tree resulted result in the object tree displaying duplicate dysfunctional trees, or displaying them in the wrong order. 

The changes in this PR include:
- A rework on how the object trees are sorted. 
  - The tree used to be destroyed and then rebuild in the new order. Now the tree stays intact and each subTree is responsible for ordering its own children. 
- The this.object field has been removed from tree/subtrees. In favour of storing objects directly as children. This prevents the majority of object tree instantiations (each individual object used to have their own subtree containing **only** that object).
- The ObjectTree view has been changed to render nested lists instead of a single table. 
- Sorting the Object tree no longer forces all expanded the subtrees to collapse.

Unintended results.
The view change also appears to have fixed the OGUI-1661 issue by accident.


QualityControl/public/object/ObjectTree.class.js
```diff
@@ -144,5 +147,29 @@ export default class ObjectTree extends Observable {
    */
   addChildren(objects) {
     objects.forEach((object) => this.addChild(object));
+    this.notify();
+  }
+
+  sortChildren(field, order) {
+    this.open = this.name === 'qc' ? true : false;
```
mariscalromeroalejandro: I don't think it's appropriate to evaluate whether to show or hide based on the name property as this might change in the future. It would be better to consider another characteristic to make this decision

NarrowsProjects: This was the way the original worked, but I suppose I could check if it is the second lowest by checking if the parent tree has a parent

QualityControl/public/pages/objectTreeView/component/objectTreeItem.js
```diff
@@ -46,14 +46,15 @@ export const leafItem = (leafObject) => {
  * @returns {vnode} - virtual node element
  */
 export const sideTreeLeafItem = (leafObject) => {
-  const { name, path } = leafObject;
+  const { name } = leafObject;
   const { object, layout } = model;
   const displayName = name.split('/').pop();
```
mariscalromeroalejandro, line 54: you are assuming that the name uses a "/" as a delimiter, please do a previous check or indicate it in the documentation like: `leafObject should have a property 'name' with a delimiter /`

QualityControl/public/pages/objectTreeView/component/objectTreeItem.js
```diff
@@ -0,0 +1,77 @@
...
+  ]);
+};
+
+/**
+ * Creates a list item for a leafObject (end node that represents an object)
+ * @param {object} leafObject - the leaf object
+ * @param {QCObject} object - object managing model
+ * @returns {vnode} - virtual node element
+ */
+export const leafItem = (leafObject, object) => {
```
mariscalromeroalejandro: rename the second parameter as it can be confusing

mariscalromeroalejandro (commented):
overall code reads well and follows good practices

QualityControl/lib/dtos/QCObjectNameDto.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+      if (hasWhitespace) {
+        logger.debugMessage(`Invalid path: Segment contains whitespace (${segment}) in path (${path})`);
+        return false;
+      }
+      segmentStart = i + 1;
+    }
+  }
+
+  if (char === '/') {
```
mariscalromeroalejandro:
```suggestion
  if (path.endsWith('/'))
```

QualityControl/lib/dtos/QCObjectNameDto.js
```diff
@@ -0,0 +1,82 @@
...
+ * Validates if a path segment is valid (non-empty and no whitespace)
+ * @param {string} path - The path to validate
+ * @returns {boolean} True if the path is valid
+ */
+function isValidPath(path) {
+  if (typeof path !== 'string' || path === '') {
+    logger.debugMessage(`Invalid path: Path is empty or not a string (${path})`);
+    return false;
+  }
+
```
mariscalromeroalejandro, line 30: I would do:
`const segments = path.split('/');
  for (const segment of segments) {`

and check `if !segment` or i`f (/\s/.test(segment)`

more easy and readable

mariscalromeroalejandro (changes requested):

QualityControl/lib/dtos/QCObjectNameDto.js
NarrowsProjects: You are right about the readability, but the reason why I went character by character is because this will check thousands of object paths, and character loops are more efficient than using split, as no new array is created.

mariscalromeroalejandro: you are right, as it is hit many times per second your character loop fits better

mariscalromeroalejandro (dismissed):

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Method to tackle various states for the selected objects
+ * @param {Model} model - root model of the application
+ * @returns {vnode} - virtual node element
+ */
+export function objectPanel(model) {
+  const selectedObjectName = model.object.selected.name;
+  if (model.object.objects && model.object.objects[selectedObjectName]) {
+    return model.object.objects[selectedObjectName].match({
```
mariscalromeroalejandro, line 17: consider store this: `model.object.objects[selectedObjectName]` in a variable and check its value before using it

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,84 @@
...
+ * @param {Model} model - root model of the application
+ * @param {JSON} object - {qcObject, info, timestamps}
+ * @returns {vnode} - virtual node element
+ */
+export const drawPlot = (model, object) => {
+  const { name, validFrom, id } = object;
+  const href = validFrom ?
+    `?page=objectView&objectName=${name}&ts=${validFrom}&id=${id}`
+    : `?page=objectView&objectName=${name}`;
+  const info = object;
```
mariscalromeroalejandro: is this necessary?

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,84 @@
...
+        'a.btn',
+        {
+          title: 'Open object plot in full screen',
+          href,
+          onclick: (e) => model.router.handleLinkEvent(e),
+        },
+        iconResizeBoth(),
+      )),
+    ]),
+    h('', { style: 'height:77%;' }, draw(model, name, { stat: true })),
```
mariscalromeroalejandro: consider break the return into named constants:
* openButton
* plotSection 
* infoPanel
etc

(readability)

`return h('', { style: 'height:100%; display: flex; flex-direction: column' }, [
  h('.resize-button.flex-row', [openButton]),
  plotSection,
  infoPanel
]);`

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,84 @@
...
+    ]),
+  ]);
+};
+
+/**
+ * Shows status of current tree with its options (online, loaded, how many)
+ * @param {QCObject} object -  object model that handles state around object
+ * @returns {vnode} - virtual node element
+ */
+export function statusBarLeft(object) {
```
mariscalromeroalejandro, line 89: I think you can refactor the logic a little bit

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,84 @@
...
+  let itemsInfo = '';
+  if (!object.currentList) {
+    itemsInfo = 'Loading objects...';
+  } else if (object.searchInput) {
```
mariscalromeroalejandro: `itemsInfo = object.searchInput
    ? ${object.searchResult.length} found of ${total} items
    : ${total} items;`

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,84 @@
...
+  ]);
+};
+
+/**
+ * Shows status of current tree with its options (online, loaded, how many)
+ * @param {QCObject} object -  object model that handles state around object
+ * @returns {vnode} - virtual node element
+ */
+export function statusBarLeft(object) {
+  let itemsInfo = '';
```
mariscalromeroalejandro:
```suggestion
  let itemsInfo = 'Loading objects...';
```

QualityControl/test/public/pages/object-tree.test.js
```diff
@@ -53,32 +58,51 @@ export const objectTreePageTests = async (url, page, timeout = 5000, testParent)
...
+  });
+
+  await testParent.test('should sort virtual table of objects by name in ascending order', async () => {
     await page.locator(SORTING_BUTTON_PATH).click();
     await page.locator(sortOptionPath(NAME_ASC_INDEX)).click();
     await delay(50); // Wait for sort to finish
 
-    const objectIds = await page.evaluate(() =>
-      [...document.querySelectorAll('[title="qc/test/object"] li')].map((e)=> e.id));
+    const objectIds = await page.evaluate((rowPath) =>
```
mariscalromeroalejandro, line 93: if we are mapping now titles, this should be called better "objectTitles"

mariscalromeroalejandro (changes requested):

QualityControl/public/pages/objectTreeView/component/objectPanel.js
NarrowsProjects: No, not at all. It happened to already be like that. But I'm working here anyway, so I'll change it

QualityControl/public/pages/objectTreeView/component/objectPanel.js
```diff
@@ -0,0 +1,104 @@
...
+ * Method to tackle various states for the selected objects
+ * @param {Model} model - root model of the application
+ * @returns {vnode} - virtual node element
+ */
+export function objectPanel(model) {
+  const { objects, selected } = model.object;
+  const selectedObjectName = selected.name;
+
+  if (objects?.[selectedObjectName]) {
+    return model.object.objects[selectedObjectName].match({
```
mariscalromeroalejandro, line 17:
```suggestion
const selectedObject = objects?.[selectedObjectName];
if (selectedObject) {
  return selectedObject.match({
}
```

mariscalromeroalejandro (changes requested):
well done with this fix, and perfect adding tests! just a small comment added

PR #2878, opened 2025-05-14
[OGUI-1673] Add deployment service and update accordingly existing request handler

graduta:
- [x] PR labels are selected

PR which:
* adds a deployment high-level service to be used for the creation of new environments only
* update the creation payload of deployment to throw if there are no hosts. the errors will trickle up to the controller
* adds a method to retrieve hosts to ignore in workflow services


Control/lib/control-core/CoreUtils.js
```diff
@@ -99,23 +99,22 @@ class CoreUtils {
...
-      });
-      payload.vars.hosts = JSON.stringify(hostsList);
-    } catch (error) {
-      console.error(error)
+      } catch (error) {
+        console.error(error);
+      }
+    });
+    if (hostsList.length === 0) {
+      throw new Error(`No hosts remained after ignoring the ones in Consul`);
```
mariscalromeroalejandro: maybe include the hostsToIgnore for context??

Control/lib/services/WorkflowTemplate.service.js
```diff
@@ -94,12 +98,56 @@ class WorkflowTemplateService {
...
+   * As this should not block data taking operations, the function should not throw an error if the gRPC call fails
+   * or if the JSON is not valid
+   * @param {string} runType - run type for which the hosts should be ignored
+   * @return {Array<string>} - list of hosts to be ignored
+   * @throws
+   */
+  async retrieveHostsToIgnore(runType) {
+    let hostsToIgnoreString = '{}';
+    try {
+      (hostsToIgnoreString = await this._apricotGrpc.getRuntimeEntryByComponent(COG, RUN_TYPE_TO_HOST_MAPPING));
```
mariscalromeroalejandro: parenthesis needed? confusing syntax

Control/lib/services/WorkflowTemplate.service.js
```diff
@@ -94,12 +98,56 @@ class WorkflowTemplateService {
...
+      });
+      this._logger.errorMessage(grpcErrorToNativeError(grpcError));
+      return [];
+    }
+
+    try {
+      const hostsToIgnoreMap = JSON.parse(hostsToIgnoreString);
+      return Array.isArray(hostsToIgnoreMap[runType])
+        ? hostsToIgnoreMap[runType]
+        : [];
```
mariscalromeroalejandro: Avoid accessing twice to the value in hostsToIgnoreMap, also better readability and good for later error logging if we need to inspect it 
```suggestion
      const hostsToIgnore = hostsToIgnoreMap[runType]
      return Array.isArray(hostsToIgnore)
        ? hostsToIgnore
        : [];
```

Control/lib/control-core/CoreUtils.js
```diff
@@ -99,23 +99,22 @@ class CoreUtils {
...
-      const hostsList = JSON.parse(hostsAsString);
-      hostsToIgnore.forEach((knownHost) => {
-        try {
-          const index = hostsList.findIndex((host) => knownHost === host);
-          if (index >= 0) {
-            hostsList.splice(index, 1);
-          }
-        } catch (error) {
-          console.error(error);
+    const hostsAsString = payload?.vars?.hosts ?? '[]';
```
mariscalromeroalejandro, line 106: is it not better to provide the hosts as a normal array of strings and then convert it to a string instead of passing a string with an array of strings??

graduta: Yes, it would but this would need a refactor on the front-end side as well. Currently the GUI loads the configuration and parses it on the client side from the version recieved from ECS. ECS is sending every key-value as a string

Control/lib/services/WorkflowTemplate.service.js
graduta: Indeed, this was an older syntax remainder from when instead of `hostToIgnoreString`, the payload was destructed as a JSON on the same line.

Control/lib/services/WorkflowTemplate.service.js
graduta: Indeed, thank you for the suggestion. I applied it

Control/lib/control-core/CoreUtils.js
graduta: Done

PR #2884, opened 2025-05-20
[OGUI-1676] Fix https protocol usage

mariscalromeroalejandro:
- [X] PR labels are selected

This PR addresses an issue in the utility method responsible for making GET requests to retrieve bookkeeping run types. Previously, the method did not correctly respect the specified protocol, always defaulting to HTTP regardless of the configuration. Fixes done:
* Updated the utility method to detect the protocol (HTTP or HTTPS) based on the configuration or the provided request URL.
* Ensured the GET request uses the correct protocol
* Verified functionality in both staging and production environments to confirm that the correct protocol is being used in all cases.


graduta (changes requested):
Looking at this utility functions, I would be in favour of adding the protocol functionality also to the HEAD request from the same file. 
Moreover, please make sure you are adding tests for these

QualityControl/lib/utils/errorHandler.js
```diff
@@ -0,0 +1,44 @@
...
+ * @returns {void}
+ */
+export function errorHandler(errToLog, errToSend, res, status = 500, facility = 'utils') {
+  errorLogger(errToLog, facility);
+  res.status(status).send({ message: errToSend.message || errToSend });
+}
+
+/**
+ * Global Error Logger for AliECS GUI
+ * @param {Error} err - error that should be logged
```
NarrowsProjects, line 34: This documentation states that the logger is for ALiECS GUI, but it's in the QualityControl project

QualityControl/lib/utils/errorHandler.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Response} res - Response object to send to
+ * @param {number} status - status code 4xx 5xx, 500 will print to debug
+ * @param {string} facility - service that sends the log
+ * @returns {void}
+ */
+export function errorHandler(errToLog, errToSend, res, status = 500, facility = 'utils') {
+  errorLogger(errToLog, facility);
+  res.status(status).send({ message: errToSend.message || errToSend });
+}
+
```
NarrowsProjects, line 31: You're manually sending the status and the message. But the framework has the function:
`updateAndSendExpressResponseFromNativeError` which you could probably use for this instead.

That is, unless there is a functional requirement that I don't know about

NarrowsProjects (changes requested):

NarrowsProjects (dismissed):

QualityControl/lib/utils/httpRequests.js
```diff
@@ -0,0 +1,139 @@
...
+        const body = JSON.parse(chunks.join(''));
+        resolve(body);
+      } catch {
+        reject(new Error('Unable to parse JSON'));
+      }
+    });
+  });
+}
+
+/**
```
graduta, line 93: This doc block can be improved by:
- GH actions suggestions
- for options I would recommend you also add documentation for the specific options you checkout in the method:
e.g.
```
* @param {object} options - Merged HTTP options.
* @param {boolean} options.rejectUnauthorized - Boolean flag is the request should reject unauthorized or not
```

QualityControl/lib/utils/httpRequests.js
```diff
@@ -0,0 +1,139 @@
...
+    ...options ?? {},
+  };
+}
+
+/**
+ * Returns the correct HTTP(S) client based on protocol.
+ * @param {string} protocol - 'http:' or 'https:'
+ * @returns {typeof http | typeof https}
+ */
+function getClient(protocol) {
```
graduta: This is quite a vague method name. Could you please provide a better one? Perhaps `getHttpModule`

graduta (changes requested):
After raising a PR please try to keep an eye over the GH warnings as well. These are there to ensure a certain standard is followed and I think it helps with readability and maintainability of the code.
Moreover, another small and easy to fix comment, for private methods, please name the methods starting with `_`.

QualityControl/test/lib/utils/errorHandler.test.js
```diff
@@ -0,0 +1,62 @@
...
+    beforeEach(() => {
+      res = {
+        status: stub().returnsThis(),
+        send: stub(),
+      };
+    });
+
+    test('should successfully respond with built error message when there is a message and no status', () => {
+      errorHandler('Error', 'Error', res);
+      ok(res.status.calledOnce);
```
graduta, line 32: I guess here you want to ensure that the default status is used when no status is provided.
I think it would be better that you check what status was called with (similar to the other tests).

graduta (approved):

PR #2927, opened 2025-06-11
[OGUI-1700] Add base-app-layout for Configuration GUI

sKrzysieK:
- [x] PR labels are selected


Deaponn (dismissed):

Configuration/package-lock.json
graduta, line 1: This file is not needed, right?

Configuration/webapp/app/test/mocha-index.cjs
graduta, line 1: You are missing the copyright header. Perhaps the linter script is not configured for test files? Nevertheless, it should be added

Configuration/webapp/app/test/public/page-root-mocha.cjs
```diff
@@ -0,0 +1,55 @@
+const assert = require('assert');
+// const test = require('../mocha-index.cjs');
```
graduta: There should not be commented code in the codebase. 
Either is needed, case in which we do not comment it, or is not needed, case in which we remove

Configuration/webapp/app/components/user-section/UserSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import React from 'react';
+import { Box, IconButton, Menu, MenuItem, Avatar } from '@mui/material';
+
+interface UserSectionProps {
+  userName: string;
+}
+
+const UserSection: React.FC<UserSectionProps> = ({ userName }) => {
```
graduta: Normally we keep the export on the same line as the component definition. Is there a specific reason for which you did not do that here?
If not, please use similar to `root.tsx`

graduta: Missing documentation to describe the purpose, goal, parameters, etc. of the component

Configuration/webapp/app/components/user-section/UserSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
graduta: As far as I know, it is best practice to import and use only what is needed. Thus, rather than importing all React, you import only `useState`

Configuration/webapp/app/components/layout/MainLayout.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { Box } from '@mui/material';
+import React, { type FC, type PropsWithChildren } from 'react';
+import LeftDrawer from './drawer/LeftDrawer';
+import Content from './content/Content';
+
+const MainLayout: FC<PropsWithChildren> = ({ children }) => (
```
graduta, line 26: Missing component documentation (params, purpose, goal, return, etc.)

graduta (changes requested):
I believe we discussed on keeping `test` directory outside of the `app` directory.
Moreover, majority, if not all new components are missing documentation on their export definition.

graduta (approved):

PR #2928, opened 2025-06-11
[OGUI-1701] Implement session-service

Deaponn:
- [x] PR labels are selected


sKrzysieK (dismissed):

graduta (changes requested):
While fetching session data in this way, it means that you will have to do this request at the loading of the application and stop any other request from being sent. Thus, delaying the load of the application.

I would recommend you write a fetch wrapper that always check if a token is present in current state (e.g. sessionService.js) and if not, makes the request to server to get it.

If the token is present, then use that one in the query.

You can have a look at the fetchClient.js in the framework which does exactly this for the rest of the GUIs: https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Frontend/js/src/fetchClient.js

graduta (commented):
Another comment that I have is:
- if the token is to be stored in sessionStorage of the browser, this of course it will help for example when the browser page is refreshed.
- but this will also create an issue, that is you are not tackling the case in which the token is invalid and you need to refresh the token.

Instead, if you save the token the state of the app, this will be refreshed automatically at each page entire load

This comment is open for discussion

Configuration/webapp/app/routes/runs/details.tsx
```diff
@@ -21,15 +21,15 @@ import { RunLogs } from '~/components/runs/run-logs';
 import { Spinner } from '~/ui/spinner';
 
 const getRun = async (runNumber: number): Promise<Run> => {
-  const response = await fetch(`http://localhost:8080/api/runs/${runNumber}`);
+  const response = await fetch(`http://localhost:8080/control/api/runs/${runNumber}`);
```
graduta, line 24: It is best to keep only the path and parameters to the request instead of the host and port full path. 
For example, this will work the same
```js
const response = await fetch(`control/api/runs/${runNumber}`);
```

Configuration/webapp/app/routes/runs/details.tsx
```diff
@@ -21,15 +21,15 @@ import { RunLogs } from '~/components/runs/run-logs';
...
   if (!response.ok) {
     throw data(`Failed to fetch run with run number ${runNumber}`, { status: response.status });
   }
   return response.json() as Promise<Run>;
 };
 
 const getLogs = async (runNumber: number): Promise<Log[]> => {
-  const response = await fetch(`http://localhost:8080/api/runs/${runNumber}/logs`);
+  const response = await fetch(`http://localhost:8080/control/api/runs/${runNumber}/logs`);
```
graduta, line 32: Same as above

Configuration/webapp/app/services/session.ts
```diff
@@ -0,0 +1,47 @@
...
+const sessionData: Record<string, string> = {};
+
+/**
+ * Function for fetching session data from the Control server
+ */
+export async function fetchSessionData() {
+  // only to get the data from server redirect
+  // (line 264, commit 3ba4600 of github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/http/server.js)
+  // this should be replaced with endpoint designed for authentication only
+  const response = await fetch('http://localhost:8080/control');
```
graduta, line 24: You should not be using the `host` and `port` in the request as these can change depending on environment. The browser native method `fetch` is smart enough to figure those by itself.
It will work the same with:
```js
const response = await fetch('control');
```

Configuration/webapp/app/root.tsx
```diff
@@ -36,6 +38,10 @@ import MainLayout from './components/layout/MainLayout';
  * @returns {React.ReactElement} Root
  */
 export function Layout({ children }: { children: React.ReactNode }) {
+  useEffect(() => {
```
graduta: What do you think about not allowing the app to load until the token is fetched? In the end if there is no token, no HTTP endpoint will provide the data needed to build the UI

graduta (changes requested):
The session token is loaded as expected with some minor considerations to be made.
Will you also add the wrapper for the `fetch` requests that appends the token automatically in this PR or different one?

Configuration/webapp/app/root.tsx
Deaponn: I was considering implementing the session service as a custom hook which would work like so:
```ts
const { sessionData, isLoading, error } = useSessionData();
```
and then the UI can be built differently based on if the data is still loading or not, or if error is present or not. However I had troubles implementing tests for something like this, but I will work on it

Configuration/webapp/app/services/session.ts
Deaponn: This makes the following error occur while running tests:
```ts
1) `pageRoot` test-suite
       should load session data into memory:
     TypeError: Failed to parse URL from control
      at node:internal/deps/undici/undici:13510:13
      at async fetchSessionData (app/services/session.ts:24:20)
      at async getSessionData (app/services/session.ts:37:5)
```

Deaponn:
The wrapper for the fetch utility (and the refactor from `sessionService` to `useSession` custom hook) will be developed in another PR. I have already created Jira task for that

sKrzysieK (approved):

PR #2929, opened 2025-06-12
[OGUI-1719] Improved Error Message Display with Styled Alert Box

mariscalromeroalejandro:
- [x] PR labels are selected

This PR introduces a new, user-friendly design for error message display in the object viewer. Instead of raw text output, the error now appears inside a styled alert box with visual cues.

* Full-height alert container with soft red (“danger”) background
* Warning icon (⚠️) at the top

### DIFF
--- a/QualityControl/public/app.css
+++ b/QualityControl/public/app.css
@@ -103,23 +103,18 @@
 
 .error-box {
   background-color: #f8d7da;
-  color: #842029;
   height: 100%;
   max-width: 100%;
   word-break: break-word;
   overflow-wrap: break-word;
   border-radius: 5px;
-  border: 1px solid #842029;
+  border: 1px solid var(--color-danger);
 }
 
 .error-icon {
   font-size: 2em;
-  margin-bottom: 0.5em;
-  color: #842029;
+  margin-bottom: .5em;
 }
 
-.error-text {
-  max-width: 100%;
-}
 
 
--- a/QualityControl/public/app.css
+++ b/QualityControl/public/app.css
@@ -115,6 +115,3 @@
   font-size: 2em;
   margin-bottom: .5em;
 }
-
-
-
--- a/QualityControl/public/app.css
+++ b/QualityControl/public/app.css
@@ -100,3 +100,26 @@
   border: 1px solid #ddd;
   background: var(--color-gray-light);
 }
+
+.error-box {
+  background-color: #f8d7da;
+  color: #842029;
+  height: 100%;
+  max-width: 100%;
+  word-break: break-word;
+  overflow-wrap: break-word;
+  border-radius: 5px;
+  border: 1px solid #842029;
+}
+
+.error-icon {
+  font-size: 2em;
+  margin-bottom: 0.5em;
+  color: #842029;
+}
+
+.error-text {
+  max-width: 100%;
+}
+
+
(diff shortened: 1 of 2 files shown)


NarrowsProjects (dismissed):

QualityControl/public/object/objectDraw.js
```diff
@@ -123,9 +124,10 @@ export function draw(model, tabObject, options) {
     // Not asked yet or loading
     return h('.flex-column.items-center.justify-center', [h('.animate-slow-appearance', 'Loading')]);
   } else if (objectRemoteData.isFailure()) {
-    return h('.scroll-y.p1.f6.text-center', {
-      style: 'word-break: break-all;',
-    }, objectRemoteData.payload);
+    return h('.error-box.flex.flex-column.items-center.justify-center.p2.f6.text-center', {}, [
+      h('p.error-icon', { title: 'Error' }, iconWarning()),
```
graduta: We use `span` for wrapping around icons

QualityControl/public/object/objectDraw.js
```diff
@@ -123,9 +124,10 @@ export function draw(model, tabObject, options) {
     // Not asked yet or loading
     return h('.flex-column.items-center.justify-center', [h('.animate-slow-appearance', 'Loading')]);
   } else if (objectRemoteData.isFailure()) {
-    return h('.scroll-y.p1.f6.text-center', {
-      style: 'word-break: break-all;',
-    }, objectRemoteData.payload);
+    return h('.error-box.flex.flex-column.items-center.justify-center.p2.f6.text-center', {}, [
```
graduta: I would be interested to know where does `flex` class coming from?

graduta: There is no need to add `items-center` as you are using flex-column and add `justify-center`

graduta: There is no need for adding `p2` to the div of the box as you use already 100% to fill it

QualityControl/public/app.css
```diff
@@ -100,3 +100,26 @@
...
+  word-break: break-word;
+  overflow-wrap: break-word;
+  border-radius: 5px;
+  border: 1px solid #842029;
+}
+
+.error-icon {
+  font-size: 2em;
+  margin-bottom: 0.5em;
+  color: #842029;
```
graduta: We already have the `danger` class for ALICE color red. Could we not use that perhaps rather than introducing a new color? Or if we do, it would be best to introduce it as a CSS variable so that it can be reused later

QualityControl/public/app.css
```diff
@@ -100,3 +100,26 @@
...
+}
+
+.error-text {
```
graduta: What value does this bring?

QualityControl/public/app.css
```diff
@@ -100,3 +100,26 @@
...
+  max-width: 100%;
+  word-break: break-word;
+  overflow-wrap: break-word;
+  border-radius: 5px;
+  border: 1px solid #842029;
+}
+
+.error-icon {
+  font-size: 2em;
+  margin-bottom: 0.5em;
```
graduta: I think margin is not really needed here, especially when we have plots of size 1x1. This will reduce the space to display the object name
image

versus no margin
image

moreover, the CSS convention is that for values between 0 and 1, we write them as `.5` instead of `0.5`

graduta (changes requested):

QualityControl/public/object/objectDraw.js
mariscalromeroalejandro: I thought it would do `display: flex;`. But I agree `.flex` isn’t doing anything in this context, it can be safely removed

graduta (approved):

NarrowsProjects (approved):

PR #2931, opened 2025-06-14
[OGUI-1688] Update API endpoint to retrieve list of objects based on filter v2

NarrowsProjects:
- [x] PR labels are selected

This PR alters the ObjectController::getObjects function and  and corresponding ObjectService functions to accept filters as an optional parameter. 

The objectService function will decide between the ccdb's tree and latest API based on the presence of the filters 

It also adds middleware for the following endpoints:
/object
/object/:id
/objects

### DIFF
--- a/QualityControl/lib/api.js
+++ b/QualityControl/lib/api.js
@@ -19,7 +19,6 @@ import { layoutOwnerMiddleware } from './middleware/layouts/layoutOwner.middlewa
 import { layoutIdMiddleware } from './middleware/layouts/layoutId.middleware.js';
 import { layoutServiceMiddleware } from './middleware/layouts/layoutService.middleware.js';
 import { statusComponentMiddleware } from './middleware/status/statusComponent.middleware.js';
-// import { getObjectsValidationMiddleware } from './middleware/objects/ObjectGet.middleWare.js';
 
 /**
  * Adds paths and binds websocket to instance of HttpServer passed
--- a/QualityControl/lib/api.js
+++ b/QualityControl/lib/api.js
@@ -19,6 +19,7 @@ import { layoutOwnerMiddleware } from './middleware/layouts/layoutOwner.middlewa
 import { layoutIdMiddleware } from './middleware/layouts/layoutId.middleware.js';
 import { layoutServiceMiddleware } from './middleware/layouts/layoutService.middleware.js';
 import { statusComponentMiddleware } from './middleware/status/statusComponent.middleware.js';
+// import { getObjectsValidationMiddleware } from './middleware/objects/ObjectGet.middleWare.js';
 
 /**
  * Adds paths and binds websocket to instance of HttpServer passed
@@ -46,11 +47,16 @@ export const setup = (http, ws) => {
     layoutRepository,
     jsonFileService,
     filterController,
+    objectGetByIdValidation,
+    objectsGetValidation,
+    objectGetContentsValidation,
   } = setupQcModel();
   statusService.ws = ws;
-  http.get('/object/:id', objectController.getObjectById.bind(objectController));
-  http.get('/object', objectController.getObjectContent.bind(objectController));
-  http.get('/objects', objectController.getObjects.bind(objectController), { public: true });
+
+  http.get('/object/:id', objectGetByIdValidation, objectController.getObjectById.bind(objectController));
+  http.get('/object', objectGetContentsValidation, objectController.getObjectContent.bind(objectController));
+
+  http.get('/objects', objectsGetValidation, objectController.getObjects.bind(objectController), { public: true });
 
   http.get('/layouts', layoutController.getLayoutsHandler.bind(layoutController));
   http.get('/layout/:id', layoutController.getLayoutHandler.bind(layoutController));
--- a/QualityControl/lib/controllers/ObjectController.js
+++ b/QualityControl/lib/controllers/ObjectController.js
@@ -31,7 +31,6 @@ export class ObjectController {
      * @type {QCObjectService}
      */
     this._objService = objService;
-    this._logger = LogManager.getLogger('object/controller');
   }
 
   /**
--- a/QualityControl/lib/controllers/ObjectController.js
+++ b/QualityControl/lib/controllers/ObjectController.js
@@ -44,7 +44,7 @@ export class ObjectController {
     try {
       const { prefix, fields, filters } = req.query;
 
-      const list = await this._objService.retrieveLatestVersionOfObjects(prefix, fields, true, filters);
+      const list = await this._objService.retrieveLatestVersionOfObjects({ prefix, fields, filters });
       res.status(200).json(list);
     } catch (error) {
       const responseError = new Error('Failed to retrieve list of objects latest version');
--- a/QualityControl/lib/controllers/ObjectController.js
+++ b/QualityControl/lib/controllers/ObjectController.js
@@ -78,6 +78,17 @@ export class ObjectController {
     }
   }
 
+  /**
+   * Using `browse` option, request a list of `last-modified` and `valid-from` for a specified path for an object
+   * Use the first `validFrom` option to make a head request to CCDB; Request which will in turn return object
+   * information and download it locally on CCDB if it is not already done so;
+   * From the information retrieved above, use the location with JSROOT to get a JSON object
+   * Use JSROOT to decompress a ROOT object content and convert it to JSON to be sent back to the client for
+   * interpretation with JSROOT.draw
+   * @param {Request} req - HTTP request object with "query" information
+   * @param {Response} res - HTTP response object to provide information on request
+   * @returns {void}
+   */
   async getObjectById(req, res) {
     try {
       const qcgId = req.params?.id;
--- a/QualityControl/lib/controllers/ObjectController.js
+++ b/QualityControl/lib/controllers/ObjectController.js
@@ -48,7 +48,7 @@ export class ObjectController {
     } catch (error) {
       const responseError = new Error('Failed to retrieve list of objects latest version');
 
-      logger.errorMessage(`Error validating query parameters: ${error}`);
+      logger.errorMessage(`Error whilst retrieving objects: ${error}`);
       updateAndSendExpressResponseFromNativeError(res, responseError);
     }
   }
@@ -73,7 +73,7 @@ export class ObjectController {
     } catch (error) {
       const responseError = new Error('Failed to retrieve object content');
 
-      logger.errorMessage(`Error validating query parameters: ${error}`);
+      logger.errorMessage(`Error whilst retrieving object content: ${error}`);
       updateAndSendExpressResponseFromNativeError(res, responseError);
     }
   }
@@ -99,7 +99,7 @@ export class ObjectController {
     } catch (error) {
       const responseError = new Error('Unable to identify object or read it by qcg id');
 
-      logger.errorMessage(`Error validating query parameters: ${error}`);
+      logger.errorMessage(`Error whilst retrieving object: ${error}`);
       updateAndSendExpressResponseFromNativeError(res, responseError);
     }
   }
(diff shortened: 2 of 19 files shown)


QualityControl/lib/controllers/ObjectController.js
```diff
@@ -12,8 +12,9 @@
  * or submit itself to any jurisdiction.
  */
 'use strict';
-import { InvalidInputError, NotFoundError, updateAndSendExpressResponseFromNativeError } from '@aliceo2/web-ui';
-import { LogManager } from '@aliceo2/web-ui';
+import { LogManager, updateAndSendExpressResponseFromNativeError } from '@aliceo2/web-ui';
+
+const logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'qcg'}/object-ctrl`);
```
mariscalromeroalejandro, line 17: purpose of having 2 loggers?

QualityControl/lib/services/QcObject.service.js
```diff
@@ -90,18 +90,31 @@ export class QcObjectService {
...
    * @rejects {Error}
    */
-  async retrieveLatestVersionOfObjects(prefix = this._dbService.PREFIX, useCache = true) {
-    if (useCache && this._cache?.objects) {
+  async retrieveLatestVersionOfObjects(
+    prefix = this._dbService.PREFIX,
+    fields,
+    useCache = true,
+    filters = undefined,
+  ) {
```
mariscalromeroalejandro: consider destructuring an options object instead of using a long parameter list

QualityControl/lib/services/QcObject.service.js
```diff
@@ -90,18 +90,31 @@ export class QcObjectService {
...
+    const hasFilters = filters !== undefined;
```
mariscalromeroalejandro:
```suggestion
    const hasFilters = filters !== undefined && Object.keys(filters).length > 0;
```

QualityControl/lib/services/QcObject.service.js
```diff
@@ -90,18 +90,31 @@ export class QcObjectService {
...
    * @param {boolean} [useCache = true] - if the list should be the cached version or not
+   * @param {Array<string>} [filters = undefined] - Run number by which the objects are filtered.
    * @returns {Promise.<Array<QcObjectLeaf>>} - results of objects with required fields
    * @rejects {Error}
    */
-  async retrieveLatestVersionOfObjects(prefix = this._dbService.PREFIX, useCache = true) {
-    if (useCache && this._cache?.objects) {
+  async retrieveLatestVersionOfObjects(
+    prefix = this._dbService.PREFIX,
+    fields,
```
mariscalromeroalejandro:
```suggestion
    fields = [],
```

QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -126,18 +125,24 @@ export class CcdbService {
...
-    const headers = {
-      accept: 'application/json',
-      'x-filter-fields': fields.length > 0 ? fields.join(',') : `${PATH},${CREATED},${LAST_MODIFIED}`,
-    };
-    const { objects } = await httpGetJson(this._hostname, this._port, `/latest/${prefix}.*`, { headers });
+  async getObjectsLatestVersionList(
+    prefix = this._PREFIX,
+    filters,
+    fields,
+  ) {
```
mariscalromeroalejandro: Consider options object:

 ```suggestion
  async getObjectsLatestVersionList({ prefix = this._PREFIX, filters = {}, fields = [PATH, CREATED, LAST_MODIFIED] } = {}) {
```

**Review by mariscalromeroalejandro — CHANGES_REQUESTED** [review]

File: QualityControl/lib/services/QcObject.service.js (hunk 4 above)
Comment (NarrowsProjects) [reply QualityControl/lib/services/QcObject.service.js]: This is dealt with by the dto's which set fields to [] by default.

File: QualityControl/lib/services/ccdb/CcdbService.js (hunk 5 above)
Comment (NarrowsProjects) [reply QualityControl/lib/services/ccdb/CcdbService.js]: For only three arguments?

Comment (mariscalromeroalejandro) [reply QualityControl/lib/services/ccdb/CcdbService.js]: yes, filters and fields can be mixed in order by mistake

**Review by mariscalromeroalejandro — DISMISSED** [review]

File: QualityControl/lib/api.js [hunk 6]
```diff
@@ -19,6 +19,7 @@ import { layoutOwnerMiddleware } from './middleware/layouts/layoutOwner.middlewa
 import { layoutIdMiddleware } from './middleware/layouts/layoutId.middleware.js';
 import { layoutServiceMiddleware } from './middleware/layouts/layoutService.middleware.js';
 import { statusComponentMiddleware } from './middleware/status/statusComponent.middleware.js';
+// import { getObjectsValidationMiddleware } from './middleware/objects/ObjectGet.middleWare.js';
```
graduta: Do we need code that is commented out?

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -79,57 +65,31 @@ export class ObjectController {
...
+      res.status(200).json(object);
+    } catch (error) {
+      const responseError = new Error('Failed to retrieve object content');
+
+      logger.errorMessage(`Error validating query parameters: ${error}`);
+      updateAndSendExpressResponseFromNativeError(res, responseError);
     }
   }
 
-  /**
```
graduta: Is there a reason for which the documentation of the method was removed entirely?

QualityControl/lib/middleware/objects/objectGetByIdValidationMiddlewareFactory.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {FilterService} filterService - Service providing run types
+ * @returns {function(req, res, next): undefined} - Express middleware function
+ */
+export function objectGetByIdValidationMiddlewareFactory(filterService) {
+  const { runTypes } = filterService;
+  const ObjectGetByIdDto = createObjectGetByIdDto({ runTypes });
+
+  return async (req, res, next) => {
+    try {
+      req.params.id = await qcgIdDto.validateAsync(req.params?.id);
```
graduta: Why is this Dto referring to qcg (which is the name of the application) instead of to the actual entity? In this case, is it `an object`, is it `a layout`? 


As far as I can tell from the qcgIdDto implementation, this is a general string check that has min 1 character. Thus, it should be placed in a separate file and named accordingly

QualityControl/lib/middleware/objects/objectGetByIdValidationMiddlewareFactory.js
```diff
@@ -0,0 +1,41 @@
...
+      req.query = await ObjectGetByIdDto.validateAsync(req.query);
+      return next();
```
graduta, line 33: You do not `return next()` but simply call the next method

QualityControl/lib/controllers/ObjectController.js
graduta, line 17: Logger was on purpose added to the constructor and not outside of it. Please revert the change

QualityControl/lib/services/QcObject.service.js
```diff
@@ -89,19 +89,28 @@ export class QcObjectService {
...
    * * make a new request and get data directly from data service
    * * @example Equivalent of URL request: `/latest/qc/TPC/object.*`
-   * @param {string|Regex} prefix - Prefix for which CCDB should search for objects.
-   * @param {Array<string>} [fields = []] - List of fields that should be requested for each object
-   * @param {boolean} [useCache = true] - if the list should be the cached version or not
+   * @param {object} options - An object that contains query parameters among other arguments
+   * @param {string|Regex} options.prefix - Prefix for which CCDB should search for objects.
+   * @param {Array<string>} options.fields - List of fields that should be requested for each object
+   * @param {boolean} options.useCache - if the list should be the cached version or not
+   * @param {Array<string>} options.filters - Run number by which the objects are filtered.
```
graduta: It is not only RunNumber, is it?

graduta (changes requested):

QualityControl/lib/api.js
NarrowsProjects: Indeed, we do not

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -79,57 +65,31 @@ export class ObjectController {
...
-      }
+    try {
+      const { path, validFrom, filters, id } = req.query;
+
+      const object = await this._objService.retrieveQcObject(path, validFrom, id, filters);
+      res.status(200).json(object);
+    } catch (error) {
+      const responseError = new Error('Failed to retrieve object content');
+
+      logger.errorMessage(`Error validating query parameters: ${error}`);
```
graduta: There is no error to validate the query here. The input parameters are evaluated in the middleware now

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -79,57 +65,31 @@ export class ObjectController {
...
+    try {
+      const qcgId = req.params?.id;
+      const { validFrom, filters, id } = req.query;
+
+      const object = await this._objService.retrieveQcObjectByQcgId(qcgId, id, validFrom, filters);
+      res.status(200).json(object);
+    } catch (error) {
+      const responseError = new Error('Unable to identify object or read it by qcg id');
+
+      logger.errorMessage(`Error validating query parameters: ${error}`);
```
graduta: The same here, input is validated in the middleware, thus this is a different kind of error

QualityControl/lib/controllers/ObjectController.js
NarrowsProjects: No, there is not

QualityControl/lib/middleware/objects/objectGetByIdValidationMiddlewareFactory.js
NarrowsProjects: In this specific situation, return next() doesn't do anything. 

But, it is common practice/convention to do so, because next() doesn't prevent all code that comes after it from running.

QualityControl/lib/controllers/ObjectController.js
NarrowsProjects: It's no difficult change, but I would like to mention, that the other controllers that have a logger (UserController and LayoutController) initialise their loggers outside of the constructor.

The only classes that have a logger in their constructor are SequaliseDatabase, and QCObjectService.

Do you still want me to put it in the constructor?

QualityControl/lib/middleware/objects/objectGetByIdValidationMiddlewareFactory.js
NarrowsProjects: This is a qcObjectId (I think) (refer to QCObject.service.js::109). So, It does belong in that file, it should just be named differently.

QualityControl/lib/middleware/objects/objectGetByIdValidationMiddlewareFactory.js
graduta: Indeed, next will not return and continue execution. In this case, there is nothing to execute after though

graduta: So, let's keep it as you suggested just so that we avoid potential issues in the future

QualityControl/lib/controllers/ObjectController.js
graduta: Basically, this gets called when the file is imported. In the previous version of the Logger, this was creating a problem as it would print the logger before the class was ever used.

If we put it in the class, it will only get instantiated when the class is instantiated.

But with the this LogManager, we do not have the problem anymore, so feel free to leave it as it is

graduta (approved):

PR #2932, opened 2025-06-14
[OGUI-1689] Implement object tree filters based on metadata v2

NarrowsProjects:
- [x] PR labels are selected

This PR depends on https://github.com/AliceO2Group/WebUi/pull/2931

This PR adds a metadata filter to the top of the objectTreePage that will filter the objectTree for matching objects.
It also filters the object versions in the sidePanel where relevant.

When the filter is applied, the objectInfoPanel closes


QualityControl/public/common/filters/model/filterModel.js
graduta, line 1: Being a class, the filter model should be named with the convention of capital starting letter `FilterModel.js`

mariscalromeroalejandro (dismissed):

QualityControl/public/common/filters/filterViews.js
```diff
@@ -0,0 +1,79 @@
...
+    ],
+  );
+};
+
+/**
+ * Button which will allow the user to update filter parameters after the input
+ * @param {Function} triggerFilter - Function to trigger the filter mechanism
+ * @returns {vnode} - virtual node element
+ */
+const triggerFiltersButton = (triggerFilter) => h('button.btn.btn-primary#triggerFilterButton', {
```
graduta: I would recommend that the name is `onClickHandler/onClickCallback` instead of `triggerFilter` or simply `onclick`

QualityControl/public/common/filters/filterViews.js
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { filters } from '../../../common/filters/filter.js';
```
graduta: I think this import is incorrect. Is it not actually `../../common/filters/filter.js`? Which if you are looking at it is actually just `./filter.js`?

QualityControl/public/common/filters/filterViews.js
```diff
@@ -0,0 +1,79 @@
...
+import { FilterType } from '../../../common/filters/filterTypes.js';
```
graduta: Ditto

QualityControl/public/common/filters/filterViews.js
```diff
@@ -0,0 +1,79 @@
...
+ * Creates an input element for a specific metadata field;
+ * @param {object} config - The configuration for this particular field
+ * @param {object} filterMap - An object that contains the keys and values of the filters
+ * @param {Function} onInputCallback - A callback function that triggers upon Input
+ * @param {Function} onEnterCallback - A callback function that triggers upon Enter
+ * @param {Function} onChangeCallback - A callback function that triggers upon Change
+ * @returns {undefined}
+ */
+const createFilterElement = (config, filterMap, onInputCallback, onEnterCallback, onChangeCallback) => {
+  const { filterInput, dynamicSelector } = filters;
```
graduta: Also, this confused me, you should simply import the desired one from the import statement. 
```js
import { filterInput, dynamicSelector } from './filter.js';
```

Of course with the corresponding change in `filter.js` file for export
```js
export {
  filterInput,
  dynamicSelector,
};
```

QualityControl/public/common/filters/filterViews.js
```diff
@@ -0,0 +1,79 @@
...
+  );
+};
+
+/**
+ * Button which will allow the user to update filter parameters after the input
+ * @param {Function} triggerFilter - Function to trigger the filter mechanism
+ * @returns {vnode} - virtual node element
+ */
+const triggerFiltersButton = (triggerFilter) => h('button.btn.btn-primary#triggerFilterButton', {
+  onclick: triggerFilter,
```
graduta: I think from accessibility point of view it is important to add a `title` to buttons.

graduta (changes requested):
I think because you created `filterViews.js` from scratch rather than renaming the `public/layout/view/panels/filters.js`, you forgot to delete the second file from the codebase which now is dead code.

I see also an important change of behaviours of the filters with this change:
- right now (in `dev`) when the filter gets a new run number or a new selection from the dropdown of `runTypes`, the objects are refreshed
- in your change, this does not happen anymore
- is it on purpose? (I am not saying it is wrong, it will protect our resources so that users do not simply select by mistake and query 20-30 objects for that). I am asking because of the following scenario:

* user inputs RunNumber 2 but decides not to click on `Update` to fetch the objects but instead the user enlarges an object (clicks button to see object full page). This in turn will try to fetch the object with RunNumber 2. So, it might create a bit of confusion here

graduta (approved):

PR #2933, opened 2025-06-15
[OGUI-1690] Persist filters between pages v2

NarrowsProjects:
- [x] PR labels are selected

This PR depends on https://github.com/AliceO2Group/WebUi/pull/2932

This PR makes the run number filter, as well as every other filter persist by adding it in the URL. 

This is achieved by permanently adding the filter element to the header. And using the filter in QCObject.service and Layout.service API requests.


QualityControl/public/common/filters/model/FilterModel.js
```diff
@@ -31,6 +31,7 @@ export default class FilterModel extends Observable {
     this.model = model;
     this.filterService = new FilterService(this);
     this._filterMap = {};
+    this.visible = true;
```
graduta: Normally, for boolean values the convention for naming is something like `is/has`
`isVisible` - in this cse

QualityControl/public/common/filters/filterViews.js
```diff
@@ -51,16 +51,22 @@ const createFilterElement = (config, filterMap, onInputCallback, onEnterCallback
...
+  if (!visible) {
+    return null;
+  }
+
   return h(
-    '.w-100.flex-row.p2.g2',
+    '.w-100.flex-row.p2.g2.justify-center#filterElement',
     [
       triggerFiltersButton(onEnterCallback),
+      clearFiltersButton(clearFilterCallback),
```
graduta, line 69: Clicking clear, does not remove value from input of RunNumber or from dropdown unless another action is performed.
e.g. 
- set RunNumber to 42
- set RunType to something 
- click Clear --- nothing happens
- select a different RunType and at this point RunNumber is reset

QualityControl/public/layout/Layout.js
```diff
@@ -61,7 +61,7 @@ export default class Layout extends Observable {
     this.cellHeight = 100 / this.gridListSize * 0.95; // %, put some margin at bottom to see below
     this.cellWidth = 100 / this.gridListSize; // %
     // GridList.grid.length: integer, number of rows
-    this.filterModel = model.filterModel;
+    this.applyTreeFilter = true;
```
graduta: Same for the naming, here something like `shouldApplyTreeFilter` improves readability

QualityControl/public/layout/Layout.js
```diff
@@ -754,4 +769,23 @@ export default class Layout extends Observable {
...
+      this.listObjects();
+    }
+  }
+
+  /**
+   * Wrapper function for qcObjectServices. If applyTreeFilter is false,
+   * the objectTree will be loaded without applying the filters.
+   */
+  listObjects() {
+    const filterMap = this.applyTreeFilter ? undefined : {};
```
graduta: Are you sure this is correct?

QualityControl/public/layout/view/panels/settings.js
```diff
@@ -31,6 +31,7 @@ export default (model) =>
       (e)=> model.layout.setLayoutProperty('description', e.target.value),
     ),
     displayObjectTime(model),
+    applyFilters(model.layout),
```
graduta: I would recommend this to be last so that is as close as possible to the sideTree. 
I have to admit I did not see the option when testing and I only found out about it during codee review

QualityControl/public/pages/layoutListView/model/LayoutListModel.js
```diff
@@ -115,4 +115,9 @@ export default class LayoutListModel extends Observable {
       },
     };
   }
+
+  /**
+   * This function is required for all pages that have the filter element in their header
+   */
+  triggerFilter() {}
```
graduta: Instead of adding empty methods, you could make use of the `extends` property of classes.
Something like
```js
export default class BaseViewModel extends Observable {
  /**
   * Default triggerFilter implementation to be overwritten by the subclasses
   */
  triggerFilter() {
  }
}
```

Use:
```js
export default class AboutViewModel extends BaseViewModel {
 [...] existing code without the empty triggerFilter
}
```

graduta (changes requested):
I believe, the "Clear" button should also fetch the new objects

QualityControl/public/layout/Layout.js
NarrowsProjects: Yes, if model.services.object.listObjects is provided undefined, it will send a request with the active filters,
If it is provided (undefined, {}) it will use the empty object instead of filter.

graduta (changes requested):
I have identified a bug which is introduced by this PR. To reproduce:

- Go to a layout with objects
- Make sure you have at least an object that loads correctly
- Click on the expand icon to see the object in objectView page full screen
- At this point, you see in the URL parameters objectId and layoutID
- select a different version timestamp of the object
- objectId and layoutID were replaced by objectName. This means that also the "Back to layout" button has been replaced by "BacK"
- At this point cannot get back to the previous layout that was at

graduta (approved):

PR #2940, opened 2025-06-17
[OGUI-1703] Implement wrapper interfaces matching basic tokenization diagram

OmegaCreations:
- [x] PR labels are selected

This PR implements basic typescript interfaces and enums matching **wrapper.proto** definitions. Due to no typescript naming conventions inside Bookkeeping, I keep the same approach as the .proto files.


Tokenization/backend/wrapper/models/message.model.ts
```diff
@@ -0,0 +1,54 @@
...
+//                 ENUMS
+// ======================================
+
+/**
+ * @enum Represents the types of events that can occur in a duplex message exchange.
+ * @property MESSAGE_EVENT_EMPTY: No event, used for initialization or no response.
+ * @property MESSAGE_EVENT_NEW_TOKEN: Event for replacing with newly generated token.
+ * @property MESSAGE_EVENT_REVOKE_TOKEN: Event for revoking an existing token.
+ */
+enum DuplexMessageEvent {
```
graduta, line 25: Bear in mind that in `proto` enums are marked as numbers.
I would advise you do the same in TS to keep the mapping consistent

Moreover, as the object is already called MessageEvent, how about you remove that from the key?
Perhaps something like
```ts
enum DuplexMessageEvent {
  UNKNOWN = 0,
  NEW_TOKEN = 1,
  REVOKE_TOKEN = 2
}
```
Of course with similar changes on the proto side of MessageEvent

OmegaCreations: Done. Also about naming convention in grpc. It is highly recommended to use enum data with prefixes based on enum name. Besides being a convention it is good to use it for static languages like c++/java because there might be conflicts between namespaces - bearing in mind it is due to human error I would stay with these prefixes as it is good to use for bigger projects and we already keep this convention in e.g. Bokkeeping

graduta (approved):

PR #2941, opened 2025-06-17
[OGUI-1704] Create Serialization Utilities

OmegaCreations:
- [x] PR labels are selected

This PR implements basic **serialization/deserialization** methods for `application/json` data.


graduta (commented):
Given that we are using gRPC, do we need serialization utility functions? Or could we use the standards proto buffers?

OmegaCreations:
> Given that we are using gRPC, do we need serialization utility functions? Or could we use the standards proto buffers?

@graduta Besides that it is not needed in the scope of current User Stories, I belive that if we want to send different requests via wrapper it would "might?" be helpful to serialize requests.

Of course if we want we can modify `.proto` file to send http requests like:

```
enum HttpMethod {
  UNSPECIFIED = 0;
  GET = 1;
  POST = 2;
  PUT = 3;
  PATCH = 4;
  DELETE = 5;
  HEAD = 6;
  OPTIONS = 7;
}

message HttpHeader {
  string key = 1;
  repeated string values = 2;
}

message HttpRequest {
  string url = 1;
  HttpMethod method = 2;
  repeated HttpHeader headers = 3;
  bytes body = 4; // data
  string content_type = 5; // "application/json"
}

message HttpResponse {
  int32 status = 1;
  repeated HttpHeader headers = 2;
  bytes body = 3;
  string content_type = 4;
}
```

instead of something like:

```
message HttpMessage {
  bytes payload = 1;
}
```

which will be just serialized json payload. Please let me know what you prefer. If we want more clear structure we can reject this PR and in the new User Stories I will add sending data with more precised http structures on gRPC end. 

I guess structurizing requests and repsonses would be more "compliant" with how proto files are intended to be written and there won't be any double framing/parsing with information about payload length/structure because gRPC is doing it also for every message. It also might be more helpful for other eventual developers to know what is happening by just reading .proto files.

OmegaCreations:
After more thinking, I believe you're right and with standard proto buffers code and logic will be clearer and it would be more optimal to write it this way.

PR #2943, opened 2025-06-17
[OGUI-1698] Implement configuration endpoint

piechnikk:
- [x] PR labels are selected


Control/lib/controllers/QCConfiguration.controller.js
```diff
@@ -0,0 +1,88 @@
...
+   * @param {Response} res
+   */
+  async getConfigurationByKey(req, res) {
+    const { key } = req.query;
+    if (!key) {
+      updateAndSendExpressResponseFromNativeError(res, new InvalidInputError("Missing configuration key"));
+    }
+
+    try {
+      const value = await this._qcConfigurationService.getConfigurationByKey(key);
```
Deaponn: Endpoint GET `/api/configurations` requires only the prefix of the files, but this endpoint requires the whole path to the file. Maybe consider making it more consistent? For example, in GET `/api/configurations`, make it so the default `prefix` stays like now, but when `prefix` query parameter is defined, then use the whole value, don't build it using string interpolation (or in GET `/api/configuration` make it so only filename needs to be present without the whole `_qcConfigurationsPath` part, and build the `key` using string interpolation)

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Initialize Lock service based on the provided list of detectors
+   * @param {Array<String>} detectors = [] - list of detectors to be used for the lock mechanism
+   * @return {void}
+   */
+  // setLockStatesForDetectors(detectors = []) {
+  //   for (const detectorName of detectors) {
+  //     this._locksByDetector[detectorName] = new DetectorLock(detectorName);
+  //   }
+  // }
```
Deaponn: Unused code

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,125 @@
...
+      const parsedData = [];
+
+      Object.entries(data || {}).forEach(([key, value]) => {
+        try {
+          if (!JSON.parse(value)) {
+            return;
+          }
+          if (!recurse && key.replace(`${prefix}/`, "").includes("/")) {
+            return;
+          }
```
Deaponn: Consider changing the order of this two `if` statements from lines 86 and 89, since the first one is more likely to be slower?

Deaponn (commented):
Left some ideas

Control/lib/api.js
```diff
@@ -90,6 +92,10 @@ module.exports.setup = (http, ws) => {
   const broadcastService = new BroadcastService(ws);
   const cacheService = new CacheService(broadcastService);
   const environmentCacheService = new EnvironmentCacheService(broadcastService, eventEmitter);
+  const qcConfigurationService = new QCConfigurationService(consulService);
+  qcConfigurationService.testConsulStatus();
```
graduta: There is not much point in having this check here for the service as a few lines below, (L101), the consul controller is doing the exact same thing. 
I would say that ideally, this should be done (via the consul service itself and not high-services) in the `initialize data` but that is not part of this ticket. 
I would recommend you remove this call and the method itself from the service

Control/lib/api.js
```diff
@@ -230,6 +236,17 @@ module.exports.setup = (http, ws) => {
     statusController.getAliECSIntegratedServicesStatus.bind(statusController),
   );
 
+  // Configuration
+  const qcValidateService = qcConfigurationService.validateService.bind(qcConfigurationService);
```
graduta: I would recommend you extract this in an independent middleware method such as `getDetectorsLockOwnershipMiddlewareFactory` and then use it from there as right now you are duplicating code by adding the same lines of code that we could find in the consul controller

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Check if consulService is present:
+   * * If yes, allow request to continue
+   * * If not, send response accordingly
+   * @param {Request} req
+   * @param {Response} res
+   * @param {Next} next
+   */
+  validateService(req, res, next) {
```
graduta: See my comments from `api.js` but basically this should be added as a middleware function and reused both in the API for qc configuration and the API for consul controller

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+      next();
+    } else {
+      errorHandler("Unable to retrieve configuration of consul service", res, 502);
+    }
+  }
+
+  /**
+   * Method to check if consul service can be used
+   */
+  async testConsulStatus() {
```
graduta: This method is simply copied from Consul.controller.js and we should not do that.
Instead, I would recomment you create a utility function that deals with this check and call it only once in `initializeData`.
Or even better have a function `checkServices`

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+      const data = await this._consulService.getOnlyRawValuesByKeyPrefix(prefix);
+      const parsedData = [];
+
+      Object.entries(data || {}).forEach(([key, value]) => {
+        try {
+          if (!recurse && key.replace(`${prefix}/`, "").includes("/")) {
+            return;
+          }
+          if (!JSON.parse(value)) {
+            return;
```
graduta: I believe you should also add tests for empty cases

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+          parsedData.push(key);
+        } catch (e) {
+          // skip
+        }
+      });
+
+      return parsedData;
+    } catch (error) {
+      if (error instanceof NotFoundError) {
+        throw new NotFoundError(`No configurations found for prefix: ${prefix}`);
```
graduta: It is important that you add tests for failing scenarios as well

graduta: The controller should be the one parsing the errors as it also benefits of using the method `updateAndSendExpressResponseFromNativeError`
Is there a specific reason you wanted here to parse the error and modify it?

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Get configuration by key from Consul
+   * @param {String} key - the key of the configuration
+   */
+  async getConfigurationByKey(key) {
+    try {
+      const value = await this._consulService.getOnlyRawValueByKey(key);
+      if (!value) {
+        throw new NotFoundError(`Configuration not found for key: ${key}`);
```
graduta: You should be covering this scenario be tests as well

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+      }
+      return value;
+    } catch (error) {
+      throw new ServiceUnavailableError(`Error retrieving configuration for key: ${key}`);
```
graduta: This should be caught by the controller and dealt with.

Control/lib/controllers/QCConfiguration.controller.js
```diff
@@ -0,0 +1,88 @@
...
+const { getConsulConfig } = require("../config/publicConfigProvider.js");
+
+/**
+ * Gateway for all Consul Consumer calls
+ */
+class QCConfigurationController {
+  /**
+   * Setup QCConfigurationController
+   * @param {QCConfigurationService} qcConfigurationService
+   * @param {JSON} config
```
graduta: Missing description of expected parameter format

Control/lib/controllers/QCConfiguration.controller.js
```diff
@@ -0,0 +1,88 @@
...
+    try {
+      const parsedData = await this._qcConfigurationService.getKeysOfValidConfigurations(prefixPath, recurse);
+      if (!parsedData || parsedData.length === 0) {
+        updateAndSendExpressResponseFromNativeError(res, new NotFoundError("No configurations found"));
+      }
+
+      res.status(200).json(parsedData);
+    } catch (error) {
+      errorLogger(error, this._logger);
+      updateAndSendExpressResponseFromNativeError(res, error);
```
graduta: You should cover failing scenarios by tests

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,114 @@
...
+   * Get keys of configurations stored in Consul
+   * @param {String} prefix - prefix to filter the keys
+   * @param {boolean} [recurse=false] - whether to recurse into subdirectories
+   */
+  async getKeysOfValidConfigurations(prefix, recurse = false) {
+    try {
+      const data = await this._consulService.getOnlyRawValuesByKeyPrefix(prefix);
+      const parsedData = [];
+
+      Object.entries(data || {}).forEach(([key, value]) => {
```
graduta: It would be best to extract this filtering and parsing in its own utility function so that you can properly add unit tests to it

graduta (changes requested):

Control/lib/middleware/validateConsulServiceMiddlewareFactory.js
```diff
@@ -0,0 +1,37 @@
...
+   * @param {Request} req - HTTP Request object
+   * @param {Response} res - HTTP Response object
+   * @param {Next} next - HTTP Next object to use if checks pass
+   * @returns {void} continue if checks pass, uses response object to respond with error if checks fail
+   */
+  return async (req, res, next) => {
+    if (consulService) {
+      next();
+    } else {
+      res.status(502).json({ message: "Unable to retrieve configuration of consul service" });
```
graduta: I would recommend you use the `updateAndSendExpressResponseFromNativeError` with a `ServiceUnavailableError` instead

Control/lib/api.js
```diff
@@ -300,8 +325,21 @@ function initializeIntervals(intervalsService, statusService, runService, bkpSer
...
   await apricotService.init();
   lockService.setLockStatesForDetectors(apricotService.detectors);
+  await testConsulStatus(consulService);
 }
+
+/**
+ * Method to check if consul service can be used
+ * @param {ConsulService} consulService
+ */
+async function testConsulStatus(consulService) {
```
graduta: There is no point in using `async` in here as the service itself is using the `then/catch`

Control/lib/api.js
```diff
@@ -300,8 +325,21 @@ function initializeIntervals(intervalsService, statusService, runService, bkpSer
  * Function to initialize in order dependent services
  * @param {ApricotService} apricotService - request initial set of data from AliECS/Apricot
  * @param {LockService} lockService - initialize service with data from Apricot
+ * @param {ConsulService} consulService - service for communicating with Consul
  */
-async function initializeData(apricotService, lockService) {
+async function initializeData(apricotService, lockService, consulService) {
   await apricotService.init();
   lockService.setLockStatesForDetectors(apricotService.detectors);
+  await testConsulStatus(consulService);
```
graduta: No need to await here as soon as you remove the `async` on L340

Control/lib/api.js
```diff
@@ -300,8 +325,21 @@ function initializeIntervals(intervalsService, statusService, runService, bkpSer
...
+/**
+ * Method to check if consul service can be used
+ * @param {ConsulService} consulService
+ */
+async function testConsulStatus(consulService) {
+  consulService
+    .getConsulLeaderStatus()
+    .then((data) => logger.info(`Service is up and running on: ${data}`))
+    .catch((error) => logger.error(`Connection failed due to ${error}`));
+}
```
graduta: Missing end of file newline

Control/test/config/testConfigForConsul.js
```diff
@@ -0,0 +1,52 @@
...
+  
+  nock(CONSUL_URL)
+    .persist()
+    .get(`${KV_PATH}key1?raw=true`)
+    .reply(200, JSON.stringify({key: "value"}))
+}
+
+module.exports = {
+  initializeNockForConsul
+};
```
graduta: Missing end of file newline

Control/test/api/configuration/api-get-configuration.test.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+const request = require('supertest');
+const { ADMIN_TEST_TOKEN, TEST_URL } = require('../generateToken.js');
+
+describe(`'API - GET - /configuration' test suite`, () => {
```
graduta: In the controller, I see that you also return NotFound if key is missing. https://github.com/AliceO2Group/WebUi/pull/2943/files#diff-2b9ec462d6f2ebfebb0b7afc11a0cd164e848b6f02ffb553519931f728c70e71R74 
I would recommend that you also cover this case by tets

Control/lib/controllers/QCConfiguration.controller.js
```diff
@@ -0,0 +1,88 @@
...
+
+    this._logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? "cnf"}/qc-configuration-controller`);
+  }
+
+  /**
+   * Method to get configurations names
+   * @param {Request} req - HTTP Request object
+   * @param {Response} res - HTTP Response object
+   */
+  async getConfigurationsKeys(req, res) {
```
graduta: Please keep in mind the project conventions. Unfortunately, they may be lacking in some places which can create confusion but I documented them here and will try to keep on adding to them:
https://github.com/AliceO2Group/WebUi/blob/dev/documentation/CONVENTIONS.md 

In this case `controller` methods should end with `Handler` in the naming

Control/test/api/configuration/api-get-configurations.test.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+*/
+
+const request = require('supertest');
+const { ADMIN_TEST_TOKEN, TEST_URL } = require('../generateToken.js');
+
+describe(`'API - GET - /configurations' test suite`, () => {
```
graduta, line 20: Looking in the controller handler, I see sometime you may throw NotFoundError
It would be nice to also have this covered by the API tests

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,95 @@
...
+   * Get keys of configurations stored in Consul
+   * @param {String} prefix - prefix to filter the keys
+   * @param {boolean} [recurse=false] - whether to recurse into subdirectories
+   */
+  async getKeysOfValidConfigurations(prefix, recurse = false) {
+    let data;
+
+    try {
+      data = await this._consulService.getOnlyRawValuesByKeyPrefix(prefix);
+    } catch (e) {
```
graduta: I believe your controller should be the one catching the error and not try to deal with it here.
Normally the service function has one goal only, in this case to get the parsed keys to the parent caller

Control/lib/services/QCConfiguration.service.js
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Get configuration by key from Consul
+   * @param {string} key - the key of the configuration
+   */
+  async getConfigurationByKey(key) {
+    try{
+      return await this._consulService.getOnlyRawValueByKey(key);
+    }catch (error) {
+      this._logger.error(`Error getting configuration by key: ${key}`, error);
+      throw new NotFoundError(`Configuration not found for key: ${key}`);
```
graduta: Same applies here, you are assuming a NotFoundError but it can be service unavailable, key non-existent, etc.
I would recommend you let the controller handle this and let this service method simply throw error

graduta (changes requested):
Very good progress! I added a few small comments to make sure the new features are in-line with our conventions.

graduta (approved):

PR #2952, opened 2025-06-23
[OGUI-1729] Implement restrictions endpoint

Deaponn:
- [x] PR labels are selected


Control/lib/controllers/QCConfiguration.controller.js
```diff
@@ -94,10 +96,37 @@ class QCConfigurationController {
     }
   }
 
+  /**
+   * Method to get configuration restrictions by key
+   * @param {Request} req
+   * @param {Response} res
```
sKrzysieK: Please add a description for parameters
```
* @param {Request} req - HTTP Request object
* @param {Response} res - HTTP Response object
```

Control/lib/controllers/QCConfiguration.controller.js
```diff
@@ -94,10 +96,37 @@ class QCConfigurationController {
...
+      } else {
+        updateAndSendExpressResponseFromNativeError(res, new ServiceUnavailableError('Consul service unavailable'));
+      }
+    }
+  }
+
   /**
    * Method to edit configuration value
    * @param {Request} req
    * @param {Response} res
```
sKrzysieK: same here

Control/test/lib/controllers/mocha-qcConfiguration.controller.test.js
```diff
@@ -115,6 +121,60 @@ describe(`'QCConfigurationController' test suite`, () => {
...
+      req.params.key = undefined;
+      
+      await qcConfigurationController.getConfigurationRestrictionsByKeyHandler(req, res);
+      
+      assert.ok(statusStub.calledWith(400));
+      assert.deepStrictEqual(jsonStub.firstCall.args[0].message, 'Missing configuration key');
+    });
+
+    it('should return 400 if key is an empty string', async () => {
+      req.params.key = '  ';
```
sKrzysieK, line 148: I am a little bit confused here - test description is `should return 400 if key is an empty string` but I think that we test the case when specific key does not exist. Consider adjusting test description.

Control/test/lib/controllers/mocha-qcConfiguration.controller.test.js
```diff
@@ -115,6 +121,60 @@ describe(`'QCConfigurationController' test suite`, () => {
...
+        
+      await qcConfigurationController.getConfigurationRestrictionsByKeyHandler(req, res);
+        
+      assert.ok(statusStub.calledWith(400));
+      assert.deepStrictEqual(jsonStub.firstCall.args[0].message, 'Missing configuration key');
+    });
+
+    it('should return 404 when service throws a "404" error for a non-existent key', async () => {
```
sKrzysieK, line 156: what's the difference between the test above and this one? Why for `' '` we get `400` error code and for any other non-existent key we get `404`?

sKrzysieK (commented):
Generally LGTM but there are a few things I'd like to be clarified

Control/test/lib/controllers/mocha-qcConfiguration.controller.test.js
Deaponn: Thats right. As this test and its description was inspired by [the following code](https://github.com/AliceO2Group/WebUi/blob/dev/Control/test/lib/controllers/mocha-qcConfiguration.controller.test.js#L141), I will correct other tests descriptions. Unless we decide to interpret "empty string" as one that satisfies `key.trim() === ''`, this would be right thing to do

Control/test/lib/controllers/mocha-qcConfiguration.controller.test.js
Deaponn: I think it was originally interpreted that key ` ` could not be valid, and it is another case from providing wrong key. This should be adjusted IMO, to be discussed with others.

Control/lib/adapters/QCConfigurationAdapter.js
```diff
@@ -0,0 +1,42 @@
...
+ * possible types are "string", "boolean", "number", "array<`${NestedTypeMap}`>", `${NestedTypeMap}`
+ * @param {string | Array | Object} value that we want to get the TypeMap of
+ * @returns {string | TypeMap} derived type from the given value, could be a string, or further nested TypeMap
+ */
+const deriveValueType = (value) => {
+  // TODO: implement function _combineTypes, so we can derive Type of value[0] and
+  // then combine it with Types of value[1], value[2] and so on to get the overall Type of values held in the array
+  if (value instanceof Array) { return "array"; }
+  if (value instanceof Object) { return computeRestrictions(value); }
+  if (value.toLowerCase() === "true" || value.toLowerCase() === "false") { return "boolean"; }
```
graduta: We use the `toLocaleLowerCase()` in the project

Control/lib/adapters/QCConfigurationAdapter.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Derive the type of value and return it as a string
+ * possible types are "string", "boolean", "number", "array<`${NestedTypeMap}`>", `${NestedTypeMap}`
+ * @param {string | Array | Object} value that we want to get the TypeMap of
+ * @returns {string | TypeMap} derived type from the given value, could be a string, or further nested TypeMap
+ */
+const deriveValueType = (value) => {
+  // TODO: implement function _combineTypes, so we can derive Type of value[0] and
```
graduta: TODOs should be accompanied by Jira Ticket numbers

Control/lib/services/QCConfiguration.service.js
```diff
@@ -37,6 +38,7 @@ class QCConfigurationService {
    * Get keys of configurations stored in Consul
    * @param {String} prefix - prefix to filter the keys
    * @param {boolean} [recurse=false] - whether to recurse into subdirectories
+   * @returns {Array<string>} names of configurations which are valid JSON
```
graduta: Any `async` function will always return a `Promise` with certain data types inside.
I see on the L51 you used Promise as return type. You should do the same here

Control/lib/services/QCConfiguration.service.js
```diff
@@ -79,10 +83,22 @@ class QCConfigurationService {
     return parsedData;
   }
 
+  /**
+   * Get configuration restrictions by key from Consul
+   * @param {string} key - the key of the configuration
+   * @return {Promise<TypeMap>}
+   * @throws {NotFoundError | ServiceUnavailableError}
```
graduta: Being a promise it will not 'throw' but reject. Above on L51 and below L101, you added the Error type to the promise. 
It would be good to keep one convention only.

Control/lib/typedefs/Restrictions.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @typedef {Object.<string, TypeMapEntry>} TypeMap
+ * 
+ * Object which is a map of types.
+ * Keys are taken from existing configuration.
+ * Values are describing what is expected type of value held there.
+ * 
+ * For example for the given configuration:
+ * ```
+ *  {
```
graduta, line 23: It would be better if you use a qc alike configuration object rather than invented one

Control/test/lib/adapters/mocha-consul-configuration.adapter.test.js
```diff
@@ -0,0 +1,48 @@
...
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+/* eslint-disable max-len */
```
graduta: Does this comment apply? I do not see any excessive long lines

Control/test/lib/adapters/mocha-consul-configuration.adapter.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+const assert = require("assert");
+
+const {
+  computeRestrictions,
+} = require("../../../lib/adapters/QCConfigurationAdapter.js");
+
+describe(`'QCConfigurationAdapter' test suite`, () => {
+  it("should work for minimal input", async () => {
```
graduta: Perhaps you can add a test for `undefined` or `null` passed values?

Control/lib/adapters/QCConfigurationAdapter.js
```diff
@@ -0,0 +1,42 @@
...
+  // then combine it with Types of value[1], value[2] and so on to get the overall Type of values held in the array
+  if (value instanceof Array) { return "array"; }
```
graduta: I believe the standard for checking an object is an array is `Array.isArray()`

Control/lib/adapters/QCConfigurationAdapter.js
```diff
@@ -0,0 +1,42 @@
...
+  if (value instanceof Object) { return computeRestrictions(value); }
+  if (value.toLowerCase() === "true" || value.toLowerCase() === "false") { return "boolean"; }
+  if (!Number.isNaN(Number(value))) { return "number"; }
+  return "string";
```
graduta: What about checking if the value is of typeof string and if not, returning `unknown`? This could be helpful in some cases to display warnings if the current stored value is in bad shape

graduta (changes requested):
For Control project, we limit ourselves to using `'` single quotes for strings, imports, etc. Please update so across your PR

Control/lib/services/QCConfiguration.service.js
```diff
@@ -79,10 +83,21 @@ class QCConfigurationService {
...
+    return QCConfigurationAdapter.computeRestrictions(configuration);
+  }
+  
   /**
    * Edit configuration by key in Consul
-   * @param {String} key - the key of the configuration
-   * @param {String} value - the configuration
+   * @param {string} key - the key of the configuration
+   * @param {string} value - the configuration
+   * @returns {Promise<JSON>} - JSON object with the status of the transaction
```
sKrzysieK: Is `JSON` a proper docstring type?

sKrzysieK (dismissed):
Looks good to me

Control/lib/services/QCConfiguration.service.js
Deaponn: Good catch! I will change it to `Object`, this seems more suitable
edit:
I found where I took this type from: [here you can see the type](https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/services/consul.service.js#L149), I think we should wait for opinion from others

graduta (approved):

PR #2971, opened 2025-07-08
[OGUI-1705] Design basic CentralSystem - gRPCWrapper stream communication

OmegaCreations:
- [x] PR labels are selected

This PR introduces a new gRPCWrapper class and ConnectionManager class files with constructor both for client and central system. Creates basic insecure duplex stream between client and central system.

In the scope of this task, do the reasearch if we need unary communication - I don't see the reason to use unary communication for now. We don't do much load of data between the client and central system so additional protocol overhead shouldn't be a problem. Moreover, we won't to mainly persist stream connection between services so additional unary communication wouldn't be needed.


Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // Handle stream end event
+    call.on("end", () => {
+      this.logger.infoMessage(`Client ${call.getPeer()} ended stream.`);
+      call.end();
+    });
+
+    // Handle stream error event
+    call.on("error", (err) =>
+      this.logger.infoMessage(
```
graduta: perhaps in case of error, we use the `errorMessage` instead of `infoMessage`?

Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  // class properties
+  private server: grpc.Server;
+
+  /**
+   * Initializes the Wrapper for CentralSystem.
+   * @param port The port number to bind the gRPC server to.
+   */
+  constructor(private protoPath: string, private port: number) {
+    this.server = new grpc.Server();
```
graduta, line 41: Should the server not recieve a configuration object or parameters such as:
- host on which it should be running.
- port on which it should be running, etc.

Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,100 @@
...
+   * @desciprion Starts the gRPC server and binds it to the specified in class port.
+   */
+  public listen() {
+    const addr = `localhost:${this.port}`;
+    this.server.bindAsync(
+      addr,
+      grpc.ServerCredentials.createInsecure(),
+      (err, _port) => {
+        if (err) {
+          this.logger.infoMessage("Server bind error:", err);
```
graduta: If there is an error, we need to log it as error

Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,100 @@
...
+      }
+    );
+  }
+}
+
+// Instantiate the CentralSystemWrapper on port 50051, but don't start automatically
+const PROTO_PATH = path.join(__dirname, "../proto/wrapper.proto");
+const centralSystem = new CentralSystemWrapper(PROTO_PATH, 50051);
+// Start listening explicitly
+centralSystem.listen();
```
graduta: If you do this, these lines of code will be executed at the time the file is imported. We should instead have this in a method of the class to be initialized. 
Have a look at a gRPC client example that we use in another project: https://github.com/AliceO2Group/WebUi/blob/dev/Control/lib/control-core/GrpcServiceClient.js 
You can also use as an example this gRPC server from one of our projects
- https://github.com/AliceO2Group/Bookkeeping/blob/main/lib/server/GRPCServer.js

graduta (changes requested):

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Done

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Done, thank you for noticing.

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Hi, yes, but in this example we have method `.listen()` that is starting to listen. The constructor is only loading .proto files that will be used later. Both in `gRPCWrapper` and in `CentralSystemWrapper` I have added these lines of codes below for siplified "e2e" test to check if everything works and if communication between client-central is fine. In the end these lines of initialization will be removed and only used in CentralSystem code. 

Am I understanding this correctly, or did you mean something else? Please let me know.

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Done, changed it to config object.

Tokenization/backend/wrapper/central/CentralSystem.ts
graduta: But the idea of being it configurable is not possible anymore because you are hardcoding the parameters passed in the configuration object here to `localhost` and `50051`

The CentralSystemWrapper should be instantiated only when needed in the project with configuration objects which are automatically deployed per environment.

Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.logger.infoMessage(`Client ${call.getPeer()} ended stream.`);
+      call.end();
+    });
+
+    // Handle stream error event
+    call.on("error", (err) =>
+      this.logger.errorMessage(
+        `Stream error from client ${call.getPeer()}:`,
+        err
+      )
```
graduta, line 92: This is not the correct way of using the `errorMessage` method.
If you provide a second parameter to the call, it will be expected to be an infologger configuration object and not an error.

You should concatenate the string with your error in one string

Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,111 @@
...
+   * @desciprion Starts the gRPC server and binds it to the specified in class port.
+   */
+  public listen() {
+    const addr = `${this.host}:${this.port}`;
+    this.server.bindAsync(
+      addr,
+      grpc.ServerCredentials.createInsecure(),
+      (err, _port) => {
+        if (err) {
+          this.logger.errorMessage("Server bind error:", err);
```
graduta: Same here, please have a look at the library itself:
https://github.com/AliceO2Group/WebUi/blob/dev/Framework/Backend/log/Logger.js#L117

Tokenization/backend/wrapper/test/central/CentralSystem.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import { CentralSystemWrapper } from "../../central/CentralSystem";
+import * as grpc from "@grpc/grpc-js";
+
+describe("CentralSystemWrapper", () => {
+  let wrapper: CentralSystemWrapper;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    wrapper = new CentralSystemWrapper("dummy.proto", 12345);
```
graduta: This constructor expects a JSON object with a configuration and not multiple parameters.

```js
new CentralSystemWrapper(
  {
    host: "dummy.proto",
    port: 12345,
  }
);
```

Tokenization/backend/wrapper/central/CentralSystem.ts
```diff
@@ -0,0 +1,111 @@
+import * as grpc from "@grpc/grpc-js";
+import * as protoLoader from "@grpc/proto-loader";
+import path from "path";
+import { LogManager } from "@aliceo2/web-ui";
+import { CentralSystemConfig } from "../models/config.model";
+
+/**
+ * @description Central System gRPC wrapper that manages client connections and handles gRPC streams with them.
+ */
+export class CentralSystemWrapper {
```
graduta, line 23: The export should match the file name

Tokenization/backend/wrapper/models/message.model.ts
graduta, line 1: I believe this file was added already in another PR: https://github.com/AliceO2Group/WebUi/pull/2940/files 
Perhaps you did not merge `dev` since the other PR was merged?

Tokenization/backend/wrapper/client/ConnectionManager/ConnectionManager.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  /**
+   * @description Schedules a reconnect with exponential backoff.
+   */
+  private scheduleReconnect() {
+    this.reconnectAttempts++;
+    const delay = Math.min(1000 * 2 ** this.reconnectAttempts, 30000);
+    setTimeout(() => {
```
graduta, line 112: I see that you are tracking the reconnection attempts which is very good but you are not storing the current timeout attempt reconnection. This can lead to overlaps of timeouts which would attempt multiple connections even if one of them is successul.

What do you think about making sure a new timeout is not set if one exists already?

Tokenization/backend/wrapper/client/ConnectionManager/ConnectionManager.ts
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * @description Disconnects from the gRPC stream and resets attempts.
+   */
+  public disconnect() {
+    if (this.stream) {
+      this.stream.end();
+      this.stream = undefined;
+    }
+    this.reconnectAttempts = 0;
+    this.logger.infoMessage(`Disconnected from CentralSystem service`);
```
graduta: Is the logging message supposed to be outside the `if` condition?
If there is no stream, you did not really disconnect, right?

Tokenization/backend/wrapper/client/gRPCWrapper.ts
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * @description Starts the Connection Manager stream connection with Central System
+   */
+  public connectToCentralSystem(): void {
+    this.ConnectionManager.connectToCentralSystem();
+  }
+}
+
+const PROTO_PATH = path.join(__dirname, "../proto/wrapper.proto");
+const grpc = new gRPCWrapper({
```
graduta: Same as previous file, I believe these should be called by the client and not hardcoded in the file. `path`, proto path, central address, etc.

The wrapper file should offer the developers the flexbility of when to instantiate a connection

graduta (changes requested):
Added a few more comments which I believe need to be tackled before proceeding

Tokenization/backend/wrapper/client/ConnectionManager/ConnectionManager.ts
OmegaCreations: In the newest PR's I have removed this scheduler due to some bugs. I would like to work on that one in the scope of newer small task. I will create a ticket with this one. It will help because many changes were made and this one probably should be done in a separate file/class.

Tokenization/backend/wrapper/client/gRPCWrapper.ts
OmegaCreations: Removed.

Tokenization/backend/wrapper/client/ConnectionManager/ConnectionManager.ts
OmegaCreations: Thank you for noticing. Fixed.

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Fixed

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Fixed, thank you for providing context. Helped a lot

Tokenization/backend/wrapper/test/central/CentralSystem.test.ts
OmegaCreations: Fixed

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Changed file name.

Tokenization/backend/wrapper/models/message.model.ts
OmegaCreations: Yes, sorry, I was merging `main` instead of `dev`. Should be fine now.

Tokenization/backend/wrapper/central/CentralSystem.ts
OmegaCreations: Removed instantiation

graduta (dismissed):

graduta (approved):

PR #2973, opened 2025-07-10
[OGUI-1715] Implement initial structure and functionality of the Central System.

dwedr:
- [x] PR labels are selected

This PR sets up the initial structure for the Central System / Token Exchange Service. It includes basic functionality, placeholder logic, src/ and dist/ separation, TypeScript config, and required Node modules.


dwedr:
I think I mentioned this in the last meeting, but it’s worth noting:
Regarding tests — they will be done separately for each functionality, even in this simple version, as parts of other tasks. This is to stick to the way they are created and tracked in Jira.

PR #3022, opened 2025-08-29
[OGUI-1722] Runs mode backend: adapt for frontend support 

mariscalromeroalejandro:
- [x] PR labels are selected

This PR updates the backend to support the frontend in utilizing the `Runs Mode` functionality. Changes include:

- Added a new endpoint to fetch the status of runs by providing the run number.
- Implemented middlewares for validation in controllers when a run number is provided.
- Updated tests to cover the new endpoint and validation logic.


QualityControl/lib/middleware/filters/runMode.middleware.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Function} next - The next middleware function in the stack.
+ * @returns {Promise<void>}
+ */
+export const runModeMiddleware = async (req, res, next) => {
+  const { inRunMode = false, filters = {} } = req.query;
+  if (!inRunMode) {
+    return next();
+  }
+
+  const parsedRunNumber = await validateRunNumber(filters?.RunNumber, res);
```
graduta: Ideally we should catch the error and deal with the response in the middleware and not the helper function

graduta: In this manner the helper function has one role only, validatation

QualityControl/lib/services/FilterService.js
```diff
@@ -13,6 +13,7 @@
  */
 
 import { LogManager } from '@aliceo2/web-ui';
+import { RunStatus } from '../../common/library/runStatus.enum.js';
 const logger = LogManager.getLogger('filter/service');
```
graduta: Please update the logger to be instantiated in the constructor rather than at top of the wile.
This will make a difference between the logger being instantiated at FileImport versus Class Instatiate

graduta: Moreover, please see my PR about using loggers in QC. Because we deploy multiple instances of QC on the same machine we have to always use the `process.env` parameter to let us know what instance we use.
For reference: https://github.com/AliceO2Group/WebUi/pull/3008

graduta (changes requested):
A few changes to do on the error bubble up and good to go afterwards

graduta (approved):

PR #3038, opened 2025-09-05
[OGUI-1757] Update layouts controller to allow objects path

Houwie7000:
- [X] PR labels are selected


QualityControl/test/lib/controllers/LayoutController.test.js
```diff
@@ -147,6 +147,97 @@ export const layoutControllerTestSuite = async () => {
...
+      };
+      const layoutConnector = new LayoutController(jsonStub);
+
+      await layoutConnector.getLayoutsHandler(req, res);
+
+      ok(res.status.calledWith(400), 'Response status was not 400');
+      ok(res.json.calledOnce, 'Response was not sent');
+
+      const [actual] = res.json.firstCall.args;
+      ok(actual && actual.status === 400, 'status incorrect');
```
mariscalromeroalejandro: please check how we check a json response in other tests

QualityControl/lib/controllers/LayoutController.js
```diff
@@ -61,10 +61,12 @@ export class LayoutController {
   async getLayoutsHandler(req, res) {
     let fields = undefined;
     let owner_id = undefined;
+    let object_path = undefined;
 
     try {
       const validated = await LayoutsGetDto.validateAsync(req.query);
-      ({ fields, owner_id } = validated);
+      // eslint-disable-next-line no-unused-vars
+      ({ fields, owner_id, object_path } = validated);
```
mariscalromeroalejandro: please, add it as a parameter to the repository method even if you are not going to use it in this pr (to avoid this warning messages)

QualityControl/test/lib/controllers/LayoutController.test.js
```diff
@@ -147,6 +147,97 @@ export const layoutControllerTestSuite = async () => {
...
+      ok(res.json.calledOnce, 'Response was not sent');
+
+      ok(res.json.calledWith({
+        message: 'Invalid query parameters: "Object path" must be a string',
+        status: 400,
+        title: 'Invalid Input',
+      }), 'Error message was incorrect');
+    });
+
+    test('should return layout when object_path contain a valid value', async () => {
```
mariscalromeroalejandro: fix typo
```suggestion
    test('should return layout when object_path contains a valid value', async () => {
```

QualityControl/test/lib/controllers/LayoutController.test.js
```diff
@@ -147,6 +147,97 @@ export const layoutControllerTestSuite = async () => {
...
+      };
+      const layoutConnector = new LayoutController(jsonStub);
+
+      await layoutConnector.getLayoutsHandler(req, res);
+
+      ok(res.json.calledOnce, 'Response was not sent');
+      ok(res.json.calledWith(response), 'A list of layouts should have been sent back');
+    });
+
+    test('should return 400 when object_path contain an invalid character: -', async () => {
```
mariscalromeroalejandro: ditto
```suggestion
    test('should return 400 when object_path contains an invalid character: -', async () => {
```

QualityControl/test/lib/controllers/LayoutController.test.js
```diff
@@ -147,6 +147,97 @@ export const layoutControllerTestSuite = async () => {
...
+        },
+      };
+      const layoutConnector = new LayoutController(jsonStub);
+
+      await layoutConnector.getLayoutsHandler(req, res);
+
+      ok(res.status.calledWith(400), 'Response status was not 400');
+      ok(res.json.calledOnce, 'Response was not sent');
+
+      const [actual] = res.json.firstCall.args;
```
mariscalromeroalejandro: ditto

mariscalromeroalejandro (commented):
Great work on this PR. I’ve left a few comments with minor suggestions for improvement.

QualityControl/lib/dtos/LayoutDto.js
```diff
@@ -84,6 +84,7 @@ export const LayoutDto = Joi.object({
 export const LayoutsGetDto = Joi.object({
   owner_id: Joi.number().integer().optional(),
   name: Joi.string().optional(),
+  object_path: Joi.string().pattern(/^[A-Za-z0-9_/]+$/).optional().max(100).label('Object path'),
```
graduta: The DTO is not correct. Objects can also contain for example the charcater `-` which I have seen that you added tests for being invalid. 
If you have a look through our production instance you can see what characters we should allow or not

graduta (commented):
Just one small comment added.
In the future, I would recommend the order of tickets to be from ground-up.
For example, right now you are adding an API new parameter but even if the users would make use of it, it will do nothing.

Going the other way around, first adding service parameter option and then make use of this in the controller, would ensure when the API is released, the functionality is also there.

Moreover, the `object_path` should be added as `objectPath` as convention. `owner_id` it is a custom one coming from the OpenID service that CERN uses and not our naming convention.

I would also recommend that the object parameter is contained in a `filter` parameter so that in the future we can add more such parameters allowing the users to search

graduta (approved):

PR #3041, opened 2025-09-08
[OGUI-1758] object path filtering in repository layer

Houwie7000:
- [x] PR labels are selected


QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -24,16 +24,31 @@ export class LayoutRepository extends BaseRepository {
    * @param {number} [options.owner_id] - Filter layouts by owner ID
    * @param {string} [options.name] - Filter layouts by exact name match
    * @param {Array<string>} [options.fields] - Array of field names to include in each returned layout object
+   * @param {string} [options.filter] - Filter layouts by containing filter.objectPath, case insensitive
```
graduta: I believe following the changes in the previous PR, the filter now is an object rather than a string?

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -24,16 +24,31 @@ export class LayoutRepository extends BaseRepository {
...
-  listLayouts({ name, owner_id, fields = [] } = {}) {
+  listLayouts({ name, owner_id, fields = [], filter } = {}) {
     const { layouts } = this._jsonFileService.data;
+    const filterPresent = filter !== undefined;
 
-    const layoutFilter = (layout) =>
-      (owner_id === undefined || layout.owner_id === owner_id) &&
-      (name === undefined || layout.name === name);
+    // Better to lowercase once rather than at every single object_path comparison.
+    if (filterPresent && filter.objectPath !== undefined) {
```
graduta: I would recommend you make use of the optional chaining in this instance for improved readability. Something along the lines:
```js
if (filter?.objectPath !== undefined) {
  // ...
}
```
Same for the lines below
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -24,16 +24,31 @@ export class LayoutRepository extends BaseRepository {
...
+      filter.objectPath = filter.objectPath.toLowerCase();
+    }
+
+    // Filter using filter.objectPath should filter and its objectPath be defined,
```
graduta, line 40: We normally prefer to use method documentation. So in this case, rather than adding single line comments `//`, I would encourage you to add comments straight to the definition of the arrow function below

For example
```js
    const layoutFilter = filterPresent && filter.objectPath !== undefined ?
      /**
       * Function doc
       * @param {Layout} layout - param doc
       * @returns documentation agaon
       */
      (layout) =>
        (layout.tabs ?? [])
          .flatMap((tab) => tab.objects ?? [])
          .filter((object) => (object.name ?? '').toLowerCase().includes(filter.objectPath)).length > 0
      :
      /**
       * Function doc
       * @param {Layout} layout - param doc
       * @returns documentation agaon
       */
      (layout) =>
        (owner_id === undefined || layout.owner_id === owner_id) &&
        (name === undefined || layout.name === name);

graduta (changes requested):

graduta (approved):

PR #3065, opened 2025-09-19
[OGUI-1760] Update search UI for object path and add filters

Houwie7000:
- [x] PR labels are selected


QualityControl/public/pages/layoutListView/filtersPanelPopover.js
```diff
@@ -0,0 +1,92 @@
...
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+// Adopted from Bookkeeping/lib/public/components/Filters/common/filtersPanelPopover.js
+import { h, popover, PopoverAnchors, PopoverTriggerPreConfiguration } from '/js/src/index.js';
+// import { tooltip } from '../../common/popover/tooltip.js';
```
graduta: Code that is not needed should not be left as comments

QualityControl/public/pages/layoutListView/filtersPanelPopover.js
```diff
@@ -0,0 +1,92 @@
...
+ * @param {SearchFilterModel} searchFilterModel {@link SearchFilterModel} filtering model.
+ * @returns {Component} main panel header.
+ */
+const filtersToggleContentHeader = (searchFilterModel) => h('.flex-row.justify-between', [
+  h('.f4', 'Filters'),
+  h(
+    'button#reset-filters.btn.btn-danger',
+    {
+      onclick: () => {
+        searchFilterModel.resetAll();
```
graduta: For one line content we normally do not open `{}` anymore
```
onclick: () => searchFilterModel.resetAll();
```

QualityControl/public/pages/layoutListView/FilterTypes.js
```diff
@@ -0,0 +1,75 @@
...
+    },
+    reset: () => {
+      value = '';
+    },
+  };
+}
+
+/**
+ * Creates a multiple value filter, key with array value.
+ * Not used in the code yet but serves as an example.
```
graduta: We should not add code that is not used

QualityControl/public/pages/layoutListView/FilterTypes.js
```diff
@@ -0,0 +1,75 @@
...
+ * @param {string|null} friendlyName - friendly name of the filter.
+ * @param {string|null} inputPlaceholder - input placeholder text.
+ * @param {string} value - value associated with key.
+ * @returns {Filter} key-value filter object.
+ */
+export function createKeyValueFilter(key, friendlyName = null, inputPlaceholder = null, value = '') {
+  return {
+    key,
+    friendlyName: () => friendlyName ? friendlyName : key,
+    inputPlaceholder: () => inputPlaceholder ? inputPlaceholder : 'Conditions',
```
graduta: I would recommend not putting a placeholder as `Conditions` does not say much about a filter

QualityControl/public/pages/layoutListView/FilterTypes.js
```diff
@@ -0,0 +1,75 @@
...
+    getValue: () => value ? value : null,
+    // trim checks if value is a string value, test this
+    isActive: () => Boolean(value && value.trim()),
+    set: (v) => {
```
graduta, line 42: One liner

QualityControl/public/pages/layoutListView/FilterTypes.js
```diff
@@ -0,0 +1,75 @@
...
+      value = v;
+    },
+    reset: () => {
```
graduta, line 45: Same

QualityControl/public/pages/layoutListView/filtersPanelPopover.js
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Return the filters panel popover content section
+ * @param {SearchFilterModel} searchFilterModel the searchFilter model
+ * @returns {Component} the filters section
+ */
+export const filtersSection = (searchFilterModel = {}) => [
+  searchFilterModel.getAll().flatMap((filter) => [
+    h('.flex-row.g2', [
+      h('.w-30.f5.flex-row.items-center.g2', [filter.friendlyName()]),
```
graduta: `filter.friendlyName()` returns a string, there is no need for an array here

QualityControl/public/pages/layoutListView/model/LayoutListModel.js
```diff
@@ -28,8 +29,16 @@ export default class LayoutListModel extends BaseViewModel {
   constructor(model) {
     super();
     this.model = model;
-    this._searchInput = '';
     this.folders = new Map();
+    this.searchFilterModel = new SearchFilterModel();
+    this.searchFilterModel.observe(() => {
+      this.notify();
```
graduta: Is there a need for this `notify`? Currently there is no change to notify and later, the search method notifies anyway

QualityControl/public/pages/layoutListView/model/SearchFilterModel.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  /**
+   * Return the active filters in a representable way.
+   * @returns {string} Active filters: filter.friendlyName(), ... .
+   */
+  stringifyActiveFiltersFriendly() {
+    let activeFilterText = 'Active filters: ';
+    if (this.allInActive()) {
+      activeFilterText += 'None.';
```
graduta: I would recommend that if no filters are set that this text is not displayed

QualityControl/public/pages/layoutListView/model/SearchFilterModel.js
```diff
@@ -0,0 +1,167 @@
...
+      return activeFilterText;
+    } else {
+      for (const filter of this.getAllActive()) {
```
graduta: I believe L157-L162 can be written in an easier way by using the `join` method. What do you think of:
```js
const activeNames = this.getAllActive().map((filter) => filter.friendlyName());
activeFilterText += activeNames.join(', ') + '.';
```

graduta (changes requested):
Good effort. I raised a few improvements that I think will make the user experience better and improve the code's readability

QualityControl/public/pages/layoutListView/components/LayoutListHeader.js
```diff
@@ -15,18 +15,10 @@
...
-    h('input.form-control.form-inline.mh1.w-33', {
-      placeholder: 'Search',
-      type: 'text',
-      value: layoutListModel.searchInput,
-      oninput: (e) => layoutListModel.search(e.target.value),
-    }),
-  ]),
-});
+export default () => [
+  h('.w-50.text-center', [h('b.f4', 'Layouts')]),
```
graduta: Currently you are not displaying the header Label `Layouts` anymore. This is because there has been a change in `dev` where the headers are now expected to be in format of `centrlCol` and `rightCol`. 
I believe when you merged from dev, you did not realize this and kept your changes instead of merging them together. 

Please have another look at the changes in this file

QualityControl/public/pages/layoutListView/filtersPanelPopover.js
graduta: I see you replied with "👍 " on this commend but the commented out code is still there.

QualityControl/public/pages/layoutListView/FilterTypes.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {string} value - value associated with key.
+ * @returns {Filter} key-value filter object.
+ */
+export function createKeyValueFilter(key, friendlyName = null, inputPlaceholder = null, value = '') {
+  return {
+    key,
+    friendlyName: () => friendlyName ? friendlyName : key,
+    inputPlaceholder: () => inputPlaceholder ? inputPlaceholder : '',
+    getValue: () => value ? value : null,
+    // trim checks if value is a string value, test this
```
graduta, line 40: We should not just disable an eslint if inconvenient. 
I think in this case, you cannot just remove the `{}` because of the other eslint rule not allowing assignments in arrow functions.
Thus, a cleaner solution is to apply the exception of one line function (contrary to the previous comments, I understand)
```
set: () => {
  value = v:
}
```

graduta: I understand it is actually the same as you initially added. It was an oversight from my side, forgetting about the assignment eslint rule, I apologise for that

QualityControl/public/pages/layoutListView/model/SearchFilterModel.js
```diff
@@ -0,0 +1,163 @@
...
+    return Object.fromEntries(this.getAll().map((filter) => [filter.key, filter.getValue()]));
+  }
+
+  /**
+   * Return the active filters in a representable way.
+   * @returns {string} Active filters: filter.friendlyName(), ... .
+   */
+  stringifyActiveFiltersFriendly() {
+    let activeFilterText = 'Active filters: ';
+    if (this.allInActive()) {
```
graduta: If I understand this bit correctly, right now for this check, it works simply because there is only one filter present, but the moment we will add another one, the condition will not hold anymore.
I think is better to use the `getAllActive` and compare to 0, if no filter is active then no text is displayed.

QualityControl/public/pages/layoutListView/model/LayoutListModel.js
```diff
@@ -28,8 +29,15 @@ export default class LayoutListModel extends BaseViewModel {
   constructor(model) {
     super();
     this.model = model;
-    this._searchInput = '';
     this.folders = new Map();
+    this.searchFilterModel = new SearchFilterModel();
+    this.searchFilterModel.observe(() => {
+      if (!this.searchFilterModel.allInActive()) {
```
graduta: I believe the filter should only send the active filters and not all of them. For example if we have 2 parameters, objectPath and tab, we only want to attach one of them to the request.
For testing purposes, try to register another filter

graduta: Moreover, I think here the if condition should as well use the getActive like recommended in the other comment

QualityControl/public/pages/layoutListView/model/SearchFilterModel.js
```diff
@@ -0,0 +1,163 @@
...
+  constructor() {
+    super();
+
+    /**
+     * Filters storage map
+     * @type {Map<string,Filter>}
+     */
+    this.filters = new Map(); // key -> filter instance
+    this.searchInput = '';
+    this.register(createKeyValueFilter('objectPath', 'Object path', 'e.g. TPC'));
```
graduta: The registration should happen in the model using the search model as it is dependent on the page. In this case the LayoutListModel

graduta (commented):
I added a few comments for improving the logic of the code and the build of the request

QualityControl/public/pages/layoutListView/model/SearchFilterModel.js
Houwie7000: this.allInActive is only there so that the text will return empty when there are no filters with active values. When one or more filters have a value allInActive will resolve false and thus the function will return a formatted string. Under the hood the getAllActive basically did what you propose. I have modified the allInActive function to remove the duplicate code with getAllActive.

QualityControl/public/pages/layoutListView/model/LayoutListModel.js
Houwie7000: getAllAsObject refactored into getAllActiveAsObject. Beheviour is as expected now with multiple filters or multiple registered with only one active.

QualityControl/public/pages/layoutListView/model/SearchFilterModel.js
Houwie7000: Done, should the searchInput be put back into the LayoutListModel? I put it in the SearchFilterModel as it seemed good to concentrate all search related data there. However I am not actually ever using it there and only using it as a value storage.

PR #3067, opened 2025-09-22
[OGUI-1768] Run mode: smart refresh for object view(s)

mariscalromeroalejandro:
- [x] PR labels are selected

This PR adds a new check to the object view page and the object tree visualization. When in run mode with an active run, the view will now refresh only when necessary (specifically, if the ID of the last fetch differs from the currently selected ID)

### DIFF
--- a/QualityControl/public/object/QCObject.js
+++ b/QualityControl/public/object/QCObject.js
@@ -329,9 +329,10 @@ export default class QCObject extends BaseViewModel {
    * Set the current selected object by user
    * Search within `currentList`;
    * @param {QCObject} object - object to be selected and loaded
-   * @returns {undefined}
+   * @param {object} [preloadedData] - optional object data already fetched
+   *  @returns {undefined}
    */
-  async select(object) {
+  async select(object, preloadedData = null) {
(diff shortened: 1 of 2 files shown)


QualityControl/public/object/QCObject.js
```diff
@@ -549,15 +549,29 @@ export default class QCObject extends BaseViewModel {
...
    */
   async triggerFilter() {
-    // don't clear selected object refreshing in run mode
     if (!this.model.filterModel.isRunModeActivated || !this.model.filterModel.runsModeInterval) {
       this.selected = null;
     }
+    if (this.selected && this.selected.name) {
+      const { refreshNeeded, data } = await this.checkIfRefreshSelectedObject();
+      if (refreshNeeded) {
+        this.select(data?.payload ?? { name: this.selected.name });
```
graduta: The select method will call again the refresh to be fetched from the back-end.
Given that we already did that, I think you should look at how you can avoid the second request

graduta (changes requested):
It would be good to not duplicate the request of object in object tree

graduta (approved):

PR #3068, opened 2025-09-22
[OGUI-329] Add Kafka from ECS Syncronizer in QCG

graduta:
- [x] PR labels are selected

PR which:
* Adds AliEcsSyncronizer in QCG to consume Kafka Events from ECS as proto objects
* Add internal EventEmitter to update ongoing runs stored by QCG
* Adds `kafkajs` as dependency


QualityControl/lib/dtos/KafkaConfigurationDto.js
```diff
@@ -0,0 +1,53 @@
...
+    if (process.env.NODE_ENV === 'PRODUCTION') {
+      return value;
+    }
+    if (value === 'qc-gui') {
+      return helpers.error('any.invalid');
+    }
+    return value;
+  }, 'qc-gui not allowed unless in PRODUCTION'),
+  consumerGroups: Joi.object().custom((obj, helpers) => {
+    if (process.env.NODE_ENV === 'PRODUCTION') {
```
mariscalromeroalejandro: I would move this to a variable as we repeat the same logic twice:
```js
const isProduction = () => process.env.NODE_ENV === 'PRODUCTION';
```

QualityControl/lib/dtos/KafkaConfigurationDto.js
```diff
@@ -0,0 +1,53 @@
...
+ * @throws {InvalidInputError}
+ */
+export const KafkaConfigDto = Joi.object({
+  enabled: Joi.boolean().default(false),
+  clientId: Joi.string().custom((value, helpers) => {
+    if (process.env.NODE_ENV === 'PRODUCTION') {
+      return value;
+    }
+    if (value === 'qc-gui') {
+      return helpers.error('any.invalid');
```
mariscalromeroalejandro: maybe it would be better to provide a more descriptive message so debugging is easier
```suggestion
      return helpers.error('any.invalid', { message: 'clientId "qc-gui" is not allowed outside PRODUCTION');
```

QualityControl/lib/repositories/LayoutRepository.js
```diff
@@ -90,7 +90,7 @@ export class LayoutRepository extends BaseRepository {
    * Given a string, representing layout name, retrieve the layout if it exists
    * @param {string} layoutName - name of the layout to retrieve
    * @returns {Layout} - object with layout information
-   * @throws
+   * @throws {NotFoundError} - if the layout is not found
```
mariscalromeroalejandro: does this belong to this pr?

QualityControl/lib/services/RunModeService.js
```diff
@@ -98,6 +104,38 @@ export class RunModeService {
...
+
+  /**
+   * Handles run track events emitted by the event emitter.
+   * Updates the ongoing runs cache based on the transition type.
+   * @param {object} runEvent - Object containing runNumber and transition type.
+   * @param {number} runEvent.runNumber - The run number associated with the event.
+   * @param {string} runEvent.transition - The transition type (e.g., 'START_ACTIVITY', 'STOP_ACTIVITY').
+   * @returns {Promise<void>}
+   */
+  async _onRunTrackEvent({ runNumber, transition }) {
```
mariscalromeroalejandro, line 123: I would check if runNumber has been provided as a safeguard:
```js
if (!runNumber) {
    this._logger.errorMessage(`runNumber is missing in run track event`);
    return;
  }
```

mariscalromeroalejandro: maybe checking if transition is there could be useful

QualityControl/test/testUtils/delay.js
```diff
@@ -17,3 +17,21 @@
...
+ * Delay until a condition is met or a timeout occurs
+ * @param {Function} conditionFunc - Function that returns a boolean indicating if the condition is met
+ * @param {number} ms - Time in milliseconds to wait between checks
+ * @param {number} maxRetries - Maximum number of retries before giving up
+ * @returns {Promise<void>}
+ */
+export const delayAndCheck = async (conditionFunc, ms = 500, maxRetries = 10) => {
+  let retries = 0;
+  while (retries < maxRetries) {
+    if (conditionFunc()) {
```
mariscalromeroalejandro: if an async function is passed it will always be truthy... I would do:
```suggestion
    const result = await conditionFunc()
    if (result) {
```

mariscalromeroalejandro (commented):
small comments added

QualityControl/lib/QCModel.js
```diff
@@ -44,12 +48,17 @@ import { objectsGetValidationMiddlewareFactory } from './middleware/objects/obje
 import { objectGetContentsValidationMiddlewareFactory }
   from './middleware/objects/objectGetContentsValidationMiddlewareFactory.js';
 import { RunModeService } from './services/RunModeService.js';
+import { KafkaConfigDto } from './dtos/KafkaConfigurationDto.js';
+
+const LOG_FACILITY = `${process.env.npm_config_log_label ?? 'qcg'}/setup-service`;
```
Houwie7000: I feel that qcg/setup-service is not the only thing the QCModel does, if this is meant specifically for the Kafka service setup then perhaps it should be moved into scope. Otherwise maybe "/model-setup" in case more logs are added here later or for Kafka's initialisation specifically perhaps "/kafka-init"?

Houwie7000 (dismissed):

QualityControl/lib/dtos/KafkaConfigurationDto.js
graduta: Thank you for the suggestion. i believe your suggested change has a syntax error, missing `}`. Moreover, that message would not be shown as default.

I will update into using the message helper instead as per JOI docs:
https://joi.dev/api/?v=17.13.3

QualityControl/lib/dtos/KafkaConfigurationDto.js
graduta: Good suggestion. I extracted it in a util file so that we can also declare the test environment one

QualityControl/lib/services/RunModeService.js
graduta: Indeed, `runNumber` and `transition` are critical in the logic of RunMode and it does not make sense to process events if they are missing.
I have added the check in the AliEcsSyncrhonizer so that we do not spam the event emitter with incorrect kafka messages.

QualityControl/lib/repositories/LayoutRepository.js
graduta: Removed

QualityControl/lib/dtos/KafkaConfigurationDto.js
mariscalromeroalejandro: in that case I would use `helpers.error`

mariscalromeroalejandro (commented):
I replied to a message within a conversation already opened in the PR, apart from that, PR is ready to be merged

QualityControl/lib/dtos/KafkaConfigurationDto.js
graduta: `helpers.error` is to be used when we need a certain custom Joi error to be triggered further down the verification.
In our case, I believe `helpers.message` is the right way to go as it is the method that will be sending the custom message to our `try/catch`

PR #3099, opened 2025-10-03
[OGUI-1784] Update sidebar layout name locally after save

mariscalromeroalejandro:
- [x] PR labels are selected

This PR fixes sync error in which the sidebar showing 'My layouts' was not being updated accordingly after changing the name of a layout in edit mode. With this change, the sidebar refreshes correctly and the new name of the layout is shown after saving (via JSON or via GUI)


graduta (changes requested):
It would be good to add a test on front-end to make sure this does not happen in the future

graduta (approved):

PR #3101, opened 2025-10-06
[OGUI-1600] download single ROOT object service

Houwie7000:
- [x] PR labels are selected

Adds an endpoint that allows you to download a single ROOT object from the qcdb.
This bypasses the firewall blocking public access to qcdb by letting the request go though QCG instead.
QCG is protected by CERN SSO thus it remains safe.


QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    try {
+      const files = await Promise.all(promises);
+      if (files.filter((file) => file === false).length > 0) {
+        throw new Error('qcdb object request failed.');
+      }
+    } catch (error) {
+      this._logger.errorMessage(error.message);
+      res.status(500).send("Unable to retrieve ROOT object('s)");
```
graduta: The service should not deal with using the response object to send an error. 
This method should instead let the error bubble up to the controller which needs to catch it
Same for the error on L94

QualityControl/lib/api.js
```diff
@@ -27,6 +27,9 @@ import { runModeMiddleware } from './middleware/filters/runMode.middleware.js';
  * @param {HttpServer} http - web-ui based server implementation
  * @param {WebSocket} ws - web-ui websocket server implementation
  * @param {EventEmitter} eventEmitter - Event emitter instance (Kafka)
+ * @import {HttpServer} from '@aliceo2/web-ui';
```
graduta, line 30: Why is there a need for this import declaration given that we have the param and type above?

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -69,6 +78,29 @@ export class ObjectController {
...
+   * Download ROOT objects using the QcdbProxy.
+   * Only support 1 root object for now.
+   * @param {Request} req - ExpressJs req object.
+   * @param {Response} res - ExpressJs res object.
+   * @returns {void}
+   */
+  async getDownloadObjects(req, res) {
+    let objectIds = undefined;
+    try {
+      const validated = await ObjectGetDownloadDTO.validateAsync(req.query);
```
mariscalromeroalejandro, line 91: I would move validation to an endpoint

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -69,6 +78,29 @@ export class ObjectController {
...
+      ({ objectIds } = validated);
+    } catch (e) {
+      const responseError = e.isJoi ?
+        new InvalidInputError(`Invalid query parameters: ${e.details[0].message}`) :
+        new Error('Unable to process request');
+
+      this._logger.errorMessage(`Error validating query parameters: ${e}`);
+      return updateAndSendExpressResponseFromNativeError(res, responseError);
+    }
+    this._qcdbDownloadService.getQcdbRootObjects(objectIds, res);
```
mariscalromeroalejandro: why did you leave this outside the try

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+      if (res != undefined) {
+        res.setHeader('Content-Type', contentType ?? 'application/root');
+        res.setHeader('Content-Disposition', contentDisposition ?? `attachment; filename="${filename}"`);
+        if (contentLength) {
+          res.setHeader('Content-Length', contentLength);
+        }
+
+        await this._pipelineAsync(response.body, res);
+        return true;
+      } else {
```
graduta, line 105: I believe the method in this service should concern itself simply with fetching the object from QCDB and returning the response or file (it can be specified in the method declaration)
Then the controller is the one responsible for the operations done and set on the response object

graduta (changes requested):
The PR looks to be in the right direction, good job!
Besides the comments I have added, I would encourage you to also add API tests for the new endpoint that you add

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+   * Constructor
+   * @param {object} config - application's {config.js}.ccdb, needed for Qcdb hostname+port
+   */
+  constructor(config = {}) {
+    this._hostname = config.hostname ?? 'localhost';
+    this._port = config.port ?? 8080;
+    this._protocol = config.protocol ?? 'http';
+    this._target = `${this._protocol}://${this._hostname}:${this._port}`;
+
+    this._logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'qcg'}/proxy`);
```
mariscalromeroalejandro:
```suggestion
    this._logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'qcg'}/qcdb-download-svc`);
```

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+   * the body of the answer from qcdb will then be streamed into our response.
+   * @param {string} objectId - id of ROOT object to retrieve from qcdb.
+   * @param {Express.Response} res - Optional Express response object if we want to stream qcdb's answer as our own.
+   * @returns {File|boolean} - ROOT file from qcdb, false if error and true if it responded itself.
+   */
+  async requestObject(objectId, res = undefined) {
+    this._logger.infoMessage(`Object ID Request: ${objectId}`);
+    const qcdbRequest = new Request(`${this._target}/download/${objectId}`, { method: 'GET' });
+    try {
+      const response = await fetch(qcdbRequest);
```
mariscalromeroalejandro: I would do:
```suggestion
      const response = await fetch(`${this._target}/download/${objectId}`);
```
for readability

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+        await this._pipelineAsync(response.body, res);
+        return true;
+      } else {
+        // We'll return the file from the response back.
+        const blob = await response.blob();
+        const file = new File([blob], filename ?? 'export.root', { type: contentType ?? 'application/root' });
+        return file;
+      }
+    } catch (error) {
+      this._logger.errorMessage(error);
```
mariscalromeroalejandro:
```suggestion
      this._logger.errorMessage(error?.message ?? error);
```

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+      // We will stream the data from QCDB's answer directly back to the user.
+      if (res != undefined) {
+        res.setHeader('Content-Type', contentType ?? 'application/root');
+        res.setHeader('Content-Disposition', contentDisposition ?? `attachment; filename="${filename}"`);
+        if (contentLength) {
+          res.setHeader('Content-Length', contentLength);
+        }
+
+        await this._pipelineAsync(response.body, res);
+        return true;
```
mariscalromeroalejandro: I would move this to a separate method like `_streamToResponse()`

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+      } else {
+        // We'll return the file from the response back.
+        const blob = await response.blob();
+        const file = new File([blob], filename ?? 'export.root', { type: contentType ?? 'application/root' });
+        return file;
```
mariscalromeroalejandro: and maybe this to: `_getFileFromResponse()`

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+   * @param {Express.Response} res - Express response object.
+   */
+  async getQcdbRootObjects(objectIds, res) {
+    const promises = [];
+    if (typeof objectIds === 'string') {
+      promises.push(this.requestObject(objectIds, res));
+    } else {
+      // Technically this can be stopped at the DTO layer but multiple id's will be implemented soon.
+      res.status(500).send('Option to retrieve more than 1 ROOT object not implemented yet.');
+    }
```
mariscalromeroalejandro, line 127: consider adding an early return

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+
+    try {
+      const files = await Promise.all(promises);
```
mariscalromeroalejandro: why promise all if we just have one object?

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,107 @@
...
+      if (files.filter((file) => file === false).length > 0) {
+        throw new Error('qcdb object request failed.');
+      }
+    } catch (error) {
+      this._logger.errorMessage(error.message);
```
mariscalromeroalejandro:
```suggestion
      this._logger.errorMessage(error?.message ?? error);
```

QualityControl/test/lib/controllers/ObjectController.test.js
```diff
@@ -183,4 +188,34 @@ export const objectControllerTestSuite = async () => {
...
+        token: 'some token',
+      };
+      const responseMsg = {
+        message: 'Invalid query parameters: "objectIds" is required',
+        status: 400,
+        title: 'Invalid Input'
+      };
+      await objectController.getDownloadObjects(reqMock, resMock);
+      ok(resMock.status.calledWith(400));
+      ok(resMock.json.calledWith(responseMsg));
```
mariscalromeroalejandro: see how this is done in the rest of the controllers: resMock.json.calledWith({message, status, title})

mariscalromeroalejandro (changes requested):
good PR! a few comments on clean code, error handling and testing but very good work in the overall

QualityControl/lib/services/QcdbDownload.service.js
Houwie7000: This method is the same that will be used for the multiple files root downloading. There this promise all makes sure all requests happen and finish.

QualityControl/lib/api.js
Houwie7000: Without it vscode does not know the type. It only knows the name but not the full type needed for intellisense.
This is an import only for vs code's language server

Houwie7000: You can test this by removing it and then typing http. below in the method. It will not give autocompletions while with the import in the JSDoc it will.

QualityControl/lib/controllers/ObjectController.js
Houwie7000: Can you elaborate why? The LayoutController also does its validation in the controller and not the service. I can also quickly respond with a corresponding message if it is happening in the controller.

Houwie7000:
Blocking merge due to potential bug, investigating...
False alarm, it seems that ROOT files sometimes are not rendered correctly in the https://root.cern.ch/js/7.8.2/
This could be a version issue but I tested the same result between the proxy and qcdb directly and there was no difference.

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -69,6 +78,33 @@ export class ObjectController {
     }
   }
 
+  /**
+   * Download ROOT objects using the QcdbProxy.
```
graduta: Perhaps you meant the `QcdbDownloadService`

QualityControl/lib/controllers/ObjectController.js
```diff
@@ -69,6 +78,33 @@ export class ObjectController {
...
+   */
+  async getDownloadObjects(req, res) {
+    let objectIds = undefined;
+    try {
+      const validated = await ObjectGetDownloadDTO.validateAsync(req.query);
+      ({ objectIds } = validated);
+      this._qcdbDownloadService.getQcdbRootObjects(objectIds, res);
+    } catch (e) {
+      let responseError = '';
+      if (e.isJoi) {
```
graduta: We prefer not to use single letter variables.
For example in `catch(e)`, we use instead `catch(error)` or `catch(err)` but first one is preferred

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,125 @@
...
+   * Constructor
+   * @param {object} config - application's {config.js}.ccdb, needed for Qcdb hostname+port
+   */
+  constructor(config = {}) {
+    this._hostname = config.hostname ?? 'localhost';
+    this._port = config.port ?? 8080;
+    this._protocol = config.protocol ?? 'http';
+    this._target = `${this._protocol}://${this._hostname}:${this._port}`;
+
+    this._logger = LogManager.getLogger(`${process.env.npm_config_log_label ?? 'qcg'}/qcdb-download-svc`);
```
graduta: For the label, while it may sound superficial, we do have analysis looking for `-service` and not `-srv`. When not sure, have a look at the other service classes.

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,125 @@
...
+    const promises = [];
+    if (typeof objectIds === 'string') {
+      promises.push(this.requestObject(objectIds, res));
+    } else {
+      // Technically this can be stopped at the DTO layer but multiple id's will be implemented soon.
+      throw new Error('Option to retrieve more than 1 ROOT object not implemented yet.');
+    }
+
+    const files = await Promise.all(promises);
+    // Request to QCDB failed to give a file back.
```
graduta, line 130: Even if it is just a comment, please try to keep one acronym style only. I have seen you have a mixture of `qcdb`, `Qcdb`, `QCDB`
For comments, I would reccomment the acronym in capital letters, while for code, you applied it correctly as `getQcdbRootObjects`

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,125 @@
...
+    if (files.filter((file) => file === false).length > 0) {
+      throw new Error('qcdb object request failed.');
```
graduta: We can discuss it as a requirement for your next "multi object download" task, but perhaps we do not want to cancel the entire download if one object fails. 
For example if a user wants an archive of 20 objects, it would be quite difficult to have to download 19 individual ones just because of one that failed

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,125 @@
...
+   * Stream the ROOT file contained in the QCDB response into our reponse back to the user.
+   * @param {Response} response - Response from QCDB
+   * @param {Express.Response} res - Outgoing response object we'll write our data into
+   * @returns {void}
+   */
+  async _streamToResponse(response, res) {
+    const contentLength = response.headers.get('Content-Length');
+    const contentType = response.headers.get('Content-Type');
+    const contentDisposition = response.headers.get('Content-Disposition');
+    const filename = contentDisposition?.slice(17, contentDisposition.length - 1);
```
graduta: We should not be using "magic numbers" (eslint reference) but instead extract this in a const variable which name will point to why is needed.

QualityControl/lib/api.js
graduta: Ok, normally we aspire to not limit our collaborators to one specific IDE, but I guess this might be shared by other tools as well. For now let's keep it

graduta (changes requested):
I left the less critical conventions of the projects for this review so that it is easier to discuss them.

QualityControl/lib/services/QcdbDownload.service.js
Houwie7000: Makes sense, this code can be rewritten for the multi download functionality to simply filter the 'bad' file out of the results. Right now it simply helps with detecting a bad response from QCDB

Houwie7000:
Somehow moving hardcoded text to constants broke my api tests.....

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,135 @@
...
+   * @returns {void}
+   */
+  async _streamToResponse(response, res) {
+    const contentLength = response.headers.get(CONTENT_LENGTH_HEADER);
+    const contentType = response.headers.get(CONTENT_TYPE_HEADER);
+    const contentDisposition = response.headers.get(CONTENT_DISPOSITION_HEADER);
+    const filename = contentDisposition?.slice(FILENAME_START_INDEX, contentDisposition.length - 1);
+    // We will stream the data from QCDB's answer directly back to the user.
+    res.setHeader(CONTENT_TYPE_HEADER, contentType ?? CONTENT_TYPE_DEFAULT);
+    res.setHeader(CONTENT_DISPOSITION_HEADER, contentDisposition ?? `${CONTENT_DISPOSITION_DEFAULT_PARTIAL}
```
Houwie7000:
```suggestion
    res.setHeader(CONTENT_DISPOSITION_HEADER, contentDisposition ?? `${CONTENT_DISPOSITION_DEFAULT_PARTIAL}"${filename}"`);
```
Turns out quickly responding to an eslint max length issue without my brain turned on might not haven been a good idea.

QualityControl/lib/services/QcdbDownload.service.js
```diff
@@ -0,0 +1,135 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { LogManager } from '@aliceo2/web-ui';
+import { pipeline } from 'node:stream';
+import { promisify } from 'node:util';
+
+const CONTENT_LENGTH_HEADER = 'Content-Length';
```
graduta, line 19: I should have mentioned earlier (so feel free to leave it for the next pr) but for such headers we have a constant `CcdbConstants.js` where I think it might be a better place to be stored

graduta (approved):

PR #3109, opened 2025-10-07
[OGUI-1781] Resolve tab layout styling issues, remove unnecessary filtering

mariscalromeroalejandro:
- [x] PR labels are selected

This PR includes the following updates and fixes:

- Fixed tab styles in the layout view: tabs no longer wrap into three rows due to limited space.
- Removed filtering options in layouts while in both edit and run modes.
- Applied additional style improvements across the layout view.
- Updated the run mode checkbox label from “Runs Mode” to “Run Mode.”

### DIFF
--- a/QualityControl/public/common/header.js
+++ b/QualityControl/public/common/header.js
@@ -38,7 +38,7 @@ export default (model) => {
     h('.flex-row.p2.items-center', { id: 'qcg-header' }, [
       commonHeader(model),
       centerCol || h('.flex-grow'),
-      rightCol || h('.w-33'),
+      rightCol || h('.w-25'),
     ]),
     filterSpecific(model),
   ]);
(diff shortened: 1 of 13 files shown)


QualityControl/public/layout/view/header.js
graduta: `toolbarEditMode` expects one parameter only

QualityControl/public/common/header.js
```diff
@@ -70,7 +70,7 @@ const filterSpecific = (model) => {
   const { page, filterModel, layout, object, objectViewModel } = model;
 
   switch (page) {
-    case 'layoutShow': return filtersPanel(filterModel, layout);
+    case 'layoutShow': return !layout.editEnabled && filtersPanel(filterModel, layout);
```
graduta, line 73: I think you should clear the filters when going into EDIT mode because otherwise the objects sidetree and loaded objects will still be based on the filter parameters.
Thus, if a user searches by run X that does not produce object B, then in EDIT mode the user would not be able to add object B to the layout

graduta (changes requested):
A small change in filters and should be good to go

graduta (approved):

PR #3113, opened 2025-10-09
[OGUI-1585] download buttons for single ROOT objects

Houwie7000:
- [x] PR labels are selected

Waiting on previous branch merge....
Commits will look duplicated until then


QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -47,10 +49,13 @@ export const objectInfoResizePanel = (model, tabObject) => {
         { style: 'right:0.1em; width: 35em;left: auto;' },
         objectRemoteData.isSuccess() && h('.p1', qcObjectInfoPanel(objectRemoteData.payload)),
       ),
-    ])),
+    ]),
+    objectRemoteData.isSuccess() && model.services.object.getDownloadQcdbObjectElement(objectRemoteData
+      .payload.id, { style: 'margin-right: .25rem;', id: 'dl-button' }),
```
mariscalromeroalejandro: This line is a bit hard to read as it's quite long. Consider formatting it

QualityControl/public/object/objectTreePage.js
```diff
@@ -96,15 +96,20 @@ const drawPlot = (model, object) => {
   const info = object;
   return h('', { style: 'height:100%; display: flex; flex-direction: column' }, [
     h('.resize-button.flex-row', [
```
mariscalromeroalejandro: this is also difficult to read, as the container class is 'resize-button', but it is a flew row with one button for download and another one for enter full screen. Consider separating responsibilities and try to make it cleaner

QualityControl/public/services/QCObject.service.js
```diff
@@ -169,6 +170,19 @@ export default class QCObjectService {
...
 
+  getDownloadQcdbObjectElement(objectId = undefined, elementOptions = {}) {
+    if (objectId == undefined || model.session.token == undefined) {
+      return;
+    }
+    const url = `/api/object/proxy/download/?token=${this.model.session.token}&objectIds=${objectId}`;
+    return h('a.btn', {
+      title: 'Download object',
+      target: '_blank',
+      href: url,
```
mariscalromeroalejandro: see how the`loader.get` is being used across the project

QualityControl/public/view.js
```diff
@@ -54,7 +54,7 @@ function page(model) {
     case 'layoutList': return LayoutListPage(model.layoutListModel);
     case 'layoutShow': return layoutViewPage(model);
     case 'objectTree': return objectTreePage(model);
-    case 'objectView': return ObjectViewPage(model.objectViewModel);
+    case 'objectView': return ObjectViewPage(model.objectViewModel, model.services.object);
```
mariscalromeroalejandro: please don't do that, use the service in the model, not in the view

QualityControl/public/object/objectTreePage.js
```diff
@@ -96,15 +96,20 @@ const drawPlot = (model, object) => {
...
-        'a.btn',
-        {
-          title: 'Open object plot in full screen',
-          href,
-          onclick: (e) => model.router.handleLinkEvent(e),
-        },
-        iconResizeBoth(),
-      )),
+      h('.p1.text-left', { style: 'padding-bottom: 0;' }, [
+        model.services.object.getDownloadQcdbObjectElement(object.id, {
```
mariscalromeroalejandro: the view should use the model, not the service. A service should not provide with a vnode

mariscalromeroalejandro (changes requested):

QualityControl/public/services/QCObject.service.js
Houwie7000: Loader.get is not applicable here since I need to download a file and not receive a javascript object

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -33,11 +33,13 @@ export const objectInfoResizePanel = (model, tabObject) => {
     .forEach(([key, value]) => {
       uri += `&${key}=${encodeURI(value)}`;
     });
-  return h('.text-right.resize-element.resize-button.flex-row', {
+  return h('.text-right.resize-element.item-action-row.flex-row', {
     style: 'display: none; padding: .25rem .25rem 0rem .25rem;',
   }, [
 
-    h('', { style: 'padding-bottom: 0;' }, h('.dropdown.mh1', { class: isSelectedOpen ? 'dropdown-open' : '' }, [
+    h('.dropdown', { class: isSelectedOpen ? 'dropdown-open' : '',
```
graduta, line 40: We should ignore the bad example I set above of setting the style like this and instead make use of classes which helps with readability and maintainability. 
Thus, in this case I would recommend you make use of the `.m1` class. Of course, this means a bigger margin would be introduced between elements, as both would need the `m1` class. For that we have the class `g1` with the gap CSS property which needs to be applied on the parent rather than the children.

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -47,10 +49,16 @@ export const objectInfoResizePanel = (model, tabObject) => {
         { style: 'right:0.1em; width: 35em;left: auto;' },
         objectRemoteData.isSuccess() && h('.p1', qcObjectInfoPanel(objectRemoteData.payload)),
       ),
-    ])),
+    ]),
+    objectRemoteData.isSuccess() &&
+    model.objectViewModel.getDownloadQcdbObjectElement(
```
graduta: Virtual nodes should be declared and used in views and not in the models

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -47,10 +49,16 @@ export const objectInfoResizePanel = (model, tabObject) => {
...
+      objectRemoteData.payload.id,
+      { style: 'margin-right: .25rem;', id: 'dl-button' },
+    ),
     h('a.btn', {
       title: 'Open object plot in full screen',
       href: uri,
+      style: 'margin-right: .25rem;',
```
graduta: Once you add the `.g1` class on the parent this is not needed anymore

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -47,10 +49,16 @@ export const objectInfoResizePanel = (model, tabObject) => {
         { style: 'right:0.1em; width: 35em;left: auto;' },
         objectRemoteData.isSuccess() && h('.p1', qcObjectInfoPanel(objectRemoteData.payload)),
       ),
-    ])),
+    ]),
+    objectRemoteData.isSuccess() &&
+    model.objectViewModel.getDownloadQcdbObjectElement(
+      objectRemoteData.payload.id,
+      { style: 'margin-right: .25rem;', id: 'dl-button' },
```
graduta: Once you add the `.g1` class on the parent this is not needed anymore

QualityControl/public/object/objectTreePage.js
```diff
@@ -95,16 +95,21 @@ const drawPlot = (model, object) => {
...
-      )),
+    h('.item-action-row.flex-row', [
+      h('.p1.text-left', { style: 'padding-bottom: 0;' }, [
+        model.objectViewModel.getDownloadQcdbObjectElement(object.id, {
+          style: 'margin-right: .25rem;', id: 'dl-button' }),
+        h(
+          'a.btn#fs-button',
+          {
+            title: 'Open object plot in full screen',
+            style: 'margin-right: .25rem;',
```
graduta: Once you add the `.g1` class on the parent this is not needed anymore

QualityControl/public/object/objectTreePage.js
```diff
@@ -95,16 +95,21 @@ const drawPlot = (model, object) => {
...
-          onclick: (e) => model.router.handleLinkEvent(e),
-        },
-        iconResizeBoth(),
-      )),
+    h('.item-action-row.flex-row', [
+      h('.p1.text-left', { style: 'padding-bottom: 0;' }, [
+        model.objectViewModel.getDownloadQcdbObjectElement(object.id, {
+          style: 'margin-right: .25rem;', id: 'dl-button' }),
+        h(
+          'a.btn#fs-button',
```
graduta: `fs-button` is not very descriptive, could you please improve the naming?
The same for the `dl-button`.

graduta (changes requested):

graduta (approved):

PR #3119, opened 2025-10-13
[OGUI-1790] extract download button to reusable element

Houwie7000:
- [x] PR labels are selected

Use this.model in the download button code and extract the download button to its own component.

### DIFF
--- a/QualityControl/public/common/downloadButton.js
+++ b/QualityControl/public/common/downloadButton.js
@@ -15,14 +15,10 @@
 import { h, iconDataTransferDownload } from '/js/src/index.js';
 
 /**
- * Download button for downloading an object from QCDB download service.
- * @param {ObjectViewModel} objectViewModel - which contains the download url.
- * @param {string} qcObjectId - QCDB object id.
+ * Download button.
+ * @param {object} options - vnode options.
  * @returns {vnode} - Download button element.
- * @import ObjectViewModel from '../pages/objectView/ObjectViewModel';
  */
-export default (objectViewModel, qcObjectId) => h('a.btn#download-button', {
-  title: 'Download object',
-  target: '_blank',
-  href: objectViewModel.getDownloadQcdbObjectUrl(qcObjectId),
+export default (options = {}) => h('a.btn#download-button', {
+  ...options,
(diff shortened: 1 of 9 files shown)


QualityControl/public/common/downloadButton.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h, iconDataTransferDownload } from '/js/src/index.js';
+
+/**
+ * @typedef {object} downloadButtonOptions
```
graduta: Typedef across the project has the convention of using class naming standards. That is starting with a capital letter

QualityControl/public/common/downloadButton.js
```diff
@@ -0,0 +1,38 @@
...
+ * @property {string} id - html id property.
+ * @property {string} class - html class property.
```
graduta: It would be good to specify that the class parameter accepts multiple classes as a string with classes separated by spaces in it. e.g. `button danger`

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -50,11 +51,11 @@ export const objectInfoResizePanel = (model, tabObject) => {
...
     ]),
     objectRemoteData.isSuccess() &&
-    h('a.btn#download-button', {
-      title: 'Download object',
-      target: '_blank',
+    downloadButton({
       href: model.objectViewModel.getDownloadQcdbObjectUrl(objectRemoteData.payload.id),
-    }, iconDataTransferDownload()),
+      title: 'Download object',
+      class: 'download-button',
```
graduta: I do not think there is a point in passing same class across all usages. What about using the same tactic as with the "target"?
Moreover, given that in a layout we have multiple such buttons, I think it is best that you pass an id similar to `#download-button-<object_id>` so that is unique per object

graduta (changes requested):

graduta (approved):

PR #3139, opened 2025-10-27
[OGUI-1783] Backend download post request 

Houwie7000:
- [x] PR labels are selected

Add one controller endpoint for handling a post request containing the json skelaton of the layout. This data can then be used to download the right objects/tabs in a later pull request containing the other endpoint (GET for the download).
This is needed because we need the data the user is seeing on their screen at the moment they click the download button. The domain models contain all data I need from the layout excluding the data in the query parameters of the GET request. This can be seen in DownloadConfigDomain.js

This pr does not contain the code needed to do the actual downloading of the objects/tabs.


Houwie7000:
Writing more tests and licenses....
(done)

Houwie7000:
The GitHub actions is back to the point where I need to run it three times for it to succeed... It has trouble with the runsmode tests.

Houwie7000:
look at: QualityControl/common/library/typedef/TabObject.js existing typedefs

PR #3146, opened 2025-10-31
[OGUI-1798] fix inconsistent runmode test + tweak docker build

Houwie7000:
- [x] PR labels are selected

* Make runmodetests more reliable, Fix Aliecs sync test race condition.
* Update docker compose files to improve layer caching and fix nodemon. Mariadb pinned to current LTS. Removed deprecated :cached instruction (mac only feature removed in 2021) and version attribute.

Nodemon's issue was due to the docker mount being incompatible with the way nodemon normally listens for changes using something like inotify. For a mounted folder it needs legacy polling (you can notice a small delay in detection).


QualityControl/docker-compose.yml
```diff
@@ -1,8 +1,7 @@
-version: '3.8'
-
 services:
   database:
-    image: "mariadb:10.5.9"
+    # Pinned to 11.8 LTS version, community support till 04-07-2028, enterprise: 15-10-2033.
+    image: "mariadb:11.8"
```
graduta: We are on purpose using mariadb 10.5 to have the same version as at P2.
Please revert the change

QualityControl/test/public/features/runMode.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @stylistic/js/max-len */
```
graduta: Nothing should be above the licence header. 
If you really need to disable an eslint rule, please do so after the header

QualityControl/docker-compose.dev.yml
```diff
@@ -12,7 +12,9 @@ services:
...
     container_name: qcg-application-dev
     volumes:
-      # Mount the current directory to /usr/src/app cached to improve performance
-      - .:/usr/src/app:cached 
+      # Mount the current directory to /usr/src/app
+      - .:/usr/src/app
       # Ignore local node_modules to avoid conflicts with the container's node_modules
       - /usr/src/app/node_modules
+      # Mount local node_modules to container
+      # - ./node_modules:/usr/src/app/node_modules
```
graduta: Commented code should be removed

QualityControl/docker-compose.dev.yml
```diff
@@ -12,7 +12,9 @@ services:
       - "8080:8080"
     container_name: qcg-application-dev
     volumes:
-      # Mount the current directory to /usr/src/app cached to improve performance
-      - .:/usr/src/app:cached 
+      # Mount the current directory to /usr/src/app
+      - .:/usr/src/app
```
graduta: As far as I know, the cached option is better for development. What did your research conclude?

QualityControl/docker-compose.yml
```diff
@@ -1,8 +1,7 @@
-version: '3.8'
```
graduta, line 1: Is there a reason for removing the version? While I understand newer version made this line optional, it is still good practice to keep it for compatibility and clarity

graduta (changes requested):

PR #3162, opened 2025-11-16
[OGUI-1814] Add minification to the object view details page

hehoon:
- [x] PR labels are selected

PR which
- Allows for the minification of the object details in objectView page.
- The button is in line with the timestamp dropdown and grouped together with the download button.

### DIFF
--- a/QualityControl/public/common/visibilityButton.js
+++ b/QualityControl/public/common/visibilityButton.js
@@ -12,20 +12,7 @@
  * or submit itself to any jurisdiction.
  */
 
-import { h, icon, iconEye } from '/js/src/index.js';
-
-const iconNoEye = () => icon([
-  // 'path' is taken from iconEye
-  // eslint-disable-next-line @stylistic/js/max-len
-  h('path', { d: 'M4.031 1c-2.53 0-4.031 3-4.031 3s1.501 3 4.031 3c2.47 0 3.969-3 3.969-3s-1.499-3-3.969-3zm-.031 1c1.1 0 2 .9 2 2s-.9 2-2 2-2-.9-2-2 .9-2 2-2zm0 1c-.55 0-1 .45-1 1s.45 1 1 1 1-.45 1-1c0-.1-.032-.191-.063-.281-.08.16-.237.281-.438.281-.28 0-.5-.22-.5-.5 0-.2.121-.357.281-.438-.09-.03-.181-.063-.281-.063z' }),
-  h('line', {
-    x1: '0',
-    y1: '0',
-    x2: '8',
-    y2: '8',
-    stroke: 'currentColor',
-  }),
-]);
+import { h, iconChevronLeft, iconChevronRight } from '/js/src/index.js';
 
 /**
  * @typedef {object} VisibilityToggleButtonOptions
@@ -62,5 +49,5 @@ export function visibilityToggleButton(options = {}, onclick) {
   return h('a.btn', {
     ...mergedOptions,
     onclick,
-  }, isVisible ? iconEye() : iconNoEye());
+  }, isVisible ? iconChevronRight() : iconChevronLeft());
 }
--- a/QualityControl/public/common/visibilityButton.js
+++ b/QualityControl/public/common/visibilityButton.js
@@ -1,0 +1,66 @@
+/**
+ * @license
+ * Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+ * See http://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h, icon, iconEye } from '/js/src/index.js';
+
+const iconNoEye = () => icon([
+  // 'path' is taken from iconEye
+  // eslint-disable-next-line @stylistic/js/max-len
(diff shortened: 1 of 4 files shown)


QualityControl/public/common/visibilityButton.js
```diff
@@ -0,0 +1,66 @@
...
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h, icon, iconEye } from '/js/src/index.js';
+
+const iconNoEye = () => icon([
+  // 'path' is taken from iconEye
+  // eslint-disable-next-line @stylistic/js/max-len
+  h('path', { d: 'M4.031 1c-2.53 0-4.031 3-4.031 3s1.501 3 4.031 3c2.47 0 3.969-3 3.969-3s-1.499-3-3.969-3zm-.031 1c1.1 0 2 .9 2 2s-.9 2-2 2-2-.9-2-2 .9-2 2-2zm0 1c-.55 0-1 .45-1 1s.45 1 1 1 1-.45 1-1c0-.1-.032-.191-.063-.281-.08.16-.237.281-.438.281-.28 0-.5-.22-.5-.5 0-.2.121-.357.281-.438-.09-.03-.181-.063-.281-.063z' }),
```
graduta: Hi @hehoon , the icons are stored as part of the framework, more exactly in `Frontend/js/src/icons.js`
If you wish this new icon, you have to raise a PR for the framework with your addition to that file. 

In order to fast-track your ticket, I would recommend to use an existing icon. You can see all available here: https://aliceo2group.github.io/WebUi/Framework/docs/reference/frontend-css.html#icons 

Perhaps something like `iconChevronLeft`/`Right`

QualityControl/public/pages/objectView/ObjectViewModel.js
```diff
@@ -43,6 +43,11 @@ export default class ObjectViewModel extends BaseViewModel {
     this.drawingOptions = [];
     this.displayHints = [];
     this.ignoreDefaults = false;
+
+    /**
+     * Tracks whether the object information panel is currently visible.
+     */
+    this.objectInfoVisible = true;
```
graduta: I think here it would be better to use the convention of defining private variables with `_` in front of the name and use the JS `get/set` methods. 
See :
- https://github.com/AliceO2Group/WebUi/blob/dev/QualityControl/public/common/filters/model/FilterModel.js#L38 
- https://github.com/AliceO2Group/WebUi/blob/dev/QualityControl/public/common/filters/model/FilterModel.js#L292

QualityControl/public/pages/objectView/ObjectViewModel.js
```diff
@@ -127,6 +132,38 @@ export default class ObjectViewModel extends BaseViewModel {
...
+  }
+
+  /**
+   * Toggle the display state of object information.
+   * If currently visible, it becomes hidden; if hidden, it becomes visible.
+   * @example
+   * objectViewModel.toggleObjectInfoVisible();
+   */
+  toggleObjectInfoVisible() {
+    this.setObjectInfoVisible(!this.getObjectInfoVisible());
```
graduta: Within the class where the variable is defined as private, we should make use of the variable itself rather than the getter

QualityControl/public/pages/objectView/ObjectViewModel.js
```diff
@@ -127,6 +132,38 @@ export default class ObjectViewModel extends BaseViewModel {
...
+   */
+  setObjectInfoVisible(objectInfoVisible) {
+    this.objectInfoVisible = objectInfoVisible;
+    this.notify();
+  }
+
+  /**
+   * Toggle the display state of object information.
+   * If currently visible, it becomes hidden; if hidden, it becomes visible.
+   * @example
```
graduta: While the `example` addition for the visibility button is a very good one, I believe for such methods (`getObjectInfoVisible`,`setObjectInfoVisible`, `toggleObjectInfoVisible`) is a bit too much. Could you please remove these examples?

QualityControl/public/pages/objectView/ObjectViewModel.js
```diff
@@ -127,6 +132,38 @@ export default class ObjectViewModel extends BaseViewModel {
...
+  }
+
+  /**
+   * Set the display state of object information.
+   * Notifies any observers/listeners about the change.
+   * @param {boolean} objectInfoVisible - `true` to show object information, `false` to hide.
+   * @example
+   * objectViewModel.setObjectInfoVisible(true);
+   */
+  setObjectInfoVisible(objectInfoVisible) {
```
graduta: Because you have the toggle method, a setter is not even needed in this case especially once you will make use of the private variable within the class

QualityControl/public/pages/objectView/ObjectViewPage.js
```diff
@@ -60,13 +58,38 @@ const objectPlotAndInfo = (objectViewModel) =>
...
+          h(
+            isObjectInfoVisible ? '.w-30.scroll-y' : '.scroll-off',
+            {
+              style: isObjectInfoVisible ? '' : 'width: 0;',
+            },
+            [
+              h('h3.text-center', 'Object information'),
+              qcObjectInfoPanel(qcObject, { gap: '.5em' }),
+            ],
+          ),
```
graduta: I believe this bit of code became difficult to read. Moreover, in general, you want to play with the creation of the elements rather than the change of sizes, especially with JSROOT plots where you need to rerender them.

In this case, I believe it would be better to not create the information panel if the setting is `false` rather  than reducing it to `0`

Have a look at my proposal below and let me know what you think:

```js
h('.w-100.flex-row.g2.m2', { style: 'height: 0;flex-grow:1' }, [
  h('.flex-grow', {
    // Key change forces redraw when toggling info panel
    key: isObjectInfoVisible ? 'objectPlotWithoutInfoPanel' : 'objectPlotWithInfoPanel',
  }, draw(qcObject, {}, drawingOptions)),
  isObjectInfoVisible && h('.scroll-y.w-30', {
    key: 'objectInfoPanel',
  }, [
    h('h3.text-center', 'Object information'),
    qcObjectInfoPanel(qcObject, { gap: '.5em' }),
  ]),
]),
```

graduta (changes requested):
Good job! With a few changes, the PR will be ready to be merged

QualityControl/public/pages/objectView/ObjectViewPage.js
hehoon: Your proposal makes sense. And avoiding the creation of the element when the panel is hidden is a cleaner approach!

QualityControl/public/common/visibilityButton.js
hehoon: I'll implement this with `iconChrevronLeft` and `iconChrevronRight`, thank you!

graduta (approved):

PR #3181, opened 2025-11-19
[OGUI-1816] Object Details Panel Improvements and Copy of values

AlexJanson:
- [x] PR labels are selected

The panel displaying the details of the object such as:
- each value should on hover:
  - display the cursor mouse pointer to inform the user that clicking will trigger an action. A "title" property should be added which will be displayed on hover with text Copy!
  - onclick - the value should be copied
  - on double click - native browser behaviour should be left
- each key should have a label displayed and not the key
	- id and tag to be combined in something like `ID (etag)`
- the following values and or keys should popout 	
  - run number
  - run type
  - path
  - qc version


QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -13,34 +13,93 @@
...
+};
+
+/**
+ * Transforms a camelCase string into human-readable Title Case format,
+ * inserts a space before every uppercase letter and ensures the first
+ * character is capatilized
+ * @param {string} key - key of the object info
+ * @returns {string} - formatted label for the given key
+ */
+const defaultKeyTransform = (key) => {
```
graduta: Looking at the docs, I believe this method can have a better naming to suggest what it does and be placed in a util file

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -13,34 +13,93 @@
...
+      ...hasValue && infoRowAttributes(formattedKey, formattedValue),
+      style: 'cursor: pointer; user-select: text;',
+    }, formattedValue),
+  ]);
+};
+
+/**
+ * Transforms a camelCase string into human-readable Title Case format,
+ * inserts a space before every uppercase letter and ensures the first
+ * character is capatilized
```
graduta: Small typo and to keep the same wording as the previous line perhaps: `character is uppercase`?

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -13,34 +13,93 @@
...
+ * @param {string} key - key of the object info
+ * @returns {string} - formatted label for the given key
```
graduta: `param` and `returns` docs should also be improved the moment the function is moved

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -13,34 +13,93 @@
...
+  const formattedValue = infoPretty(key, value);
+  const formattedKey = getUILabel(key);
+
+  const hasValue = value != null && value !== '' && (!Array.isArray(value) || value.length);
+
+  return h(`.flex-row.g2.info-row${highlightedClasses}`, [
+    h('b.w-25.w-wrapped', formattedKey),
+    h('.w-75', {
+      ...hasValue && infoRowAttributes(formattedKey, formattedValue),
+      style: 'cursor: pointer; user-select: text;',
```
graduta: Is the `user-select` needed in this case? I believe by default the text is selectable
Moreover, could please use classes rather than individual styling (refering to the `cursor: pointer` in this case

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -13,34 +13,93 @@
...
+  const highlightedClasses = HIGHLIGHTED_FIELDS.includes(key) ? '.highlighted' : '';
+  const formattedValue = infoPretty(key, value);
+  const formattedKey = getUILabel(key);
+
+  const hasValue = value != null && value !== '' && (!Array.isArray(value) || value.length);
+
+  return h(`.flex-row.g2.info-row${highlightedClasses}`, [
+    h('b.w-25.w-wrapped', formattedKey),
+    h('.w-75', {
+      ...hasValue && infoRowAttributes(formattedKey, formattedValue),
```
graduta: `hasValues` is a boolean. If so, there is no need to destructuring

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -58,3 +117,32 @@ const infoPretty = (key, value) => {
   }
   return h('', value);
 };
+
+/**
+ * Default function to configure the info row vnode attributes
+ * @param {Notification} notification - Notification API
```
graduta: Notification API referes to the native browser notification system
In your case, this is the Notification from WebUI framework, right?

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -58,3 +117,32 @@ const infoPretty = (key, value) => {
...
+  (key, value) => ({
+    onclick: async (e) => {
+      // to allowing the default behaviour for clicking multiple times
+      const clickCount = e.detail;
+      if (clickCount === 1) {
+        if (!isContextSecure()) {
+          return;
+        }
+
+        notification.show('Value has been successfully copied to clipboard', 'success', 1500);
```
graduta: The process of copying the value happens only on L141.
Should the notification not be placed after that line in the try catch?

QualityControl/public/common/object/objectInfoCard.js
```diff
@@ -58,3 +117,32 @@ const infoPretty = (key, value) => {
...
+        if (typeof value !== 'string') {
```
graduta: To avoid this, I would propose that you change the default return statement from `infoPretty` to `JSON.stringify(value)` so that it always returns a string

graduta (changes requested):

graduta (dismissed):

graduta (approved):

PR #3182, opened 2025-11-19
[OGUI-1819] JSROOT invalid redraw

hehoon:
- [x] PR labels are selected

* Fixes JSROOT plots not resizing due to `JSROOT.resize` being `undefined`. Uses `JSROOT.redraw` instead.
* Also takes transitions/animations into account, and only redraws after the transition/animation is completed.
* Merged the logic from `objectDraw.js` to `draw.js` (and removed `objectDraw.js`).


generatedkeys (dismissed):

QualityControl/public/common/object/draw.js
```diff
@@ -49,19 +51,38 @@ const rootPlotPanel = (object, options, drawingOptions) => {
   options = { ...defaultOptions, ...options };
 
   const attributes = {
-    key: root.name, // Completely re-create this div if the chart is not the same at all
-    id: object.etag,
+    key: name, // Completely re-create this div if the chart is not the same at all
```
graduta: I believe we should be using the `etag` in this case because the name is not unique

QualityControl/public/common/object/draw.js
```diff
@@ -87,4 +108,133 @@ function drawOnCreate(id, root, drawingOptions) {
...
+const fingerprintResize = (width, height) =>
+  `${width}:${height}`;
+
+/**
+ * Generates the final drawing options for JSROOT
+ * @param {object} root - root object in JSON representation
+ * @param {string[]} drawingOptions - list of options to be used for drawing object
+ * @returns {string} The final drawing options formatted as a string
+ */
+const getDrawingOptions = (root, drawingOptions) =>
```
graduta: Given that `generateDrawingOptionList` is within the utils part, I would say this one should go there as well

QualityControl/public/object/objectDraw.js
graduta, line 1: Given that there is no other codebase using this and you have successfully refactors the usage, this file should be removed entirely rather than just deprecated

graduta (changes requested):
The UX is definitely an improvement and on the right path but at the same time we have to make sure existing compatibility is not broken:
- when editing a layout and trying to add a new object via (double click), does not work anymore)
+ the ones we already discussed

QualityControl/public/common/object/draw.js
hehoon: In the commit I only refactored this code, but I agree with you that it should indeed use `etag` instead of `name`

QualityControl/public/object/objectDraw.js
hehoon: I wasn't entirely sure if other branches were still using this file. Since you have a better overview, I'll go ahead and remove it.

graduta (approved):

PR #3183, opened 2025-11-19
[OGUI-1752] TanStack auth setup

sKrzysieK:
TanStack Query endpoint auth setup

- [x] PR labels are selected


Configuration/webapp/app/components/user-section/UserSection.tsx
```diff
@@ -65,9 +61,12 @@ export const UserSection: FC<UserSectionProps> = ({ userName }) => {
         onClose={handleClose}
         className="user-section__menu"
       >
-        <MenuItem onClick={displayProfileData}>Profile</MenuItem>
-        <MenuItem onClick={handleClose}>My account</MenuItem>
-        <MenuItem onClick={handleClose}>Logout</MenuItem>
+        <section style={{ padding: 10 }}>
+          <h5>Welcome, {userName}!</h5>
```
Deaponn: **suggestion**: Use MUI alternatives of these tags. I know about [Typography](https://mui.com/material-ui/api/typography/) but I'm not sure of `<section>` alternative

Configuration/webapp/app/hooks/useAuth.ts
```diff
@@ -0,0 +1,38 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { useEffect, useState } from 'react';
+import { getSessionData } from '~/services/session';
+
+export interface Session {
+  personid: string;
```
Deaponn, line 19: **suggestion**: I think it should be named using camelCase

Configuration/webapp/app/api/axiosInstance.ts
```diff
@@ -24,4 +25,13 @@ const axiosInstance = axios.create({
   withCredentials: false,
 });
 
+axiosInstance.interceptors.request.use(async (config) => {
+  const { token } = await getSessionData();
```
Deaponn, line 30: **question (blocking)**: Not sure if this interceptor is ran once, or for every request?
We need to take into account the possibility that the token will expire. If this happens, we should re-fetch the session data.
Since I am not sure, that comment is blocking

Configuration/webapp/app/hooks/useAuth.ts
sKrzysieK: this is the name of session parameter obtained from session service and I believe it is written exactly like this

Deaponn: Oh, sorry for that then. My bad

graduta (approved):

PR #3201, opened 2025-11-30
[OGUI-1830] Add a unique error message for objects in ccdb not found due to filters

hehoon:
- [x] PR labels are selected

* Add unique errors for objects in ccdb that are not found due to having filters


QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -176,7 +176,7 @@ export class CcdbService {
    * ```
    * @param {CcdbObjectIdentification} partialIdentification - fields such as path, validFrom, etc.
    * @returns {Promise.<CcdbObjectIdentification>} - returns object full identification
-   * @throws {Error}
+   * @throws {Error} throws if the object cannot be found
```
graduta: Perhaps it should be `NotFoundError` when the object is not found? More specifically when CCDB returns 0 objects

QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -185,18 +185,32 @@ export class CcdbService {
...
-        [CCDB_RESPONSE_BODY_KEYS.ID]: qcObject[CCDB_RESPONSE_BODY_KEYS.ID],
-      };
-    } else {
-      throw new Error(`Object: ${url} could not be found`);
+    let result = null;
+    try {
+      result = await httpGetJson(this._hostname, this._port, url, { headers });
+    } catch {
+      const errorMessage = this._buildFilterErrorMessage(
+        `Object at url '${url}' and path '${partialIdentification.path}' could not be found.`,
```
graduta: At this point we do not know that the actual error is that object could not be found. It can be a timeout, a bad service, etc.

QualityControl/lib/services/ccdb/CcdbService.js
```diff
@@ -185,18 +185,32 @@ export class CcdbService {
...
+      );
+      throw new Error(errorMessage);
+    }
+
+    if (!result?.objects?.length) {
+      const errorMessage = this._buildFilterErrorMessage(
+        `Object at url '${url}' and path '${partialIdentification.path}' could not be found.`,
+        partialIdentification.filters,
+      );
+      throw new Error(errorMessage);
```
graduta: Only at this point we know that based on the filters, object could not be found and a `NotFoundError` should be thrown

graduta (changes requested):

graduta (approved):

PR #3209, opened 2025-12-02
[OGUI-1820] Header of pages should efficiently use remaining space

AlexJanson:
- [x] PR labels are selected

The QCG pages are making use of the header for displaying filters or tabs (e.g. layoutShow/objectTree) but in some other cases it is being used for nothing (layoutList)

Find a clever way of efficiently splitting the space of the header between the 3 potential components (left, center right)

Moreover, for the layoutShow, display the tabs list with a scrollable component instead of 2-3 lines of list of tab


graduta (changes requested):
While this is definitely an improvement, I think it can be better. 
Could you please try for the next meeting to come up with a proposal for`tabs` slider to be placed outside of the header, one level below (between header and filters)
Try to make it in such a way that the 3 elements are nicely cascading and are delimited in a nice UI manner.
Moreover, refrain from using hard values for margins and paddings especially in `px`. Try to use the existing defined classes in the framework `bootstrap.css` such as .p1, .p2, m1, m2 etc. Or the variables defined in the CSS files

QualityControl/public/app.css
```diff
@@ -155,3 +155,36 @@
 .cursor-pointer {
   cursor: pointer;
 }
+
+.header-layout {
+  & > :not(:last-child) {
+    margin-right: 8px;
```
graduta: Do we need this margin? To me it looks ok without it
But if you really want spacing, you should look at using the `g1`, `g2`, etc. classes which adds spacing between componets and not at the  end

QualityControl/public/app.css
```diff
@@ -155,3 +155,36 @@
...
+  &::after,
+  &::before {
+    content: '';
+    flex: 1;
+  }
+}
+
+.header-layout-container {
+  display: flex;
+  flex-wrap: nowrap;
```
graduta: I do not think there is a need for `nowrap` as this is the default value of the `flex-wrap` property

QualityControl/public/app.css
```diff
@@ -155,3 +155,36 @@
...
+  &.edit {
+    display: grid;
+    grid-template-columns: auto 1fr auto;
+  }
+
+  &::after,
+  &::before {
+    content: '';
+    flex: 1;
+  }
```
graduta, line 173: Instead of this, you should simply make use in `header-layout` class  of the property: 
`justify-content: center`

graduta (changes requested):
I believe the tabs group look better now that they have their own row.
It would be great if you could look at simplyfing the CSS and please bare in mind the issue when EDIT/SAVE actions are triggered as it is duplicating the buttons

QualityControl/public/app.css
AlexJanson: The issue I've run into using only `justify-content: center` is that the scrollbar does not manage to reach the first element.

image

I've found [this](https://stackoverflow.com/questions/62288432/problem-when-using-with-overflow-x-scroll-and-justify-content-center) stack overflow talking about the same problem.

graduta (approved):

PR #3211, opened 2025-12-02
[OGUI-1803] Add Restrictions for arrays

Deaponn:
This PR introduces calculating the Restrictions for arrays held in configuration objects.
The new data is calculated on the backend and displayed on the frontend.

- [x] PR labels are selected


Configuration/webapp/app/components/form/components/widgets/ArrayWidget.tsx
```diff
@@ -0,0 +1,60 @@
...
+  itemsRestrictions: ArrayRestrictions;
+}
+
+export const ArrayWidget = ({
+  sectionTitle,
+  items,
+  itemsRestrictions,
+}: ArrayWidgetProps): ReactElement => {
+  const [viewForm, setViewForm] = useState<boolean>(true);
+  const [arrayRestrictions] = itemsRestrictions; // [arrayRestrictions, objectBlueprint, arrayBlueprint]
```
sKrzysieK, line 41: If we use `itemsRestrictions` prop only to destruct it and get its first item, can we just pass this `itemsRestrictions[0]` to component directly?

Deaponn: For now, yes. But in the future, when we implement adding new items to the arrays, we will need all three values held in the `itemsRestrictions` array, being an `objectBlueprint` and an `arrayBlueprint`. So I figured it would be okay to do it this way and left a comment on how the target deconstruction should look like.

piechnikk (dismissed):

sKrzysieK (dismissed):

sKrzysieK (approved):

piechnikk (approved):

PR #3214, opened 2025-12-03
[OGUI-1738] Difference Highlighting

sKrzysieK:
Implemented difference highlighting by displaying default value below edited fields and introducing a button to reset to default on the right side of the input

- [x] PR labels are selected


Configuration/webapp/app/components/form/components/widgets/FormTextInput.tsx
```diff
@@ -27,26 +30,46 @@ interface FormTextInputProps extends Omit<WidgetProps, 'type' | 'value'> {
...
+        type={type}
         label={label}
         {...field}
         error={Boolean(error)}
         color={isDirty ? 'secondary' : 'primary'}
         focused={isDirty}
-        helperText={error?.message}
+        slotProps={{
+          input: {
+            endAdornment: isDirty ? (
```
Deaponn, line 57: **question**: have you considered putting the `DefaultValueSection` component in here, to the left of the `RemoveButton`? That way we can use up some of the blank space in these wide inputs and prevent content jumping when the `helperText` appears and takes up some space

Configuration/webapp/app/components/form/components/DefaultValueSection.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { Box, Typography } from '@mui/material';
+
+export const DefaultValueSection = ({
```
Deaponn: **question**: have you considered naming this component more in terms like `OldValueSection`? My line of thought is that some values already present in the configuration are not necessary default, they just were in this place before the user opened it

Deaponn (commented):
Looks good to me, just two quick questions about UI and naming

Configuration/webapp/app/components/form/components/widgets/FormTextInput.tsx
sKrzysieK: I have, but I strongly believe it is not a most UI friendly solution - input component is designed to hold only current value and adding any other text might be misleading for a user. 
Moreover, I think we should reserve the input's left adornment for buttons (e.g. present `Reset to default` button and `Copy value` button in the future). 
However, I definitely agree with jumping content issue and I will take a look at it.

Configuration/webapp/app/components/form/components/DefaultValueSection.tsx
sKrzysieK: It definitely does make sense, thank you for pointing this out, I will rename this.

piechnikk (dismissed):

Deaponn (dismissed):

Deaponn (approved):

piechnikk (approved):

PR #3218, opened 2025-12-04
[OGUI-1838] Change raw view to be an editable modal

piechnikk:
- [x] PR labels are selected

Replaced the previous raw value display with an editable modal


Configuration/webapp/app/components/form/raw-view/RawEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Box } from '@mui/material';
+import { Spinner } from '~/ui/spinner';
+import type { FormItem } from '../Form';
+
+interface RawEditorProps {
+  data: FormItem;
+}
+
+export const RawEditor: FC<RawEditorProps> = ({ data }) => {
+  const formattedJson = JSON.stringify(data, null, 2);
```
sKrzysieK: It works fine but to make sure this `formattedJSON` is not recreated on every single rerender you could consider wrapping it in `useMemo` with `[data]` dependency array

Configuration/webapp/app/components/form/Form.tsx
```diff
@@ -64,21 +65,22 @@ export const Form: FC<FormProps> = ({ sectionTitle, items, itemsRestrictions })
...
-        )}
-      </AccordionDetails>
-    </Accordion>
+        </AccordionDetails>
+      </Accordion>
+
+      <RawViewModal
+        open={isRawModalOpen}
+        onClose={() => setIsRawModalOpen(false)}
+        title={`${sectionTitle}`}
```
sKrzysieK: I believe you can skip the template literal here as below:

```suggestion
        title={sectionTitle}
```

Configuration/webapp/app/components/form/Form.tsx
```diff
@@ -46,7 +47,7 @@ function isFormRestrictions(obj: FormRestrictions[string]): obj is FormRestricti
 }
 
 export const Form: FC<FormProps> = ({ sectionTitle, items, itemsRestrictions }) => {
-  const [viewForm, setViewForm] = useState<boolean>(true);
+  const [isRawModalOpen, setIsRawModalOpen] = useState<boolean>(false);
```
sKrzysieK, line 71: FYI: we do not need to pass `boolean` as generic param here as it will be infered by typescript automatically
You can leave it for now as it is not causing any harm

sKrzysieK (dismissed):

Configuration/webapp/app/components/form/AccordionHeader.tsx
```diff
@@ -50,7 +47,7 @@ export const AccordionHeader: FC<AccordionHeaderProps> = ({
     >
       <Typography sx={{ marginRight: 'auto', alignContent: 'center' }}>{title}</Typography>
       <IconButton onClick={viewFormToggleCallback}>
```
Deaponn: **question**: what is your opinion about renaming this callback to `showRawEditorModal` or similar? `viewFormToggle` is no longer accurate in my opinion

Configuration/webapp/app/components/form/Form.tsx
```diff
@@ -64,21 +65,24 @@ export const Form: FC<FormProps> = ({ sectionTitle, items, itemsRestrictions })
...
-          <Typography component="pre">{JSON.stringify(items, null, 2)}</Typography>
-        )}
-      </AccordionDetails>
-    </Accordion>
+        </AccordionDetails>
+      </Accordion>
+
+      {isRawModalOpen && (
+        <RawViewModal
+          open={isRawModalOpen}
```
Deaponn: **question**: why do we need `open` prop on `RawViewModal` component? It is always passed as `true` because the line 78 anyway

Configuration/webapp/app/components/form/raw-view/RawViewModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { useEffect, useMemo, useState, type FC } from 'react';
+import Dialog from '@mui/material/Dialog';
+import DialogTitle from '@mui/material/DialogTitle';
+import DialogContent from '@mui/material/DialogContent';
+import IconButton from '@mui/material/IconButton';
+import Box from '@mui/material/Box';
+import Tooltip from '@mui/material/Tooltip';
+import CloseIcon from '@mui/icons-material/Close';
+import ContentCopyIcon from '@mui/icons-material/ContentCopy';
+// import { RawViewer } from './RawViewer';
```
Deaponn: please remove commented code

Configuration/webapp/app/components/form/AccordionHeader.tsx
piechnikk: I think it's a good idea, I'll change this

Configuration/webapp/app/components/form/Form.tsx
piechnikk: I think it's a leftover from before I added conditional rendering, I'll change this

Deaponn (dismissed):

graduta (approved):

PR #3226, opened 2025-12-09
[OGUI-1833] Handle multiple forms navigation

sKrzysieK:
Handle navigation between forms properly, which included seamless value reset and protecting user from leaving configuration with unsaved changes.

- [x] PR labels are selected


Configuration/webapp/app/hooks/useUnsavedChangesBlocker.tsx
```diff
@@ -0,0 +1,134 @@
...
+    shouldBlockRef.current = false;
+
+    if (targetPath) {
+      void navigate(targetPath);
+    }
+  }, [navigate]);
+
+  const handleSaveAndProceed = useCallback(async () => {
+    setShowModal(false);
+    await onSave();
```
Deaponn, line 112: **question**: what if the `onSave` callback fails? For example backend returns `4xx` or `5xx`, is it the responsibility of `onSave` function to handle such cases, so the user won't loose his modifications?

Deaponn (dismissed):
lgtm, I just have one conceptual question

Configuration/webapp/app/hooks/useUnsavedChangesBlocker.tsx
sKrzysieK: good point, I will introduce notifying user about save errors in the next PR where I will add saving configurations

piechnikk (dismissed):

Deaponn (dismissed):

piechnikk (approved):

PR #3229, opened 2025-12-11
[OGUI-1840 & OGUI-1841] Refactor the response object of `retrieveRunStatus`

hehoon:
- [x] PR labels are selected

* The method `retrieveRunStatus` is renamed to `retrieveRunInformation`
* The method now returns an object with established run information:
  * `runStatus`: this is the field created right now by QCG and should be kept to ensure front-end compatibility
  * time at which the run has started - `startTime`
  * time at which the run has run ended - `endTime`
  * the run belongs to a partition also known as environment: `environmentId`
  * the run also is defined by multiple properties. Depending on which oneas are used the run has a definition: `definition`
  * the run has a quality that decides if it should be stored or not for long time: `runQuality`
  * run normally runs only during an LHC beam mode: `lhcBeamMode`
  * A run has multiple detectors taking data, thus we should also get the list of detectors and qualityies: `detectorQualities`
* All methods making use of this have been updated accordingly:
  * Methods in `FilterService`
  * Methods in `FilterController`

### DIFF
--- a/QualityControl/lib/services/BookkeepingService.js
+++ b/QualityControl/lib/services/BookkeepingService.js
@@ -130,21 +130,7 @@ export class BookkeepingService {
   /**
    * Retrieves the information of a specific run from the Bookkeeping service
    * @param {number} runNumber - The run number to check the status for
-   * @returns {Promise<object>} - Returns a promise that resolves to the run information:
-   * - runStatus: A custom field created here for the front-end:
-   *   - `RunStatus.ONGOING` if the run is ongoing
-   *   - `RunStatus.ENDED` if the run has completed (has timeO2End)
-   *   - `RunStatus.NOT_FOUND` if the data cannot be found
-   *   - `RunStatus.UNKNOWN` if there was an error or data is not available
-   * - time at which the run has started - `startTime`
-   * - time at which the run has run ended - `endTime`
-   * - the run belongs to a partition also known as environment: `environmentId`
-   * - the run also is defined by multiple properties.
-   *   Depending on which oneas are used the run has a definition: `definition`
-   * - the run has a quality that decides if it should be stored or not for long time: `runQuality`
-   * - run normally runs only during an LHC beam mode: `lhcBeamMode`
-   * - A run has multiple detectors taking data,
-   *   thus we should also get the list of detectors and qualityies: `detectorQualities`
+   * @returns {Promise<RunInformation|WrappedRunStatus>} - Returns a promise that resolves to the run information
    */
   async retrieveRunInformation(runNumber) {
     if (!this.active) {
--- a/QualityControl/lib/services/BookkeepingService.js
+++ b/QualityControl/lib/services/BookkeepingService.js
@@ -155,7 +155,7 @@ export class BookkeepingService {
         definition,
         runQuality,
         lhcBeamMode,
-        detectorQualities,
+        detectorsQualities = [],
         timeO2End,
       } = data;
       const runStatus = timeO2End ? RunStatus.ENDED : RunStatus.ONGOING;
@@ -167,7 +167,7 @@ export class BookkeepingService {
         definition,
         runQuality,
         lhcBeamMode,
-        detectorQualities,
+        detectorsQualities,
         ...wrapRunStatus(runStatus),
       };
     } catch (error) {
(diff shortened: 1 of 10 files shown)


hehoon:
The failing test is resolved in: https://github.com/AliceO2Group/WebUi/pull/3230

QualityControl/lib/services/BookkeepingService.js
```diff
@@ -150,18 +161,51 @@ export class BookkeepingService {
...
 
+  /**
+   * Wraps a given run status value into a standardized result object.
+   * Use this helper when you need to return only a `runStatus` field without any
+   * additional payload. This ensures that all callers receive a consistent
+   * object shape, matching the structure returned by `retrieveRunInformation`.
+   * @param {RunStatus} runStatus The run status to wrap. Must be a valid `RunStatus` enum value.
+   * @returns {{ runStatus: RunStatus }} A simple object containing only the provided `runStatus`.
+   */
+  wrapRunStatus(runStatus) {
```
graduta: I think this should be in a DTO/util rather than here relying on the service.

graduta (changes requested):

QualityControl/lib/services/BookkeepingService.js
hehoon: 👍 Changes can be viewed in: https://github.com/AliceO2Group/WebUi/pull/3229/commits/e9be197d99766cfb8d47f83d1a6bba5eeaa47f35

QualityControl/test/lib/services/FilterService.test.js
```diff
@@ -17,6 +17,7 @@
 import { FilterService } from '../../../lib/services/FilterService.js';
 import { RunStatus } from '../../../common/library/runStatus.enum.js';
 import { stub, restore } from 'sinon';
+import { BookkeepingService } from '../../../lib/services/BookkeepingService.js';
```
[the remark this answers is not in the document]
hehoon: As of https://github.com/AliceO2Group/WebUi/pull/3229/commits/e9be197d99766cfb8d47f83d1a6bba5eeaa47f35 `BookkeepingService` is no longer used. The import has been removed in: https://github.com/AliceO2Group/WebUi/pull/3229/commits/c160e01d07a12dc2813bbd730b635599e331b669

QualityControl/test/lib/services/BookkeepingService.test.js
graduta, line 1: Given that Bookkeeping service is now returning more information, your tests should also be testing that.
Try to update the mock responses with fields that are not to be used so that you test here that your method only keeps fields that are intended

QualityControl/lib/services/BookkeepingService.js
```diff
@@ -127,17 +128,28 @@ export class BookkeepingService {
...
 
   /**
-   * Retrieves the status of a specific run from the Bookkeeping service
+   * Retrieves the information of a specific run from the Bookkeeping service
    * @param {number} runNumber - The run number to check the status for
-   * @returns {Promise<RunStatus>} - Returns a promise that resolves to the run status:
-   *                                 - RunStatus.ONGOING if the run is ongoing
-   *                                 - RunStatus.ENDED if the run has completed (has timeO2End)
-   *                                 - RunStatus.NOT_FOUND if there was an error or data is not available
+   * @returns {Promise<object>} - Returns a promise that resolves to the run information:
```
graduta: This return doc would go well in a typedef file

QualityControl/lib/services/BookkeepingService.js
```diff
@@ -150,15 +162,36 @@ export class BookkeepingService {
...
 
-      return data.timeO2End ? RunStatus.ENDED : RunStatus.ONGOING;
+      const {
+        startTime,
+        endTime,
+        environmentId,
+        definition,
+        runQuality,
+        lhcBeamMode,
+        detectorQualities,
```
graduta: As `detectorQualities` will probably be used with a `map` function to map to components, I think it would be good to add a default of `[]` in case bookkeeping fails to return you an expected value

graduta (changes requested):
Nice progress! I added some comments for you to review

graduta (approved):

PR #3234, opened 2025-12-13
[OGUI-1661] Object tree breaking when collapsing parents without first collapsing children

hehoon:
- [x] PR labels are selected

To fix the issue the following has been done:
* Replace `map` with `flatMap` in recursive treeRow traversal

In addition to this:
* Prevent redundant `notify()` calls during deep tree operations
* Perform minor code cleanup for clarity and maintainability

### DIFF
--- a/QualityControl/public/object/ObjectTree.class.js
+++ b/QualityControl/public/object/ObjectTree.class.js
@@ -65,22 +65,34 @@ export default class ObjectTree extends Observable {
 
   /**
    * Open all nodes of the tree
-   * @returns {undefined}
    */
   openAll() {
-    this.open = true;
-    this.children.forEach((child) => child.openAll());
+    this._openAllRecursive();
     this.notify();
   }
 
   /**
+   * Recursively open all nodes without notifying.
+   */
+  _openAllRecursive() {
+    this.open = true;
+    this.children.forEach((child) => child._openAllRecursive());
+  }
+
+  /**
    * Close all nodes of the tree
-   * @returns {undefined}
    */
   closeAll() {
-    this.open = false;
-    this.children.forEach((child) => child.closeAll());
+    this._closeAllRecursive();
     this.notify();
+  }
+
+  /**
+   * Recursively close all nodes without notifying.
+   */
+  _closeAllRecursive() {
+    this.open = false;
+    this.children.forEach((child) => child._closeAllRecursive());
   }
 
   /**
--- a/QualityControl/public/object/ObjectTree.class.js
+++ b/QualityControl/public/object/ObjectTree.class.js
@@ -56,30 +56,6 @@ export default class ObjectTree extends Observable {
   }
 
   /**
-   * Open all or close all nodes of the tree
-   * @returns {undefined}
-   */
-  toggleAll() {
-    this.open ? this.closeAll() : this.openAll();
-  }
-
-  /**
-   * Open all nodes of the tree
-   */
-  openAll() {
-    this._openAllRecursive();
-    this.notify();
-  }
-
-  /**
-   * Recursively open all nodes without notifying.
-   */
-  _openAllRecursive() {
-    this.open = true;
-    this.children.forEach((child) => child._openAllRecursive());
-  }
-
-  /**
    * Close all nodes of the tree
    */
   closeAll() {
--- a/QualityControl/public/object/objectTreePage.js
+++ b/QualityControl/public/object/objectTreePage.js
@@ -253,9 +253,7 @@ const treeRowElement = (key, title, onclick, icon, className = '', style = {}) =
     onclick,
     class: className,
   }, [
-    h('td.highlight.flex-row.items-center.g1', {
-      style,
-    }, [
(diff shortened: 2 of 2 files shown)


QualityControl/public/object/ObjectTree.class.js
graduta, line 1: Currently the `openAll` is used by `toggleAll` method, but in reality `toggleAll` is never called as the number of paths is very large and it is not desired to have all of them opened (neither efficient). We only provide the fucntionality for closing all
If a user wants a certain path objects opened, the search functionality makes use of a virtual node to load them.

Thus, I believe we should remove the toggleAll and openALl and _openAllRecursive methods

QualityControl/public/object/objectTreePage.js
```diff
@@ -192,66 +192,93 @@ const treeRows = (model) => !model.object.tree ?
...
  */
-const branchRow = (path, tree, padding) =>
-  h('tr.object-selectable', { key: path, title: path, onclick: () => tree.toggle() }, [
-    h('td.highlight', [
-      h('span', { style: { paddingLeft: padding } }, tree.open ? iconCaretBottom() : iconCaretRight()),
-      ' ',
-      tree.name,
+const branchRow = (key, isTreeOpen, branchName, onClick, styling = {}) =>
+  h('tr.object-selectable', {
+    key: key,
```
graduta: No need to specify the variable value if it has the same name as the key of the object.

So in your case, you can simply put:
```js
{
  key,
  id: key,
  title: branchName,
  onclick  // (of course you should adapt the name in method declaration as it is case sensitive)
}

File: QualityControl/public/object/objectTreePage.js [hunk 2]
```diff
@@ -192,66 +192,93 @@ const treeRows = (model) => !model.object.tree ?
...
-    key: path, title: path, onclick: selectItem, class: className, id: path,
+    key: key,
+    id: key,
+    title: leafName,
+    onclick: onClick,
+    class: className,
   }, [
-    h('td.highlight', [
-      h('span', { style: { paddingLeft: padding } }, iconBarChart()),
-      ' ',
```
graduta, line 237: If the spacing as element is removed, make sure to add some padding to compnensate (see screenshots for comparison of now and before)

Image

Image

graduta (changes requested):
Very good bug fix!
Looking at branchRow and leafRow, I think these have a lot of duplicated code and we should make one reusable component. What do you think?

QualityControl/public/object/objectTreePage.js
```diff
@@ -190,68 +190,73 @@ const treeRows = (model) => !model.object.tree ?
...
+  if (children.length > 0) {
+    // Add a branch row (expandable / collapsible element)
+    const branch = treeRowElement(
+      pathString,
+      name,
+      () => tree.toggle(),
+      open ? iconCaretBottom : iconCaretRight,
+      '',
+      {
+        paddingLeft: `${level + 0.25}em`,
```
graduta: You do not need the padding at all now after the refactor.
You also removed the `span` element, so the default padding of 0.3 from table element is now applied. 

Remove the paddingLeft from the call of treeRowElement both places and you will see it looks like before which is what we want

QualityControl/public/object/objectTreePage.js
```diff
@@ -190,68 +190,73 @@ const treeRows = (model) => !model.object.tree ?
...
- * @returns {vnode} - virtual node element
- */
-const branchRow = (path, tree, padding) =>
-  h('tr.object-selectable', { key: path, title: path, onclick: () => tree.toggle() }, [
-    h('td.highlight', [
-      h('span', { style: { paddingLeft: padding } }, tree.open ? iconCaretBottom() : iconCaretRight()),
-      ' ',
-      tree.name,
+    h('td.highlight.flex-row.items-center.g1', {
+      style: styling,
```
graduta: Given that you are not applying any chances, why not naming the variable `style` and use the short version `{ style }`?

QualityControl/public/object/objectTreePage.js
```diff
@@ -190,68 +190,73 @@ const treeRows = (model) => !model.object.tree ?
...
+ * @param {object} styling - Optional CSS styling for the inner branch row element (table data)
  * @returns {vnode} - virtual node element
  */
-const leafRow = (path, selectItem, className, padding, leafName) =>
+const treeRowElement = (key, name, onclick, icon, className = '', styling = {}) =>
   h('tr.object-selectable', {
-    key: path, title: path, onclick: selectItem, class: className, id: path,
+    key,
+    id: key,
+    title: name,
```
graduta: Same here `title` vs `name`

graduta (changes requested):

QualityControl/public/object/objectTreePage.js
hehoon: As discussed, `paddingLeft` controls the tree indentation and is therefore required.

graduta (approved):

PR #3235, opened 2025-12-13
[OGUI-689] Tree refresh should keep state of opened objects

hehoon:
- [x] PR labels are selected

Changes:
* The tree on the object tree page now persists its expanded node state in local storage.

Note:
* The estimated storage size (in bytes) is approximately: `A * (N + 7) * 2`
  * `A` is the number of expanded (opened) nodes
  * `N` is the average character length of expanded (opened) node names
* The storing of tree node states will break if two identical names are stored directly under the same parent. This limitation already existed in the previous implementation and is unchanged by the current persistence mechanism.
* A temporary workaround was added to handle local storage resets in tests, since some tests are not fully isolated and depend on shared state. Ideally, all tests should be isolated, with local storage and page state fully reset between each test.

### DIFF
--- a/QualityControl/public/common/enums/storageKeys.enum.js
+++ b/QualityControl/public/common/enums/storageKeys.enum.js
@@ -20,4 +20,5 @@
 export const StorageKeysEnum = Object.freeze({
   OBJECT_VIEW_LEFT_PANEL_WIDTH: 'object-view-left-panel-width',
   OBJECT_VIEW_INFO_VISIBILITY_SETTING: 'object-view-info-visibility-setting',
+  OBJECT_TREE_OPEN_NODES: 'object-tree-open-nodes',
 });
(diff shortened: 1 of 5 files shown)


graduta (changes requested):
The features works very well, good job!

I believe there is one case that you missed though. If the user clicks on the `Collapse all` button from the header of the page, the state of the objects is not saved, so a refresh of the page will reopen the previous state

hehoon:
> The features works very well, good job!
> 
> I believe there is one case that you missed though. If the user clicks on the `Collapse all` button from the header of the page, the state of the objects is not saved, so a refresh of the page will reopen the previous state

This should be solved as of https://github.com/AliceO2Group/WebUi/pull/3235/commits/5b80e3ca570616da7e80795feac0f3951e8512c2

EDIT: Also see: https://github.com/AliceO2Group/WebUi/pull/3235/commits/221980ef794106b863da0b0718e9db9500b2d319

graduta (dismissed):

graduta (approved):

PR #3236, opened 2025-12-15
[OGUI-1847] Fix arrays saved as objects and other small fixes

Deaponn:
This PR fixes a couple of bugs:
 - one where arrays are converted into objects on configuration save (`[true, 'text', 99]` becomes `{ 0: true, 1: 'text', 2: 99 }`)
 - one where edit difference highlighting fails for FormToggleInput component
 - one where empty objects and lists are not parsed correctly which causes them to be lost while saving 

- [x] PR labels are selected


piechnikk (approved):
lgtm

sKrzysieK (approved):

PR #3246, opened 2025-12-18
[OGUI-1851] ObjectInfo Panel on hover should be left/right depending on position

AlexJanson:
- [x] PR labels are selected

When viewing a layout, the user has the option to hover over a plot and get multiple options. Here, they can click on the `info` icon to get a panel with information about the object.

The problem lies of the fact that if the object is positioned on the left side of the screen, it will not be visible properly.

Now when we are on the left side of the window it will render the dropdown to the right of the button anchor.

### DIFF
--- a/QualityControl/public/layout/view/panels/objectInfoResizePanel.js
+++ b/QualityControl/public/layout/view/panels/objectInfoResizePanel.js
@@ -13,6 +13,7 @@
  */
 
 import { downloadButton } from '../../../common/downloadButton.js';
+import { isOnLeftSideOfViewport } from '../../../common/utils.js';
 import { defaultRowAttributes, qcObjectInfoPanel } from './../../../common/object/objectInfoCard.js';
 import { h, iconResizeBoth, info } from '/js/src/index.js';
 
@@ -24,9 +25,9 @@ import { h, iconResizeBoth, info } from '/js/src/index.js';
  * @returns {vnode} - virtual node element
  */
 export const objectInfoResizePanel = (model, tabObject) => {
-  const { name, id } = tabObject;
+  const { name } = tabObject;
   const { filterModel, router, object, services } = model;
-  const isSelectedOpen = object.selectedOpenName === id;
+  const isSelectedOpen = object.selectedOpenName === name;
   const objectRemoteData = services.object.objectsLoadedMap[name];
   let uri = `?page=objectView&objectId=${tabObject.id}&layoutId=${router.params.layoutId}`;
   Object.entries(filterModel.filterMap)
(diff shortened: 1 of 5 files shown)


QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -24,9 +24,9 @@ import { h, iconResizeBoth, info } from '/js/src/index.js';
  * @returns {vnode} - virtual node element
  */
 export const objectInfoResizePanel = (model, tabObject) => {
-  const { name } = tabObject;
+  const { name, id } = tabObject;
   const { filterModel, router, object, services } = model;
-  const isSelectedOpen = object.selectedOpen;
+  const isSelectedOpen = object.selectedOpenName === id;
```
graduta: Doing this, it means that the selection will stay open only for one object which is not what we desired. 

The functionality was requested and implemented in such a way that the users can quickly hover over the objects in sequence with the pop-up staying open so that they can compare the metadata.

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
```diff
@@ -35,18 +35,29 @@ export const objectInfoResizePanel = (model, tabObject) => {
...
+          onupdate: (vnode) => {
+            if (object.isOnLeftSide(vnode.dom.parentElement)) {
+              vnode.dom.style.left = '0.1em';
+              vnode.dom.style.right = 'auto';
+            } else {
+              vnode.dom.style.right = '0.1em';
+              vnode.dom.style.left = 'auto';
+            }
+          },
+        },
```
graduta, line 61: Would it not be better to calculate the style position above and use it here directly rather than setting a position on L50, just to modify after?

QualityControl/public/object/QCObject.js
```diff
@@ -653,4 +650,21 @@ export default class QCObject extends BaseViewModel {
...
   }
+
+  /**
+   * Determines whether the element is positioned on the left half of the viewport.
+   * This is used to decide which way a dropdown should anchor to stay within view.
+   * @param {HTMLElement} element - The DOM element (usually the button or container) to measure.
+   * @returns {boolean|undefined} Returns true if the element is on the left half of the window,
+   * false if it is on the right half, or undefined if no element is provided.
+   */
+  isOnLeftSide(element) {
```
graduta: I think this would be better in a util folder so that other components can use it if needed

graduta (changes requested):

graduta (changes requested):
I had another look at the PR as you pushed some changes.
The previous behaviour of allowing the users to quickly hover objects and pop-up to stay there. This is because of the changes you made in QCObject. I would recommend that those you do not change as the feature was there on purpose.

QualityControl/public/layout/view/panels/objectInfoResizePanel.js
graduta: As you correctly identified, this would not work as the above position is outside of the parent scope

graduta (approved):

PR #3247, opened 2025-12-18
[OGUI-1853] Move actionables of objectTree in the table header

AlexJanson:
- [x] PR labels are selected

As a user, it is difficult to see that object tree can be filtered by name. Thus, the search input text (from top right in header) should be moved in the table header in a nice UI and UX manner.

Moreover, the (collapse all tree button) should be moved also in the table header(right most).
Moreover, the sort by name buton should also become part of the header so that when the user clicks on the header Name column, to display a sort icon up, or down or nothing depending on what is currently being sorted.
See for example how Bookkeeping sorts by Title on page "Log Entries" (log-overview)

### DIFF
--- a/QualityControl/public/common/enums/columnSort.enum.js
+++ b/QualityControl/public/common/enums/columnSort.enum.js
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+ * See http://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * Enumeration for sort directions
+ * @enum {number}
+ * @readonly
+ */
+export const SortDirectionsEnum = Object.freeze({
+  NONE: 0,
+  ASC: 1,
+  DESC: -1,
+});
--- a/QualityControl/public/common/enums/columnSort.enum.js
+++ b/QualityControl/public/common/enums/columnSort.enum.js
@@ -18,7 +18,6 @@
  * @readonly
  */
 export const SortDirectionsEnum = Object.freeze({
-  NONE: 0,
   ASC: 1,
   DESC: -1,
 });
--- a/QualityControl/public/common/sortButton.js
+++ b/QualityControl/public/common/sortButton.js
@@ -1,0 +1,72 @@
+/**
+ * @license
+ * Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+ * See http://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { SortDirectionsEnum } from '../common/enums/columnSort.enum.js';
+import { h, iconCircleX, iconArrowBottom, iconArrowTop } from '/js/src/index.js';
+
+/**
+ * Get the icon for the sort direction.
+ * @param {SortDirectionsEnum} direction - direction of the sort.
+ * @returns {vnode} the correct icon related to the direction.
+ */
+const getSortIcon = (direction) => {
+  if (direction === SortDirectionsEnum.ASC) {
+    return iconArrowTop();
+  }
+  if (direction === SortDirectionsEnum.DESC) {
+    return iconArrowBottom();
+  }
+  return iconCircleX();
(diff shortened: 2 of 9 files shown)


graduta (changes requested):
Right now, the actionable buttons/search have not been moved to the table header, but instead at the top of the table.
Could you have a look at some ag-grid tables which offer the search input integrating in the header to provide a better UX? Somewhere under the name but in the headear of the table
While the collapse all can also be in the header most right

QualityControl/public/common/enums/columnSort.enum.js
```diff
@@ -0,0 +1,24 @@
...
+ * or submit itself to any jurisdiction.
+ */
+
+/**
+ * Enumeration for sort directions
+ * @enum {number}
+ * @readonly
+ */
+export const SortDirectionsEnum = Object.freeze({
+  NONE: 0,
```
graduta: Is there a way to actually not have them sorted? I did not manage to get the UI in such a state.
Perhaps we should remove the option if that is the case?

QualityControl/public/common/sortButton.js
```diff
@@ -0,0 +1,81 @@
...
+  const nextSortOrder = sortOptions[nextIndex];
+  const hoverIcon = getSortIcon(nextSortOrder);
+
+  const directionLabel = Object.keys(SortDirectionsEnum).find((key) => SortDirectionsEnum[key] === nextSortOrder);
+
+  return h(
+    'button.btn.sort-button',
+    {
+      onclick: () => onclick(label, nextSortOrder, hoverIcon),
+      title: `Sort by ${directionLabel}`,
```
graduta: Perhaps you meant: `Sort ${directionLabel} by Name`?

QualityControl/public/object/objectTreePage.js
```diff
@@ -50,9 +60,15 @@ export default (model) => {
             const objectsLoaded = object.list;
             const objectsToDisplay = objectsLoaded.filter((qcObject) =>
               qcObject.name.toLowerCase().includes(searchInput.toLowerCase()));
-            return virtualTable(model, 'main', objectsToDisplay);
+            return h('.scroll-y.flex-column.flex-grow', [
```
graduta: Is there a need for scroll-y on this? The table header should remain on top rather than being part of the scroll

QualityControl/public/object/objectTreePage.js
```diff
@@ -170,11 +186,45 @@ const statusBarRight = (model) => model.object.selected
  * @returns {vnode} - virtual node element
  */
 const tableShow = (model) =>
-  h('table.table.table-sm.text-no-select', [
-    h('thead', [h('tr', [h('th', 'Name')])]),
-    h('tbody', [treeRows(model)]),
+  h('table.table.table-sm.text-no-select', h('tbody', [treeRows(model)]));
+
+const tableHeaderRow = (model) => h('.bg-gray-light.pv2', [
```
graduta: This method does not need the entire root model.
I think it is better to send only the `object` one. 

Moreover the method is missing documentation.

QualityControl/public/object/objectTreePage.js
```diff
@@ -170,11 +186,45 @@ const statusBarRight = (model) => model.object.selected
...
+  sortableTableHead({
+    order: model.object.sortBy.order,
```
graduta: I think it would be nice to destruct the sortBy JSON before creating the node (same method tableHeaderRow but with a return statement returning the node rather than node being returned by arrow function) and assign default values to order and icon

QualityControl/public/object/objectTreePage.js
```diff
@@ -170,11 +186,45 @@ const statusBarRight = (model) => model.object.selected
...
+    icon: model.object.sortBy.icon,
+    label: 'Name',
+    sortOptions: [SortDirectionsEnum.ASC, SortDirectionsEnum.DESC],
```
graduta: Did you mean to add "None" here as well?

QualityControl/public/object/objectTreePage.js
graduta: I think the naming of these two methods is not clear enough.
`tableHeaderRow` is actually not a single row, right? But instead contains the Name of column header and another row with search input
On the other hand `tableHeader` is kind of a row but not a table and not a header but instead a `div`.

Try to come up with better descriptive namings

QualityControl/public/object/objectTreePage.js
graduta, line 1: All the newly added methods in this file are missing documentation. Please make sure to add

QualityControl/public/common/sortButton.js
```diff
@@ -0,0 +1,81 @@
...
+  sortOptions = [...Object.values(SortDirectionsEnum)],
+}) => {
+  const currentIndex = sortOptions.indexOf(order);
+  const nextIndex = (currentIndex + 1) % sortOptions.length;
+  const nextSortOrder = sortOptions[nextIndex];
+  const hoverIcon = getSortIcon(nextSortOrder);
+
+  const directionLabel = Object.keys(SortDirectionsEnum).find((key) => SortDirectionsEnum[key] === nextSortOrder);
+
+  return h(
```
graduta, line 67: I think it would be good to keep the same UX/UI for the users as for other applications that we develop.
Please have a look at Bookkeeping on what design they use and try to implement a similar one: attaching here a screenshot but perhaps running BKP locally would be more helpful

Image

graduta (changes requested):
Implementation wise, it works very well and the virtual table is once again not blocking the page. 
Good job!

I left a few comments with regards to design and documentation

graduta (approved):

PR #3256, opened 2026-01-05
[OGUI-561] Browse object tree with key arrows

generatedkeys:
- [x] PR labels are selected

# 
- Contains bug fix for [OGUI-1864]


graduta (changes requested):
Very nice starting point! Navigation with keyboards works as expected. 
There are a few corner cases that we should deal with:
- when navigating by arrows in a scrollable list of object trees, the scroll should move together with selected node if it goes outside of the page. 
- if a user selects with the mouse to either open/close a node or open a plot, the keyboard navigator should move to that selected node

PR #3257, opened 2026-01-05
[OGUI-1854] Add native browser notification helper functions

hehoon:
- [x] PR labels are selected

Changes:
- Add helpers to get, request, and check browser notification permissions

### DIFF
--- a/Framework/Frontend/js/src/BrowserNotification.js
+++ b/Framework/Frontend/js/src/BrowserNotification.js
@@ -1,0 +1,93 @@
+/**
+ * @license
+ * Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+ * See http://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+/* Global: window */
+
+import { isContextSecure } from './utilities/browserContext.js';
+
+/**
+ * Get the current browser notification permission.
+ * @returns {NotificationPermission|undefined} {@link NotificationPermission}, or `undefined` if unsupported.
+ */
+export const getBrowserNotificationPermission = () =>
+  window?.Notification?.permission;
+
+/**
+ * Request browser notification permission (async/await).
+ * @returns {Promise<NotificationPermission|undefined>} Resolves to {@link NotificationPermission}, or `undefined` if unsupported
+ */
+export const requestBrowserNotificationPermissions = async () => {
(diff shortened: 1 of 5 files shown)


Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -65,6 +65,8 @@ export class CopyToClipboardComponent extends StatefulComponent {
   /**
    * Checks if context is secure (HTTPS)
    *
+   * @deprecated Use `isContextSecure` from `./utilities/browserContext.js` instead.
+   * Calling this method directly is discouraged; it may be removed in future versions.
    * @returns {boolean} Returns `true` if context is secure
    */
   isContextSecure() {
```
graduta: As this method is private and used within the `CopyToClipboardComponent.js` only, could you please rather than deprecate it, update it to use the one you recommend?

Framework/Frontend/js/src/utilities/browserNotification.js
graduta, line 1: As this is not a class, the naming should be `browserNotification` maybe even adding something to point out is a utility package.

Given we also have BrowserStorage, maybe you want to create a directory for browser only and keep all related in there

Framework/Frontend/js/src/BrowserNotification.js
```diff
@@ -0,0 +1,93 @@
...
+ * Request browser notification permission (async/await).
+ * @returns {Promise<NotificationPermission|undefined>} Resolves to {@link NotificationPermission}, or `undefined` if unsupported
+ */
+export const requestBrowserNotificationPermissions = async () => {
+  if (!isContextSecure()) {
+    return undefined;
+  }
+
+  const permission = getBrowserNotificationPermission();
+  if (permission === 'granted') {
```
graduta: Given that the permission value can only be: `granted`, `default`, `denied`, I think it would be good to have an object with these values in the file and reuse it from there.
In this way you can also use it as reference in docs with that `getBrowserNotificationPermission` can return

graduta (changes requested):

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
hehoon: 👍 Changed in https://github.com/AliceO2Group/WebUi/pull/3257/commits/b13cd140768cf77e255b3e52262811742f6efa5a

Framework/Frontend/js/src/utilities/browserNotification.js
hehoon: 👍 Changed in https://github.com/AliceO2Group/WebUi/pull/3257/commits/8b3447ecc54a77393cabd7c5ed79fa075a30b422

Framework/Frontend/js/src/BrowserNotification.js
hehoon: 👍 Changed in https://github.com/AliceO2Group/WebUi/pull/3257/commits/e67dbdcded24de2f618ab51dfab9d87061593f51

graduta (approved):

PR #3258, opened 2026-01-05
[OGUI-1854] Notify user via native browser notification when a new run has started

hehoon:
- [x] PR labels are selected

Depends on:
- https://github.com/AliceO2Group/WebUi/pull/3257

Changes:
- Added a websocket to broadcast new runs from backend towards front-end
- Added a menu item in the profile (click on top left profile button to open menu) to enable browser notifications.
  - Front-end browser checks whether the user enabled this settings before displaying a notification


QualityControl/lib/services/RunModeService.js
```diff
@@ -134,6 +137,15 @@ export class RunModeService {
     } else if (transition === Transition.STOP_ACTIVITY) {
       this._ongoingRuns.delete(runNumber);
     }
+
+    const wsMessage = new WebSocketMessage();
```
graduta: Each event sent by Kafka on the run track has multiple transitions. For the notification of the user, we should only send one broadcast message when we receive the event of START_ACTIVITY: [L127](https://github.com/AliceO2Group/WebUi/pull/3258/changes#diff-3d9c44a9016e3ad0fda0a2a947b95a16b8398aee53a5a3565c2fa9cffc4f0000R127)
Otherwise, we risk sending up to 10 messages for a run start at various intervals of seconds
This information should have been provided in the ticket but it was not, I apologise for the confusion

QualityControl/lib/services/RunModeService.js
```diff
@@ -29,16 +29,19 @@ export class RunModeService {
...
    * @param {CcdbService} dataService - Used to fetch data from the CCDB.
    * @param {EventEmitter} eventEmitter - Event emitter to be used to emit events when new data is available
+   * @param {WebSocket} ws - web-ui websocket server implementation
    */
   constructor(
     config,
     bookkeepingService,
     dataService,
     eventEmitter,
+    ws,
```
graduta: As this is passed as a parameter, could you please use a more descriptive naming? Perhaps `webSocketService` or something along those lines

QualityControl/public/common/notifications/model/NotificationRunStartModel.js
```diff
@@ -0,0 +1,100 @@
...
+   * A native browser notification is displayed only when:
+   * - The transition is {@link Transition.START_ACTIVITY}
+   * - The user has enabled notifications
+   * @param {object} payload - WebSocket payload.
+   * @param {number} payload.runNumber - Run number that started.
+   * @param {Transition} payload.transition - Transition type.
+   * @returns {undefined}
+   */
+  async _handleWSRunTrack({ runNumber, transition }) {
+    if (transition !== Transition.START_ACTIVITY) {
```
graduta: With the backend change to only notify on start activity, this check can be removed

QualityControl/public/common/header.js
```diff
@@ -110,5 +119,37 @@ const loginButton = (model) =>
...
+        [
+          h(
+            '.switch',
+            [
+              h('input', {
+                onchange: async (event) => {
+                  let permissionGranted = false;
+                  if (event.target.checked) {
+                    const permission = await requestBrowserNotificationPermissions();
+                    permissionGranted = permission === BrowserNotificationPermission.GRANTED;
```
graduta: What about using the framework's method `areBrowserNotificationsGranted` instead?

QualityControl/public/common/header.js
```diff
@@ -110,5 +119,37 @@ const loginButton = (model) =>
       model.session.personid === 0 // Anonymous user has id 0
         ? h('p.m3.gray-darker', 'This instance of the application does not require authentication.')
         : h('a.menu-item', { onclick: () => alert('Not implemented') }, 'Logout'),
+      h(
```
graduta: It would be great to extract this into its own documented component which should recieve only the needed model `notificationRunStartModel`

QualityControl/public/common/notifications/model/NotificationRunStartModel.js
```diff
@@ -0,0 +1,100 @@
...
+      onclick: () => {
+        const { isRunModeActivated } = this.model.filterModel;
+        if (!isRunModeActivated) {
+          const viewModel = this.model.filterModel.getPageTargetModel();
+          if (viewModel) {
+            this.model.filterModel.activateRunsMode(viewModel);
+          }
+        }
+
+        this.model.filterModel.setFilterValue('RunNumber', runNumber?.toString(), true);
```
graduta: I believe this method's logic can be improved and I should have provided more details in the requirements. 

Basically, when such a notification is provided the user can either be:
1. In RunMode of a previous RunNumber (for sure it cannot be in the current one)
- in this case, the method should update the run mode with new runNumber
2. Outside of runMode
  a. If Kafka is configured, then the method should activate run mode with new run number
  b. If Kafka is not configured, the method simply update the Filter RunNumber parameter with new run number and trigger the filtering

In all scenarios, the user should be taken back to `objectTree` page no matter where they were

QualityControl/public/common/notifications/model/NotificationRunStartModel.js
```diff
@@ -0,0 +1,100 @@
...
+      }
+    });
+  }
+
+  /**
+   * Returns whether browser notifications for run start events
+   * are enabled for the current user.
+   * @returns {boolean} `true` if notifications are enabled, `false` otherwise.
+   */
+  getBrowserNotificationSetting() {
```
graduta, line 47: Should this method not take into account the browser's permissions?
I am thinking of the following use case:
- user enables notifications via QCG which in turns saves it as true in the local storage
- at some point they disable the notifications via the browser instead. At this point, QCG still thinks notifications are enabled and will be sent but instead they will not be.

graduta (changes requested):

QualityControl/public/common/notifications/model/NotificationRunStartModel.js
hehoon: Good catch, this has been adjusted 👍

QualityControl/public/common/notifications/model/NotificationRunStartModel.js
hehoon: This has been adjusted to support scenario `1` and `2a` 👍 .

To receive notifications from the websocket kafka must be enabled, so I believe scenario `2b` cannot occur. Do let me know if I missed something.

graduta (approved):

PR #3438, opened 2026-05-07
[OGUI-1892] Disable level buttons for shifter users on page load and after

graduta:
- [x] PR labels are selected

PR which:
* if user authenticated has a role of  "shifter" but not "admin":
    *  the UI will disable the buttons changing the "level" of messages filtering system
    * even if the URL contains a higher level, this level will be brought back to level 1 (operations) 
* adds tests to validated newly added changes
* adds const/enums for roles and infologger levels
* extracts button into a reusable component and defines objects for existing filters

### DIFF
--- a/InfoLogger/public/Model.js
+++ b/InfoLogger/public/Model.js
@@ -19,6 +19,7 @@ import {
 } from '/js/src/index.js';
 import { callRateLimiter, setBrowserTabTitle, hasShifterButNoAdminRole } from './common/utils.js';
 import { ConfigurationService } from './services/ConfigurationService.js';
+import { InfoLoggerLevel } from './constants/infologger-level.const.js';
 import { MODE } from './constants/mode.const.js';
 import Log from './log/Log.js';
 import Table from './table/Table.js';
@@ -45,6 +46,9 @@ export default class Model extends Observable {
 
     this.log = new Log(this);
     this.log.bubbleTo(this);
+    if (hasShifterButNoAdminRole(this.session.access)) {
+      this.log.filter.setConstraint('level', 'max', InfoLoggerLevel.OPS.index);
+    }
 
     this.table = new Table(this);
     this.table.bubbleTo(this);
@@ -311,15 +315,11 @@ export default class Model extends Observable {
 
   /**
    * Delegates sub-model actions depending new location of the page
-   * If user is shifter but not admin, set Ops as maximum level for filtering
    */
   handleLocationChange() {
     const { params } = this.router;
     if (params) {
       this.parseLocation(params);
-    }
-    if (hasShifterButNoAdminRole(this.session.access)) {
-      this.log.filter.setCriteria('level', 'max', 1);
     }
   }
 
--- a/InfoLogger/public/Model.js
+++ b/InfoLogger/public/Model.js
@@ -17,7 +17,7 @@ import {
   Observable, WebSocketClient, QueryRouter,
   Loader, RemoteData, sessionService, Notification,
 } from '/js/src/index.js';
-import { callRateLimiter, setBrowserTabTitle } from './common/utils.js';
+import { callRateLimiter, setBrowserTabTitle, hasShifterButNoAdminRole } from './common/utils.js';
 import { ConfigurationService } from './services/ConfigurationService.js';
 import { MODE } from './constants/mode.const.js';
 import Log from './log/Log.js';
@@ -81,23 +81,8 @@ export default class Model extends Observable {
     // Model can change very often we protect router with callRateLimiter
     // Router limit: 100 calls per 30 seconds max = 30ms, 2 FPS is enough (500ms)
     this.observe(callRateLimiter(this.updateRouteOnModelChange.bind(this), 500));
-
-    this._filterLevelsAllowed = [
-      { label: 'Ops', index: 1, available: true },
-      { label: 'Support', index: 6, available: false },
-      { label: 'Devel', index: 11, available: false },
-      { label: 'Trace', index: null, available: false },
-    ];
   }
 
-  /**
-   * Method to return filter levels allowed for filtering based on current user role email groups afilliation
-   * * Shifters are only allowed to filter by Ops level
-   * @returns {{label: string, index:number}[]} - filter levels allowed for filtering
-   */
-  get filterLevelsAllowed() {
-    return this._filterLevelsAllowed;
-  }
 
   /**
    * Handle websocket authentication success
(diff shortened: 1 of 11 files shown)


InfoLogger/public/logFilter/commandFilters.js
```diff
@@ -14,92 +14,97 @@
...
-  },
-  title: title,
-}, label);
+  h(
+    '.btn-group',
+    filterLevelsAllowed.map(({ label, index, available }) => _selectableButtonComponent(
+      label,
+      {
+        id: `level-${index}`,
+        title: available ? `Filter level ≤ ${index}` : `You don't have access to level ${label}`,
```
isaachilly: I think the tooltip will show `Filter level ≤ null` for Trace as it doesn't have an index.

InfoLogger/public/logFilter/commandFilters.js
```diff
@@ -14,92 +14,97 @@
...
- * @param {string} label - button's label
- * @param {string} title - button's title on mouse over
- * @param {string} value - a char to represent severity: W E F or I, can be many with spaces like 'W E'
- * @returns {vnode} - the button to toggle severity
- */
-const buttonSeverity = (model, label, title, value) => h('button.btn', {
-  className: model.log.filter.criterias.severity.in.includes(value) ? 'active' : '',
-  onclick: (e) => {
-    model.log.setCriteria('severity', 'in', value);
-    e.target.blur(); // remove focus so user can 'enter' without actually toggle again the button
```
isaachilly, line 66: Was `e.target.blur()` intentionally removed?

InfoLogger/public/logFilter/commandFilters.js
```diff
@@ -14,92 +14,97 @@
...
+        },
+      )),
+  ),
+
+  _selectableButtonComponent(
+    'Reset filters',
+    {
+      title: 'Reset date, time, matches, excludes, log levels',
+      isActive: false,
+      onclick: () => logModel.filter.resetCriteria(),
```
isaachilly, line 87: Reset filters will allow resetting to a default, more permissive criteria set.

InfoLogger/public/Model.js
```diff
@@ -311,12 +311,16 @@ export default class Model extends Observable {
...
    * Delegates sub-model actions depending new location of the page
+   * If user is shifter but not admin, set Ops as maximum level for filtering
    */
   handleLocationChange() {
     const { params } = this.router;
     if (params) {
       this.parseLocation(params);
     }
+    if (hasShifterButNoAdminRole(this.session.access)) {
+      this.log.filter.setCriteria('level', 'max', 1);
```
isaachilly: Could use infologger-level.const.js?

InfoLogger/test/public/shifter-based-actions-mocha.js
```diff
@@ -0,0 +1,80 @@
...
+      const location = await page.evaluate(() => window.location);
+      const search = decodeURIComponent(location.search);
+      assert.strictEqual(search, '?q={"severity":{"in":"I W E F"},"level":{"max":1}}');
+
+      const criterias = await page.evaluate(() => window.model.log.filter.criterias);
+      assert.strictEqual(criterias.level.max, 1);
+    },
+  );
+
+  it('should disable buttons for level filter if user is shifter but not admin', async () => {
```
isaachilly, line 61: A couple of edge cases should there be a test for checking what happens if user has both admin and shifter roles?

What should happen if user doesn't have shifter or admin roles?

InfoLogger/public/common/utils.js
```diff
@@ -60,3 +63,30 @@ export function setBrowserTabTitle(title = undefined) {
...
+  return access.includes(Role.SHIFTER) && !access.includes(Role.ADMIN);
+}
+
+/**
+ * Method to return filter levels allowed for filtering based on current user role email groups affiliation
+ * * Shifters are only allowed to filter by Ops level
+ * @param {string[]} access - array of user roles email groups affiliation
+ * @returns {{label: string, index:number}[]} - filter levels allowed for filtering
+ */
+export function getFilterLevelsAllowed(access = []) {
```
isaachilly, line 82: I comment here, but the change would be elsewhere The access is not checked server-side? If the user changes it locally they can still query things they aren't allowed to. Are we okay with this level of security?

InfoLogger/public/Model.js
```diff
@@ -311,12 +311,16 @@ export default class Model extends Observable {
 
   /**
    * Delegates sub-model actions depending new location of the page
+   * If user is shifter but not admin, set Ops as maximum level for filtering
    */
   handleLocationChange() {
```
isaachilly, line 319: Overwriting on each navigation call could be unnecessary and setCriteria could be a better more central place for it to be checked and set to a default.

isaachilly (commented):
Please see my thoughts and comments.

InfoLogger/public/common/utils.js
graduta: This is not a security problem because our interest is not to hide these logs for shifter levels, but simply to help them better navigate the logs. 

If they modify it themselves locally via scripts or curl commands, it means they have a reason for looking at a larger set of logs

InfoLogger/public/logFilter/commandFilters.js
graduta: Indeed, updated

InfoLogger/public/logFilter/commandFilters.js
graduta: From my point of view this was not needed from the start, because the key handler in Model takes care of it. The button will not impact the "enter" for quering the logs

InfoLogger/public/logFilter/commandFilters.js
graduta: Indeed, good catch! 
I have updated it so that we apply the last available/allowed level

InfoLogger/public/Model.js
graduta: Yes, I should have done that from start!

InfoLogger/public/Model.js
graduta: My reasoning was that I want to keep LogFilter decoupled from the logic of roles.
In order to do so and apply your suggestion, I introduced constraing in the filter which gets taken into account when setting criteria.
This in turn will also help with the `resetCriteria` logic

InfoLogger/test/public/shifter-based-actions-mocha.js
graduta: Added

isaachilly (approved):

graduta:
Closing this PR for now, details in the Jira ticket

PR #3440, opened 2026-05-08
[OGUI-1577] Alert user of no filters set

isaachilly:
- [X] PR labels are selected

* Adds a method to the filter model to check if any text filters are active. 
* Query handler calls this method and if false opens a confirmation dialog to the user.
* Does or doesn't execute query based on user's response.

image

### DIFF
--- a/InfoLogger/public/log/Log.js
+++ b/InfoLogger/public/log/Log.js
@@ -327,11 +327,8 @@ export default class Log extends Observable {
     }
 
     if (!this.filter.hasActiveTextFilters()) {
-      // If the browser does not support the confirm dialog, execute the query anyway
-      const shouldExecuteQueryWithoutFilters = typeof window.confirm === 'function'
-        ? window.confirm('No date or text filters set. This will return a large amount of data. Execute query anyway?')
-        : true;
-      if (!shouldExecuteQueryWithoutFilters) {
+      if (!window.confirm('No date or text filters set.'
+        + ' This will return a large amount of data. Execute query anyway?')) {
         return;
       }
     }
--- a/InfoLogger/public/log/Log.js
+++ b/InfoLogger/public/log/Log.js
@@ -325,6 +325,17 @@ export default class Log extends Observable {
     if (!this.model.frameworkInfo.isSuccess() || !this.model.frameworkInfo.payload.mysql.status.ok) {
       throw new Error('Query service is not available');
     }
+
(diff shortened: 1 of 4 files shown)


InfoLogger/public/log/Log.js
```diff
@@ -325,6 +325,17 @@ export default class Log extends Observable {
     if (!this.model.frameworkInfo.isSuccess() || !this.model.frameworkInfo.payload.mysql.status.ok) {
       throw new Error('Query service is not available');
     }
+
+    if (!this.filter.hasActiveTextFilters()) {
+      // If the browser does not support the confirm dialog, execute the query anyway
+      const shouldExecuteQueryWithoutFilters = typeof window.confirm === 'function'
+        ? window.confirm('No date or text filters set. This will return a large amount of data. Execute query anyway?')
```
graduta: There is no need for this check as browsers that we provide support for, have this feature from V1

InfoLogger/test/public/query-mode-mocha.js
```diff
@@ -17,6 +17,46 @@
...
+    // Mock the frameworkInfo to make the query method think the query service is available in its check
+    window.model.frameworkInfo = {
+      isSuccess: () => true,
+      payload: { mysql: { status: { ok: true } } },
+    };
+
+    // Default state of filters includes no text filters
+    window.model.log.filter.resetCriteria();
+    if (setTextFilter) {
+      window.model.log.filter.setCriteria('message', 'match', 'some-message');
```
graduta: As your function `hasActiveTextFilters` checks `['since', 'until', 'match', 'exclude'];`, I think it would be good for the test to also cover all 4. 
Perhaps extract the list as a const and import in tests and use it?

graduta (changes requested):

InfoLogger/public/log/Log.js
isaachilly: Yes I did look up if really there was any way for it to be an issue and I agree not, but I'd already added it by then and thought it's harmless.

isaachilly: Updated.

InfoLogger/test/public/query-mode-mocha.js
isaachilly: Updated.

graduta (approved):

PR #3463, opened 2026-05-20
[OGUI-1453] Add context-menu with filtering and more general actions to ILG table cells

isaachilly:
- [x] PR labels are selected

* **Right-Click Context Menu:** Right-clicking a cell now brings up a context menu with four actions: two for additive/reductive filtering, and two field-independent actions ("Clear Filter" and "Copy").
* Opted to use a context menu over modifying the double-click functionality for two reasons:
  1. It avoids disrupting the likely learnt user behaviour of double-click to immediately open the inspector.
  2. Right-click is an intuitive, immediately reachable, and previously unused way to interact with ILG.
* Adds tests to validate context menu opening, closing and specific action functionalities.


InfoLogger/public/log/cellContextMenu.js
```diff
@@ -0,0 +1,116 @@
...
+  top: Math.max(0, Math.min(y, window.innerHeight - MENU_HEIGHT_ESTIMATE)),
+});
+
+/**
+ * Context menu for log table cells — allows quick filter actions.
+ * Rendered at view root with position:fixed to avoid virtual scroll issues.
+ * @param {Model} model root application model
+ * @returns {Array|null} rendered menu nodes
+ */
+export default (model) => {
```
graduta: We should export named components and not default

InfoLogger/public/log/cellContextMenu.js
```diff
@@ -0,0 +1,116 @@
...
+        });
+        hideMenu();
+      }),
+    ]),
+  ];
+};
+
+/**
+ * Creates menu item for the context menu of a cell with given icon, label and click action.
+ * @param {string} icon - icon to display in the menu item
```
graduta: I do not think this icon is a string

InfoLogger/public/log/cellContextMenu.js
```diff
@@ -0,0 +1,116 @@
...
+ * @param {string} iconColor - color of the icon
+ * @param {string} label - label to display in the menu item
+ * @param {() => void} onClick - function to execute on click
+ * @returns {vnode} - the menu item as a vnode
+ */
+function createMenuItem(icon, iconColor, label, onClick) {
+  return h('.cell-context-menu-item.f7', {
+    onclick: onClick,
+  }, [
+    h('span', { style: { color: iconColor } }, icon),
```
graduta: What about using classes instead of variables? You can pass `.danger`, `.warning`, etc. These are defined  already in web-ui bootstrap.css

InfoLogger/public/log/tableLogsContent.js
```diff
@@ -71,6 +71,30 @@ const tableLogLine = (model, row) => h('tr.row-hover', {
...
+  }
+  if (field === 'time') {
+    return row.timestamp
+      ? { field: 'timestamp', value: `${model.timezone.format(row.timestamp, 'date')} ${content}` }
+      : null;
+  }
+  return row[field] != null && row[field] !== '' ? { field, value: String(row[field]) } : null;
+};
+
+const cellWithContextMenu = (model, row, field, content, extraClasses = '', extraAttrs = {}) =>
```
graduta: Missing docs

InfoLogger/public/log/tableLogsContent.js
```diff
@@ -71,6 +71,30 @@ const tableLogLine = (model, row) => h('tr.row-hover', {
   ondblclick: () => model.toggleInspector(),
 }, tableRows(model, model.table.colsHeader, row));
 
+const resolveContextMenuData = (model, row, field, content) => {
```
graduta: Missing docs

InfoLogger/public/log/tableLogsContent.js
```diff
@@ -82,21 +106,39 @@ const tableRows = (model, colsHeader, row) =>
...
-    colsHeader.system.visible && h('td.cell.cell-bordered', row.system),
-    colsHeader.facility.visible && h('td.cell.cell-bordered', row.facility),
-    colsHeader.detector.visible && h('td.cell.cell-bordered', row.detector),
-    colsHeader.partition.visible && h('td.cell.cell-bordered', row.partition),
-    colsHeader.run.visible && h('td.cell.cell-bordered', row.run),
-    colsHeader.errcode.visible && h('td.cell.cell-bordered', linkToWikiErrors(row.errcode)),
-    colsHeader.errline.visible && h('td.cell.cell-bordered', row.errline),
-    colsHeader.errsource.visible && h('td.cell.cell-bordered', row.errsource),
-    colsHeader.message.visible && h('td.cell.cell-bordered', { title: row.message }, row.message),
+    colsHeader.date.visible && cellWithContextMenu(
```
graduta: Could you please destruct the JSON to remove the duplication of variables such as 'colsHeader', 'row'.

InfoLogger/public/log/cellContextMenu.js
graduta, line 1: Please have a look at PopoverEngine and its associated components from WebUI framework so that we do not introduce duplicated functionality.
You can see how they are used also in Bookkeeping

InfoLogger/public/log/Log.js
```diff
@@ -70,6 +70,30 @@ export default class Log extends Observable {
     this.dom = {
       table: '',
     };
+
+    this.contextMenu = {
```
graduta: I think the context menu should be extracted in its own model and state

InfoLogger/public/log/cellContextMenu.js
```diff
@@ -0,0 +1,116 @@
...
+  left: Math.max(0, Math.min(
+    x,
+    window.innerWidth - MENU_WIDTH - SEVERITY_CANVAS_WIDTH_PX - (inspectorEnabled ? remToPx(INSPECTOR_WIDTH_REM) : 0),
+  )),
+  top: Math.max(0, Math.min(y, window.innerHeight - MENU_HEIGHT_ESTIMATE)),
+});
+
+/**
+ * Context menu for log table cells — allows quick filter actions.
+ * Rendered at view root with position:fixed to avoid virtual scroll issues.
```
graduta, line 42: What kind of issues you encountered?

graduta (changes requested):
Very nice context menu, good job!
Please see below a few comments and suggestions based on existing use and behaviour of ILG:
- I believe it would be good to display a hover pop-up when hovering that informs the user to right-click for more options
- the right click menu should also have the option to open the inspector
- match and exclude should append instead of replacing the filter. Currently Infologger for all fields except date accept:
    - for one world filters: multi-search by space separation. e.g. system box can accept "ECS GUI" and will search for OR
    -  message filter (match/exclude) accepts searching by long messages and the OR is applied on new line separator
    - if we simply "replace" rather than append, shifters will lose existing filters and we do not want that as we give them a pre-defined configuration most of the time with messages to exclude for example
- severity and label columns should also update the toggle filter buttons

InfoLogger/public/log/cellContextMenu.js
isaachilly: Brain fog 😶‍🌫️, I looked at the CSS file and saw the variable and completely let the fact it's the whole point of the class go over my head.

InfoLogger/public/log/cellContextMenu.js
isaachilly: This is a two-fold comment, it regards:
- What should happen to the menu when a scroll event occurs (user or autoscroll), the context menu should not drift with the content, or should it? `position: fixed` anchors it so it won't fly upward during any scroll. Note the overlay currently prevents scrolling entirely, which is a side effect rather than intentional.
- Secondly it is rendered at the view root as if it was inside the table row, the virtual scroller would recycle it as soon as that row scrolled out of view.

So it's a bit nuanced. In Query mode, you might want the menu to scroll with its row, disappearing off-screen and reappearing when the row scrolls back into view. In Live mode, where autoscroll continuously pushes rows upward, maybe you would want it fixed.

isaachilly:
> Very nice context menu, good job! Please see below a few comments and suggestions based on existing use and behaviour of ILG:
> 
>     * I believe it would be good to display a hover pop-up when hovering that informs the user to right-click for more options
> 
>     * the right click menu should also have the option to open the inspector
> 
>     * match and exclude should append instead of replacing the filter. Currently Infologger for all fields except date accept:
>       
>       * for one world filters: multi-search by space separation. e.g. system box can accept "ECS GUI" and will search for OR
>       * message filter (match/exclude) accepts searching by long messages and the OR is applied on new line separator
>       * if we simply "replace" rather than append, shifters will lose existing filters and we do not want that as we give them a pre-defined configuration most of the time with messages to exclude for example
> 
>     * severity and label columns should also update the toggle filter buttons

- What do you think about the level filtering? I've set it to have the options of either set level to the next lowest/highest mode.
- See my reply about the fixed positioning comment.
- I have not yet looked at PopoverEngine. When I started the ticket I specifically looked at similar CSS-wise components but they didn't map so well hence I wrote my some inspired purpose-built classes instead.
- Everything else mentioned I added as I understood.

graduta (changes requested):
Very nice progress! I believe there are two more issues that need addressing and then PR will be good to be shipped.

1. I did not see a reply or implementation to my suggestion on hover of cell so that the users are aware of the feature
2. It seems that a small bug is introduced where the separator line between the message column and one before is not there anymore (see screenshot)

Image

graduta (approved):

PR #3465, opened 2026-05-21
[OGUI-1216] Overrise browser-zoom to increase log size only

isaachilly:
- [x] PR labels are selected

* Intercept Ctrl/CMD plus scroll and Ctrl/CMD plus +/- to zoom the log table font and row height independently of the browser, keeping the rest of the UI at normal scale
* Zoom state has own model imported into Model
* CSS custom properties `-log-font-size` and `--row-height` derive from the Zoom model
* Added a minus, reset and add zoom buttons in the toolbar
* Added a test suite covering keyboard zoom, mouse wheel zoom, button zoom, min/max clamping, reset button state, and constant row-height-to-font-size ratio

### DIFF
--- a/InfoLogger/public/log/commandLogs.js
+++ b/InfoLogger/public/log/commandLogs.js
@@ -12,7 +12,13 @@
  * or submit itself to any jurisdiction.
  */
 
-import { h, iconPerson, iconMediaPlay, iconMediaStop, iconDataTransferDownload } from '/js/src/index.js';
+import { h,
+  iconPerson,
+  iconMediaPlay,
+  iconMediaStop,
+  iconDataTransferDownload,
+  iconMagnifyingGlass,
+} from '/js/src/index.js';
 import { BUTTON } from '../constants/button-states.const.js';
 import { MODE } from '../constants/mode.const.js';
 import { setBrowserTabTitle } from '../common/utils.js';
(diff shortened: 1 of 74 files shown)


InfoLogger/public/log/commandLogs.js
```diff
@@ -58,6 +64,11 @@ export default (model) => [
     title: 'Go to last log message (ALT + down arrow)',
   }, '↓'),
   downloadButtonGroup(model.log),
+  h('button.btn.flex-row', {
```
graduta: I think it would be nice to have a button group which contains also a + and - next to the reset zoom option

InfoLogger/public/view.js
```diff
@@ -32,7 +32,47 @@ import errorComponent from './common/errorComponent.js';
...
+        e.preventDefault();
+        const now = Date.now();
+        if (now - model.zoom.lastScrollTime < 50) {
+          return;
+        }
+        model.zoom.lastScrollTime = now;
+        e.deltaY < 0 ? model.zoom.zoomIn() : model.zoom.zoomOut();
+      };
+
+      const handleKeyDown = (e) => {
```
graduta: There is also a key handler in the Model.js
What do you think about keeping all key handlers together?

graduta (changes requested):

InfoLogger/public/log/commandLogs.js
isaachilly: I think this is then too many buttons at the top. On my Mac screen size a plus and a minus button would causes wrapping.

InfoLogger/public/view.js
isaachilly: Yes completely agree.

InfoLogger/public/log/commandLogs.js
graduta: Very well then, for me when I tried locally was still ok.

isaachilly: I will try again, I think when I attempted to do it at the start I used IconPlus etc and could've pushed it over the edge.

graduta (approved):

PR #3479, opened 2026-05-26
[OGUI-1874] Notify user on Buffer being full

isaachilly:
- [x] PR labels are selected

image

image

- Adds a dot indicator to track whether the query hit the buffer limit or not.
- Changes format of messages count in the status bar from "100000 messages" to "100,000 / 100,000 (Buffer size)".
- Displays a warning notification when the buffer size is hit.
- Adds tests for the status bar
- Enhances notifications by adding click-top-copy functionality.
- Enhances notifications by making them not hide if hovered over.


InfoLogger/public/log/statusBar.js
```diff
@@ -135,3 +154,20 @@ const statusStats = (model) => [
...
   h('span.ph1.severity-f', `${model.log.stats.fatal} fatal`),
 ];
+
+const bufferStatus = (model) => {
+  let dotColor = '#6c757d'; // grey - unknown status default
+
+  if (model.log.limitReached === true) {
+    dotColor = '#d9534f'; // red - limit reached
+  } else if (model.log.limitReached === false) {
+    dotColor = '#5cb85c'; // green - limit not reached
```
graduta: Hi, I think it is best to keep consistency across the project and use classes rather than hardcoded colors.

InfoLogger/public/log/statusBar.js
```diff
@@ -135,3 +154,20 @@ const statusStats = (model) => [
...
+  }
+
+  return h('span', {
+    style: `color: ${dotColor}; font-size: 0.8em; margin: 0 4px;`,
```
graduta: Same goes for font and margins. We have classes defined for this, please have a look at the framework's bootstrap.css

InfoLogger/test/public/status-bar-mocha.js
```diff
@@ -0,0 +1,371 @@
...
+
+    describe('Dot Color and Tooltip', () => {
+      it('should show grey dot and title when limitReached is null', async () => {
+        await page.evaluate(() => {
+          window.model.log.limitReached = null;
+          window.model.notify();
+        });
+
+        await page.waitForFunction(
+          () => document.querySelector('#status-bar-buffer-dot').style.color === 'rgb(108, 117, 125)',
```
graduta: Same comment about classes especially as in the code you used hex, and this translated in this test in rgb. I think is best to use classes

graduta (changes requested):
Please keep in mind that framework is a different NPM package. Thus, it needs to have its own ticket and PRs, followed up by a release. Only after, it can be used in the other projects.
Given that, the improvements for Notification are not to be applied in this PR at all

isaachilly:
> Please keep in mind that framework is a different NPM package. Thus, it needs to have its own ticket and PRs, followed up by a release. Only after, it can be used in the other projects. Given that, the improvements for Notification are not to be applied in this PR at all

Ah yes, my fault, will remove and stash.

graduta (approved):

PR #3480, opened 2026-05-27
[OGUI-914] Disable debug logs for "Ops" filter

isaachilly:
- [x] PR labels are selected

> [!WARNING]  
> FLP integration tests need changing to match new default page.

- When OPS is selected, the DEBUG severity button is disabled and stripped from filters automatically from both the URI and the button methods via `LogFilter.setCriteria` central enforcement.
- Log level is set to OPS at startup and in the profile defaults.

**Stealth Fix:** The reset button and context menu severity reset were calling filter methods directly which skipped the notification and more importantly in live mode would skip resetting the filter/s entirely.

### DIFF
--- a/InfoLogger/public/log/Log.js
+++ b/InfoLogger/public/log/Log.js
@@ -32,8 +32,8 @@ export default class Log extends Observable {
     this.model = model;
 
     this.filter = new LogFilter(model);
-    this.filter.setCriteria('level', 'max', 1);
     this.filter.bubbleTo(this);
+    this.filter.observe(this.onFilterChange.bind(this));
 
     this.focus = { // show date picker on focus
       timestampSince: false,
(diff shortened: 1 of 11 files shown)


InfoLogger/public/log/Log.js
```diff
@@ -32,6 +32,7 @@ export default class Log extends Observable {
     this.model = model;
 
     this.filter = new LogFilter(model);
+    this.filter.setCriteria('level', 'max', 1);
```
graduta: I think the constructor should take care of the setCriteria. In this case it will call `resetCriteria`.
As the user requested that the new default is OPS, I believe the resetCriteria should reset to default. What do you think?

InfoLogger/public/log/Log.js
```diff
@@ -401,16 +402,31 @@ export default class Log extends Observable {
...
+      this.notifyFilterChanged();
+    }
+  }
+
+  /**
+   * Reset all filter criteria and notify active mode of the change.
+   */
+  resetFilters() {
+    this.filter.resetCriteria();
+    this.notifyFilterChanged();
```
graduta: LogFilter class extends Observable. Thus, I would be in favour for you to observe the changes in LogFilter and this would in turn automatically call `notifyFilterChanged` for both on L405 and L414

InfoLogger/public/log/Log.js
```diff
@@ -401,16 +402,31 @@ export default class Log extends Observable {
...
+  }
+
+  /**
+   * Notify the active mode (live or query) that filters have changed.
+   */
+  notifyFilterChanged() {
```
graduta: A better name would be perhaps: `onFilterChange`, what do you think?

graduta (changes requested):

InfoLogger/public/log/Log.js
isaachilly: I agree with your simpler and more logical approach of building in the default into `resetCriteria`.

graduta (approved):

PR #3482, opened 2026-05-29
[OGUI-1298] Add button to cancel ongoing query request

graduta:
- [x] PR labels are selected

PR which:
* adds utility methods for using browser native fetch methods with JSON and POST instead of existing loader method from webui framework (behaviour similar to BKP and other GUIs). This is needed to allow for passing an Abort.Signal 
* Refactors the code for query and live buttons so that they do not duplicate the check for frameworkInfo request state
* Query Button is replaced by Cancel button during an ongoing query, allowing the user to cancel their current query and go back to the logs of the previous query (if any)
* mocha index has been updated so that headless mode does not use the `new` deprecated parameter anymore
* adds tests for aforementioned changes

### DIFF
--- a/InfoLogger/public/log/Log.js
+++ b/InfoLogger/public/log/Log.js
@@ -341,6 +341,7 @@ export default class Log extends Observable {
     if (!this.model.frameworkInfo.isSuccess() || !this.model.frameworkInfo.payload.mysql.status.ok) {
       throw new Error('Query service is not available');
     }
+
     if (!this.filter.hasActiveTextFilters()) {
       if (!window.confirm('No date or text filters set.'
         + ' This will return a large amount of data. Execute query anyway?')) {
--- a/InfoLogger/public/log/Log.js
+++ b/InfoLogger/public/log/Log.js
@@ -17,6 +17,7 @@ import LogFilter from '../logFilter/LogFilter.js';
 import ContextMenu from './ContextMenu.js';
 import { MODE } from '../constants/mode.const.js';
 import { TIME_MS } from '../common/Timezone.js';
+import { jsonPost } from '../common/jsonPost.js';
 
 /**
  * Model Log, encapsulate all log management and queries
@@ -48,6 +49,7 @@ export default class Log extends Observable {
     this.limitReached = null;
 
     this.queryResult = RemoteData.notAsked();
+    this.queryAbortController = null;
 
     this.list = [];
     this.item = null;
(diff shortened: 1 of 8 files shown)


InfoLogger/test/public/query-mode-mocha.js
```diff
@@ -55,15 +88,16 @@ const runQueryWithMocks = (page, { confirmReturn, textFilterOperator }) =>
       return confirmReturn;
     };
 
-    window.model.loader.post = async () => {
+    window.fetch = async () => {
       postCalls += 1;
-      return { ok: true, result: { rows: [] } };
+      return { ok: true, status: 200, json: async () => JSON.stringify({ rows: [] }) };
```
isaachilly: I think response returns an object right not a JSON string?

InfoLogger/public/log/Log.js
```diff
@@ -327,59 +329,86 @@ export default class Log extends Observable {
...
+      this.limitReached = result.count === this.limit;
       if (this.limitReached) {
         this.model.notification.show(
           `Matching results reached the buffer size of ${this.limit.toLocaleString('en-US')}.`
           + ' There might be more logs that match your filters but are not shown, consider refining your filters.',
           'warning',
         );
       }
+    } catch (error) {
+      if (abortController.signal.aborted || error.name === 'AbortError') {
```
isaachilly: Error has had `name` stripped by this point in `jsonPost.js L:37` so I think the `error.name === 'AbortError'` is redundant. What do you think?

InfoLogger/public/log/Log.js
```diff
@@ -327,59 +329,86 @@ export default class Log extends Observable {
...
-   * `list` is then reset and filled with result.
+   * Method to execute a query with the current filters configuration
+   * If the user has no filters set, a prompt is shown to confirm the execution
+   * If the user is in live mode, first stop live mode and then execute query in order to have a consistent result
+   * Recalculate the stats and go to last log once query is executed
+   * If the query is aborted by user, restore previous query result and do nothing
+   * @returns {Promise<null|object>} null if query is aborted, result of the query otherwise
    */
   async query() {
-    if (!this.model.frameworkInfo.isSuccess() || !this.model.frameworkInfo.payload.mysql.status.ok) {
```
isaachilly, line 336: I understand the `Query` button is disabled if the DB isn't ready but things like `Enter` key still can call it so I think the check is still good to have?

InfoLogger/public/log/commandLogs.js
```diff
@@ -69,6 +71,84 @@ export default (model) => [
...
+      className: BUTTON.DANGER,
+      onclick: () => logModel.cancelQuery(),
+    }, 'Cancel');
+  }
+
+  return h('button.btn.bold', {
+    id: 'query-button',
+    title: isDbReady ? 'Query database with filters (Enter)' : 'Query service not configured',
+    disabled: !isDbReady || queryResult.isLoading(),
+    className: queryResult.isLoading() ? 'loading' : queryButtonType,
```
isaachilly: We now have the `isLoading` check above so this extra one is not needed. What do you think?

isaachilly (changes requested):

InfoLogger/public/log/Log.js
graduta: Good catch, I am also adding a note in the docs to prevent removal in the future

InfoLogger/public/log/commandLogs.js
graduta: Indeed!

InfoLogger/test/public/query-mode-mocha.js
graduta: Yes. Initially I used `text()` and converted to JSON and then i moved to using `json` and missed the test logic update. 
Thank you!

InfoLogger/public/log/Log.js
graduta: You are correct, the second part of the check is dead code. I removed it

isaachilly (approved):

PR #3483, opened 2026-05-29
[OGUI-1902] Add abort listener on query endpoint with signal to query service to drop DB connection if so

graduta:
- [x] PR labels are selected

PR which:
* adds listeners on the response (close, finish) events together with an abort controller so that the server can stop any ongoing queries if client cancells the request
* move towards using `pool.connection` instead of `pool.query` so that we can drop the connection in case of signal being aborted.
* add cancellation methods with AbortError classs
* attempt to improve stability of test on front-end

### DIFF
--- a/InfoLogger/lib/utils/queryCancellation.js
+++ b/InfoLogger/lib/utils/queryCancellation.js
@@ -43,16 +43,16 @@ const throwIfQueryAborted = (signal) => {
  * Return a cleanup function to remove the listener
  * @param {AbortSignal|null} signal - optional abort signal
  * @param {object} connection - mariadb connection-like object
- * @param {function(): void} onDestroyed - callback called just before destroying connection
+ * @param {function(): void} beforeDestroy - callback called just before destroying connection
  * @returns {function(): void} cleanup callback to remove the listener
  */
-const attachAbortDestroyHandler = (signal, connection, onDestroyed) => {
+const attachAbortDestroyHandler = (signal, connection, beforeDestroy) => {
   if (!signal) {
     return () => {};
   }
 
   const abortHandler = () => {
-    onDestroyed();
+    beforeDestroy();
     connection.destroy();
   };
 
--- a/InfoLogger/lib/utils/queryCancellation.js
+++ b/InfoLogger/lib/utils/queryCancellation.js
@@ -13,15 +13,28 @@
  */
 
 /**
+ * Custom error class for query cancellation
+ */
+class AbortError extends Error {
+  /**
+   * Create an AbortError
+   * @param {string} message - error message
+   */
+  constructor(message = 'Query cancelled by client') {
+    super(message);
+    this.name = 'AbortError';
+    this.code = 'QUERY_CANCELLED';
+  }
+}
+
+/**
  * Throw an error if the given signal is already aborted.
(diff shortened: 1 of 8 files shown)


InfoLogger/lib/utils/queryCancellation.js
```diff
@@ -0,0 +1,67 @@
...
+    throw new AbortError();
+  }
+};
+
+/**
+ * Attach a one-time abort handler to the signal that destroys the connection.
+ * Return a cleanup function to remove the listener
+ * @param {AbortSignal|null} signal - optional abort signal
+ * @param {object} connection - mariadb connection-like object
+ * @param {function(): void} onDestroyed - callback called just before destroying connection
```
isaachilly: I think the parameter `onDestroyed` is confusing given it is called before like the description says. What do you think?

InfoLogger/test/lib/controller/mocha-query-controller.test.js
```diff
@@ -106,11 +107,20 @@ describe('QueryController test suite', () => {
...
       queryService.queryFromFilters.resolves(logs);
 
       await queryController.getLogs({ body }, res);
 
-      assert.ok(queryService.queryFromFilters.calledWith(body.criterias, body.options));
+      assert.ok(queryService.queryFromFilters.called);
+      const call = queryService.queryFromFilters.getCall(0);
+      assert.strictEqual(call.args[0], body.criterias);
+      assert.strictEqual(call.args[1], body.options);
+      assert.ok(call.args[2], 'AbortSignal'); // third arg should be the signal
```
isaachilly: `assert.ok` is only checking truthiness here, not the type. I think either change it to check the type with `instanceOf` or reduce it to just the truthiness check. What do you think or prefer?

InfoLogger/test/lib/utils/mocha-query-cancellation.js
```diff
@@ -0,0 +1,89 @@
...
+          assert.ok(error instanceof AbortError);
+          assert.strictEqual(error.message, 'Query cancelled by client');
+          assert.strictEqual(error.code, 'QUERY_CANCELLED');
+          return true;
+        },
+      );
+    });
+  });
+
+  describe('\'attachAbortDestroyHandler()\' - test suite  ', () => {
```
isaachilly: Just some trailing whitespace

isaachilly (changes requested):
Overall I have only minor comments, looks good to me!

InfoLogger/test/lib/utils/mocha-query-cancellation.js
graduta: Interesting to see that the auto linter from VS Code did not fix this. 
Fixed

InfoLogger/test/lib/controller/mocha-query-controller.test.js
graduta: Updated to add instanceof but also a check that it was not aborted

InfoLogger/lib/utils/queryCancellation.js
graduta: Sure, I renamed to `beforeDestroy`

isaachilly (approved):

PR #3486, opened 2026-06-02
[OGUI-1193] Filters should allow for empty fields to be searched

isaachilly:
- [X] PR labels are selected

> [!WARNING]
>  FLP integration tests need to be changed to match newly introduced `emptyFor` operator.

Adds `emptyFor` operator so users can filter on empty fields alongside the existing match/exclude text filters. Also includes a small refactor of the filter internals and test logic.

**UI**
- Adds toggle buttons at the end of each match/exclude filter (except timestamp).
- The toggle appears on hover, and stays visible while active even after the cursor leaves.

**Filter behaviour**
* Toggled on alone matches/excludes empty values.
* Toggled on with text:
    * **match:** matches the text *OR* empty values
    * **exclude:** exlcudes the text *AND* empty values.

**Implementation**
- Adds operator handling in `LogFilter` and `QueryService`.
- Refactors the `QueryService` match/exclude SQL builders (map/join instead of string concat) and adds helpers where needed.
- `LogFilter.setCriteria` now validates the operator against the field's known criteria and throws on unknown operators.

**Tests**
- Adds tests covering the new operators (unit, query-mode, and live-mode).
- Fixes ILG ESLint issues and test reliability/stability problems (see individual commit titles).
- Drops the `includes` check in severity filtering as this could pass on both `$in` and `in.

### DIFF
--- a/InfoLogger/public/logFilter/LogFilter.js
+++ b/InfoLogger/public/logFilter/LogFilter.js
@@ -33,8 +33,8 @@ import { getDisabledSeverities } from '../constants/log-level-filters.const.js';
  */
 const makeDefaultMatchExcludeOperators = () => ({
   match: '',
-  exclude: '',
   $match: null,
+  exclude: '',
   $exclude: null,
   matchEmpty: false,
   $matchEmpty: false,
--- a/InfoLogger/public/logFilter/LogFilter.js
+++ b/InfoLogger/public/logFilter/LogFilter.js
@@ -101,10 +101,10 @@ export default class LogFilter extends Observable {
           this.criterias[field]['$in'] = value ? value.split(' ') : null;
           break;
         case 'matchEmpty':
-          this.criterias[field]['$matchEmpty'] = value;
+          this.criterias[field]['$matchEmpty'] = Boolean(value);
           break;
         case 'excludeEmpty':
-          this.criterias[field]['$excludeEmpty'] = value;
+          this.criterias[field]['$excludeEmpty'] = Boolean(value);
           break;
         default:
           throw new Error('unknown operator');
@@ -115,8 +115,8 @@ export default class LogFilter extends Observable {
         this.enforceDisabledSeverities();
       }
 
-      if (operator === 'matchEmpty' || operator === 'excludeEmpty') {
         // Ensure that both matchEmpty and excludeEmpty are not active at the same time
+      if (value && (operator === 'matchEmpty' || operator === 'excludeEmpty')) {
         const oppositeKey = operator === 'matchEmpty' ? 'excludeEmpty' : 'matchEmpty';
         if (this.criterias[field][oppositeKey]) {
           this.criterias[field][oppositeKey] = false;
@@ -269,6 +269,15 @@ export default class LogFilter extends Observable {
       }
 
       /**
+       * Whether a log field value is considered empty for matchEmpty/excludeEmpty purposes.
+       * @param {*} logValue - value of the log field
+       * @returns {boolean} - true if the value is undefined, null, or an empty string
+       */
+      function isEmpty(logValue) {
+        return logValue === undefined || logValue === null || logValue === '';
+      }
+
+      /**
        * Function that applies the criteria of one filter set by the user on each received logValue
        * @param {object} logValue - value of the log field that is to be checked (e.g. message, severity, etc.)
        * @param {object} criteria - object containing the criteria if applied by the user
@@ -290,7 +299,7 @@ export default class LogFilter extends Observable {
               break;
             }
             case '$match': {
-              if (logValue === undefined || logValue === '') {
+              if (isEmpty(logValue)) {
                 if (!criteria.$matchEmpty) {
                   return false;
                 }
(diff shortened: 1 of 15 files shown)


graduta (changes requested):
Hi @isaachilly , very nice proposal for allowing users to either match/exclude the empty fields. Also, very nice refactor on some of the code!

I believe however that there are some discrepancies between proposal for front-end and back-end and I think the back-end in this case should follow the front-end.

Basically, you are adding four extra fields for almost every criteria for the `matchEmpty, $matchEmpty, excludeEmpty, $excludeEmpty`. I see the point that made you do this, that is to keep consistent with the existing code but I believe it is not the right direction and this situations is one in which we should change existing standard. 
Firstly, as per the front-end implementation, these two newly added variables cannot be set to true at the same time. It is either this or that. Thus, I believe the back-end should not expect two variables, but instead something along the lines `emptyFor: 'null/match/exclude`. 
Moreover, the proposed attributes from your side are of type `boolean`. The `$` based variables are to inform of data processing needed but in this case for boolean, it does nothing to them. I at the same time acknowledge the mistake of existing categories of text for match and exclude which do not need the dollar sign either. 

Thus, I think it would be better to change a bit the direction of the criteria + backend proposal. The front-end looks very nice and works as expected

isaachilly:
I have reduced the feature to a single `emptyFor` that holds `null/'match'/'exclude'`, which matches more the shape you proposed. Regarding the `$` duplicates, I have kept for `emptyFor` because making it an exception at this point is just be spreading inconsistency further around and it should be properly fixed in another ticket. I agree the wider `$`-prefix standard is questionable when there's no parsing step and I'd be happy to open up a follow-up ticket for rationalising it across all affected operators.

InfoLogger/public/logFilter/LogFilter.js
```diff
@@ -156,7 +175,10 @@ export default class LogFilter extends Observable {
    */
   hasActiveTextFilters() {
     return Object.values(this.criterias).some((criteria) =>
-      TEXT_FILTER_OPERATORS.some((operator) => criteria[operator]?.trim()));
+      TEXT_FILTER_OPERATORS.some((operator) => {
+        const v = criteria[operator];
```
graduta: We do not use single letter variables

InfoLogger/public/logFilter/LogFilter.js
```diff
@@ -156,7 +175,10 @@ export default class LogFilter extends Observable {
...
+        return typeof v === 'string' ? v.trim() : Boolean(v);
```
graduta: I guess this change is remainder from the previous design with Booleans. Should it not be reverted now?

InfoLogger/public/logFilter/LogFilter.js
```diff
@@ -172,15 +194,14 @@ export default class LogFilter extends Observable {
...
-    const disabled = getDisabledSeverities(this.criterias.level.max);
-    if (disabled.length === 0 || !this.criterias.severity.$in) {
-      return;
-    }
-
     const current = this.criterias.severity.$in;
     if (!current) {
       return;
     }
+    const disabled = getDisabledSeverities(this.criterias.level.max);
```
graduta, line 198: What is the reasoning for re-order checks here?

InfoLogger/public/logFilter/LogFilter.js
```diff
@@ -249,7 +279,7 @@ export default class LogFilter extends Observable {
         for (const operator in criteria) {
           let criteriaValue = criteria[operator];
           // don't apply criterias not set
-          if (criteriaValue === null) {
+          if (criteriaValue === null || criteriaValue === false) {
```
graduta: I could not find a case where it is a boolean value. Could you explain please?

InfoLogger/public/logFilter/tableFilters.js
```diff
@@ -151,24 +158,41 @@ const createInputField = (logModel, field, command, tabIndex = 1) => h('td', h('
...
-        'TASK %QC% running out of memory\n' +
-        'weird error with strict message',
-    onfocus: () => {
-      model.messageFocused = true;
-      model.notify();
-    },
-    onfocusout: () => {
-      model.messageFocused = false;
-      model.notify();
+  h('td', h('.filter-input-group', [
```
graduta: To be checked with Sylvain, but as per his [docs](https://github.com/AliceO2Group/InfoLogger/blob/master/doc/README.md),  I believe the `message` field in InfoLogger cannot be `empty`.
Thus, we do not this feature and complex logic and widget for the message

graduta (changes requested):

InfoLogger/public/logFilter/LogFilter.js
isaachilly: There is no point in calculating the disabledSeverities if `!current` and return early anyways.

InfoLogger/public/logFilter/LogFilter.js
isaachilly: Hangover from matchEmpty/excludeEmpty. Will remove.

graduta (approved):

PR #3493, opened 2026-06-05
[OGUI-1904] Use select instead of in-line buttons to reduce space across top header

graduta:
- [x] PR labels are selected

PR which:
* adjusts the HTML and CSS so that top header with commands can fit in one line even for 13.5 screens
* use a newly added select component instead of in-line buttons for the two (level, limit) filters that accept one value only
* add tests

### DIFF
--- a/InfoLogger/public/app.css
+++ b/InfoLogger/public/app.css
@@ -14,6 +14,11 @@
 
 /* make the sidebar big enough to contain inspector */
 
+:root {
+  --color-gray-button-background-hover: #a0a0a0;
+  --color-gray-button-background: #e0e0e0;
+  --color-gray-button: #404040;
+}
 .sidebar,
 .sidebar .sidebar-content { width: 20rem; }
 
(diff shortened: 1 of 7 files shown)


InfoLogger/public/logFilter/commandFilters.js
```diff
@@ -73,30 +58,63 @@ const buttonSeverity = (model, label, title, value) => {
...
- * Makes a button to set filtering level (shifter, debug, etc) with number
- * @param {Model} model - root model of the application
- * @param {string} label - button's label
- * @param {number} value - maximum level of filtering, from 1 to 21
- * @returns {vnode} - component representing the creation of a button for filtering
+ * Build a vnode for a select component with given options and onchange callback
+ * @param {string} title - tooltip shown on hover
+ * @param {string} id - id of the element
+ * @param {Array<{label: string, value: string, selected: boolean}>} options - list of options to render
+ * @param {void} onchange - callback receiving the raw string value from the select
```
isaachilly, line 65: Should this be void or function?

graduta: Function is not liked by eslint

InfoLogger/public/app.css
```diff
@@ -124,6 +124,9 @@ footer { border-top: 1px solid var(--color-gray); }
 .btn:hover { background-color: #a0a0a0; color: var(--color-white); }
 .btn:active, .btn.active, .dropdown-open > .btn { background-color: #c0c0c0; }
 
+.select-btn { background-color: #e0e0e0; color: #404040; border: 0; border-radius: .25rem; padding: 0em 0.5em; font-size: 1em; font-family: inherit; cursor: pointer; }
```
isaachilly: Just checking that no CSS vars / other classes could've been used here?

isaachilly (changes requested):

InfoLogger/public/app.css
graduta: So in this case, not. I used the variable for white color, but for this grey, we currently do not have a variable in the webui. 
I used the same hex code as the buttons declared 2 lines above.

isaachilly: Okay, that's fine. I did do a rough search myself but it was easier to raise the point with a question mark for you to check.

graduta: I added variables for the 3 buttons colors in infologger so that we can easily change if needed across the elements

isaachilly (dismissed):

PR #3563, opened 2026-08-03
[OGUI-1913] Reset test profile DB before each test run to fix alternating pass/fail

isaachilly:
`testdb.json` is used by the test server to persist user profiles, but it is never cleaned up between separate test runs (`npm run test`).

`successfully save the profile of the user when pressed the "Save Profile" menu-item` test in `user-actions-mocha.js` toggles a column's visibility and saves it. Because the file isn't reset between runs, this modified state carries over to the next time the suite runs, causing the subsequent test, which relies on the visibility being what it originally was, to fail on that next run.

- [x] PR labels are selected

### DIFF
--- a/InfoLogger/test/mocha-index.js
+++ b/InfoLogger/test/mocha-index.js
@@ -16,6 +16,7 @@
 const puppeteer = require('puppeteer');
 const assert = require('assert');
 const { spawn } = require('child_process');
+const fs = require('fs');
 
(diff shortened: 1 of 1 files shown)


graduta (approved):
Looks good. The only remark I have to make sure that in the commit you mention this is needed for local testing.
In GH Actions this is not an issue as we always start from a clean state.

PR #3602, opened 2026-09-01
[OGUI-1920] Encode filter query in route updates

isaachilly:
- [x] PR labels are selected

Ticket:
The `q` URL parameter in ILG, containing the JSON object of active filters, is not being encoded in the correct order creating opportunity for corruption when sharing or decoding.

The changes to the code include:
- Reversing the order of encoding to be correct and encoding the whole object not just its values.
- Removing any special handling of new lines or quotations as encoding handles this safely now.
- Updating tests that assert against URL strings to use their fully-encoded versions.
- Adding round-trip URL tests so encoding/decoding changes will be held up against best practice, fully-encoded URLs.

Reviewer can test with a filter value of `C:\temp` on a dev instance before and after the branch's changes and note the difference:
- Before the URL contains raw `{` and `}`and the filter value is `C:<tab>emp` as the `\t` was incorrectly consumed as a JSON escape.
- After the URL contains only valid, percent-encoded characters and the filter value is `C:\temp`.

Reviewer should note the filter JSON object is now wholly-encoded as oppose to before when it was just its values.

Reviewer should note that changes to an ILG integration test will be required as it asserts against not a fully-encoded query string.

Reviewer should note the address bar decodes double quotation marks for readability but will copy the encoded version.

### DIFF
--- a/InfoLogger/test/public/log-filter-url-mocha.js
+++ b/InfoLogger/test/public/log-filter-url-mocha.js
@@ -1,0 +1,81 @@
+/**
+ * @license
+ * Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+ * See http://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+ * All rights not expressly granted are reserved.
+ *
+ * This software is distributed under the terms of the GNU General Public
+ * License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+const assert = require('assert');
+const test = require('../mocha-index');
+
+describe('URL Encoding/Decoding Suite', async () => {
+  let page = null;
+
+  before(async () => {
+    ({ page } = test);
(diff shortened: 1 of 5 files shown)


InfoLogger/test/public/log-filter-url-mocha.js
```diff
@@ -0,0 +1,81 @@
...
+        window.model.log.filter.setCriteria('message', 'match', raw);
+        window.model.updateRouteOnModelChange();
+        return window.location.href;
+      }, value);
+
+      await page.goto(url, { waitUntil: 'networkidle0' });
+      return await page.evaluate(() => window.model.log.filter.criterias.message.match);
+    };
+
+    it('should preserve consecutive double quotes', async () => {
```
graduta: This test and similar below ones, I reckon could also be written as a list of strings to test and use in a  loop. 
While not generally done so in our codebase, I think in this case it makes sense because you can quickly spot the edge cases you are testing by having all in one place, rather than having to go test by test. What do you think?

graduta (dismissed):
I approved as my comment is not a blocker or requested change but feel free to take it into consideration

InfoLogger/test/public/log-filter-url-mocha.js
isaachilly: I think it is a good point. They are all testing the same mechanism. See my most recent push.

graduta (approved):

PR #3615, opened 2026-09-03
[OGUI-1919] Make the copy-to-clipboard component have more flexible styling

isaachilly:
- [x] PR labels are selected

Ticket:
Make the component more flexible with regards to styling. This is needed to be able to use the component in ILG. See ticket.

Code:
- New optional className/style attributes; default rendering (`btn-primary`) is unchanged.
- New optional onFailure attribute called if the clipboard write fails.
- Repeated clicks before the success timer is complete now restart it. Another way of putting it is "Copied!" is shown until 2s pass without the button being clicked.
- Clipboard availability is resolved in the constructor instead of the view.
- The tooltip doesn't read `null` if there is no message, it used to.
- No `id` = undefined attribute appears when no `id` attr is passed.

Reviewer:
To test `onFailiure` logic edit a use of the component and in the console set `writeText` to raise an Error.
To test the other attributes is straightforward.

> [!NOTE]
> `http://localhost:4000/?page=env-overview` BKP uses the component in the ENV id dropdown. You will need to temporarily link the WebUI package to the local one. To do this I edited the docker compose file as follows:
>
> ```
> # LOCAL ONLY - do not commit. Serves the local WebUi frontend
> - type: bind
>  read_only: true
>  source: ../WebUi/Framework/Frontend
>  target: /usr/src/app/node_modules/@aliceo2/web-ui/Frontend
> ```


Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -26,21 +26,44 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
+      this._successStateTimeout = setTimeout(() => {
+        this._successStateTimeout = null;
+        this.notify();
+      }, 2000);
+
       this.notify();
-    }, 2000);
-    this.notify();
+    } catch (error) {
+      if (onFailure) {
```
graduta: What do you think about using a default empty method declaration in the method signature rather than adding this `if` statement?

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -73,45 +96,48 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
 
   /**
    * Renders the button that allows copying text to the clipboard.
+   * Attributes other than those listed are not forwarded to the button element.
    *
    * @param {vnode} vnode The virtual DOM node containing the attrs and children.
+   * @param {object} vnode.attrs The attributes passed to the component.
+   * @param {string} [vnode.attrs.value=''] The text to be copied to the clipboard.
+   * @param {string} [vnode.attrs.id] The unique identifier for the copy button will become 'copy-{id}'.
+   * @param {string} [vnode.attrs.className='btn-primary'] The CSS classes to be applied to the copy button.
```
graduta: I think it would be good to document an example of passing multiple classes here. Plus as type you mention both `string|object`, are you sure about that this will work?

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -26,21 +26,44 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
+        this._successStateTimeout = null;
+        this.notify();
+      }, 2000);
+
       this.notify();
-    }, 2000);
-    this.notify();
+    } catch (error) {
+      if (onFailure) {
+        onFailure(error);
```
graduta, line 64: A second point I have here, you do not check if the onFailure is a function type that can be called

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -73,45 +96,48 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
-      message = errorMessage;
-    }
+    const { value: clipboardTargetValue = '', id, className = 'btn-primary', style, onFailure } = attrs;
 
     const defaultContent = [iconLinkIntact(), children];
     const successContent = [iconCheck(), h('', 'Copied!')];
 
     return h(
-      'button.btn.btn-primary',
+      'button.btn',
```
graduta, line 129: I see that this component was missing the `type`. As per the docs recommendation, I think it should be added: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#type

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -73,45 +96,48 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
       {
-        id: `copy-${id}`,
-        onclick: () => this.copyToClipboard(clipboardTargetValue),
-        disabled: !available,
-        title: message || null,
+        id: id ? `copy-${id}` : undefined,
+        onclick: () => this.copyToClipboard(clipboardTargetValue, onFailure),
+        disabled: !this._available,
+        title: this._message,
+        style,
```
graduta, line 136: Is there a specific usecase for providing-inline style? Normally within the projects we stayed away from this and used classes especially as in this format I believe style will override className. 
You could also consider defining the classes on Line128 together with `button.btn` so that they are all in the same place rather than spread

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -73,45 +96,48 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
+        className,
       },
-      h('div.flex-row.g1', this._successStateTimeout ? successContent : defaultContent),
+      h(
+        'div.flex-row.g1.justify-center',
```
graduta: I believe this change is the exact opposite of what you are trying to achieve with the PR for the button component. I would consider passing the styling as a parameter as well rather than forcing justify-center.

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
```diff
@@ -26,21 +26,44 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
+
+    this._available = true;
+    this._message = '';
+
+    try {
+      this.checkClipboardAvailability();
+    } catch ({ message: errorMessage }) {
+      this._available = false;
+      this._message = errorMessage;
+    }
```
graduta, line 38: Trying to understand the reasoning of moving this code block in here, I would assume you did it so that it is not checked on each render of "copy to clipboard" but this component's constructor is anyway called on re-rendered for each notify on the page (more or less).
For example where it is used in BKP, it is now being called 20 times on page load and 19 on page change (pages that contain the button), same as before
Given that normally it is recommended to avoid try/catch in constructors, I would recommend to revert the change

graduta (changes requested):

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
isaachilly: I agree but if we check it is a function (an issue you raised in another comment) we automatically handle also the case where no function is passed so I think I can skip adding this default parameter.

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
isaachilly: Added the class `loading` to demonstrate how to add multiple of them.

With regards to `string|object`, you can test with `min-width: 250px` and `{ minWidth: '100px' }`. It is also documented in the [Mithril.js documentation](https://mithril.js.org/hyperscript.html#dom-attributes).

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
isaachilly: Added.

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
isaachilly: Yes I wanted to be able to stop the button shortening when the success text is shown versus the default text, which it tends to do as the success text is "Copied!".

I looked around for alternatives but they involve a rather ugly layered approach of having 2 text elements and controlling their visibility.

Do you have an alternative or preference that I can implement?

Framework/Frontend/js/src/components/CopyToClipboardComponent.js
isaachilly: Given redraws should only call `view()` these BKP constructor calls must be and are coming from remounts. The table's body is being destroyed and remounted and thus the same for all its copy buttons.

This change was aimed at ILG and in ILG the copy button will be in the toolbar/header which is mounted once but live mode creates a new notify for each added log which would've meant calling `checkClipboardAvailability()` every log.

I suggest instead redesigning `checkClipboardAvailability()` so as to not need a try/catch block in the constructor. What do you think?

graduta (approved):

